PR #144, opened 2018-12-26
[experiment] Generate CAs

kron4eg:
Generate different CAs by kubeone, without frictions with kubeadm

```release-note
NONE
```


kron4eg:
maybe we will return to this somaday, but definitely not now

Issue #269, opened 2019-03-15
Support Kubernetes 1.14


xmudrii:
With Kubernetes 1.14 coming out on 26th March, we should start planning and working on support for Kubernetes 1.14.

xmudrii:
With #328 merged we have resolved all release-blocking issues we had. In the morning I'll start staging the v0.5.0 release.

/close

PR #328, opened 2019-04-02
Implement Retryable tasks

xmudrii:

This PR implements retryable tasks. Some tasks are ran only once while some are ran 3 times. This package is based on the `k8s.io/apimachinery/util/wait` package and the `ExponentialBackoff` function.

This is required for proper upgrades support from 1.13 to 1.14 and to stabilize the CI.

xref #269

**Release note**:
```release-note
NONE
```

/assign @kron4eg

### DIFF
--- a/pkg/installer/installation/install.go
+++ b/pkg/installer/installation/install.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/kubermatic/kubeone/pkg/certificate"
 	"github.com/kubermatic/kubeone/pkg/features"
+	"github.com/kubermatic/kubeone/pkg/task"
 	"github.com/kubermatic/kubeone/pkg/templates/machinecontroller"
 	"github.com/kubermatic/kubeone/pkg/util"
 )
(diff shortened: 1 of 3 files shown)


xmudrii:
/test pull-kubeone-e2e-digitalocean-upgrade-1.13.5-1.14.0
This is related to E2E tests and not to the upgrade logic itself. I'll create a PR to fix this.

xmudrii:
The cluster never got provisioned.
/test pull-kubeone-e2e-digitalocean-upgrade-1.13.5-1.14.0

xmudrii:
/override pull-kubeone-e2e-digitalocean-upgrade-1.13.5-1.14.0

xmudrii:
This is a known flake and we had it for a long time now. For some reason API seems to be slower than usual and it takes more time to replace the worker nodes nodes than expected. If you check nodes in the cluster, you'll see that they're getting replaced, but tests timeout by the time the third node should be replaced.

Because this is not a release blocking flake and that tests all other tests have many green runs, I'm going to skip this test.
/override pull-kubeone-e2e-digitalocean-upgrade-1.13.5-1.14.0

PR #334, opened 2019-04-03
Enable scaling machinedeployments subresource

kron4eg:
```release-note
enable scale subresource so that `kubectl scale machinedeployment` would work
```


metalmatze:
Ah, cool! Thanks for enabling this. I already tried to scale a machinedeployment and it didn't work.

PR #341, opened 2019-04-09
Update the upgrades proposal status

xmudrii:

This PR updates the upgrades proposal status to implemented in v0.3.0.

This PR also proposes to change proposals' file names to `date-proposal-name.md` instead of using proposal numbers. This way we keep information about when the proposal was created, we have all proposals sorted, but we don't need to coordinate proposal numbers across multiple PRs and keep them up to date.

**Release note**:
```release-note
NONE
```
/assign @kron4eg

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -10,4 +10,4 @@
 * [Environment variables used by KubeOne](environment_variables.md)
 * [Adding support for provider](adding_provider_support.md)
 * [Proposals](./proposals)
-  * [Cluster Upgrades](./proposals/0002-upgrades.md)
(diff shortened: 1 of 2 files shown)


xmudrii:
The E2E tests are going to fail because the 1.14.1 tests are not supported until #340 isn't merged. As this is only docs PR I'm going to skip failed tests.

/override pull-kubeone-e2e-aws-upgrade-1.13.5-1.14.1
/override pull-kubeone-e2e-digitalocean-upgrade-1.13.5-1.14.0

Issue #259, opened 2019-03-12
KubeOne API and Configuration Versioning


xmudrii:
We should version our API and configuration file to ensure that:
* we can restructure our API while preserving compatibility with the previous API version,
* users can continue using old API version until it's still supported
* we can introduce concept of alpha, beta, and stable APIs,

/kind api-change

xmudrii:
I'll start working on the proposal tomorrow.

/assign
/lifecycle active

PR #353, opened 2019-04-15
Implement the KubeOneCluster API

xmudrii:

This PR implements the `KubeOneCluster` API. The API is similar to the old API, with some changes to improve the user experience.

Group name: `kubeone.io`
Kind: `kubeonecluster`

xref #259 

* Consider using `semver` for versions instead of strings

**Release note**:
```release-note
Implement the KubeOneCluster API
```


pkg/apis/kubeonecluster/doc.go
```diff
@@ -0,0 +1,21 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// +k8s:deepcopy-gen=package
+// +groupName=cluster.kubeone.io
+
+// Package kubeonecluster is the implementation of KubeOneCluster API
+package kubeonecluster
```
kron4eg: `pkg/apis/kubeone` in replace of `pkg/apis/kubeonecluster`

pkg/apis/kubeonecluster/types.go
```diff
@@ -0,0 +1,181 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// KubeOneCluster is KubeOne Cluster API Schema
+type KubeOneCluster struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
```
kron4eg: we don't need ObjectMeta as we don't define a CRD

pkg/apis/kubeonecluster/types.go
```diff
@@ -0,0 +1,181 @@
...
+
+	Spec KubeOneClusterSpec `json:"spec,omitempty"`
```
kron4eg: I still don't see the reason for this dedicated `spec` field

pkg/apis/kubeonecluster/types.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+// KubeOneClusterSpec is spec for the Cluster object
+type KubeOneClusterSpec struct {
+	Hosts        []*HostConfig `json:"hosts,omitempty"`
```
kron4eg: I don't remember why originally it was `[]*HostConfig`, but probably we don't need that and `[]HostConfig` would be just fine.

pkg/apis/kubeonecluster/types.go
```diff
@@ -0,0 +1,181 @@
...
+	APIEndpoints []APIEndpoint `json:"apiEndpoint,omitempty"`
+	// TODO(xmudrii): Provider or another name?
```
kron4eg: `Cloud` maybe?

pkg/apis/kubeonecluster/types.go
```diff
@@ -0,0 +1,181 @@
...
+	Provider       ProviderConfig       `json:"provider,omitempty"`
+	Versions       VersionConfig        `json:"versions,omitempty"`
+	ClusterNetwork ClusterNetworkConfig `json:"clusterNetwork,omitempty"`
+	// TODO(xmudrii): Proxy or another name?
```
kron4eg: `Proxy` is OK to my subjective taste.

pkg/apis/kubeonecluster/types.go
```diff
@@ -0,0 +1,181 @@
...
+	Provider    ProviderName      `json:"provider"`
+	Credentials map[string]string `json:"credentials"`
+}
+
+// Features controls what features will be enabled on the cluster
+type Features struct {
+	PodSecurityPolicy *PodSecurityPolicy `json:"podSecurityPolicy"`
+	DynamicAuditLog   *DynamicAuditLog   `json:"dynamicAuditLog"`
+	MetricsServer     *MetricsServer     `json:"metricsServer"`
+	OpenIDConnect     *OpenIDConnect     `json:"openidConnect"`
```
kron4eg: What's up with all those pointers?

Please don't forget to update type definitions from [pkg/config/cluster.go](https://github.com/kubermatic/kubeone/blob/f5067336c5b7afaaaf4072a8a44ceb35207cc5aa/pkg/config/cluster.go#L342-L381)

pkg/apis/kubeonecluster/v1beta1/credentials.go
```diff
@@ -0,0 +1,115 @@
+/*
```
kron4eg: is it intended that this file end-up in `pkg/apis/kubeonecluster/v1beta1` instead of `pkg/apis/kubeonecluster`?

pkg/apis/kubeonecluster/validation/validation.go
```diff
@@ -0,0 +1,174 @@
...
+		if _, _, err := net.ParseCIDR(c.ServiceSubnet); err != nil {
+			allErrs = append(allErrs, field.Invalid(fldPath, c.ServiceSubnet, "invalid service subnet specified"))
+		}
+	}
+
+	return allErrs
+}
+
+// ValidateFeatures validates the Features structure
+func ValidateFeatures(f kubeonecluster.Features, fldPath *field.Path) field.ErrorList {
```
kron4eg: some validations for `Features` from `pkg/config` is lost

pkg/apis/kubeonecluster/v1beta1/credentials.go
xmudrii: This file will be moved

pkg/apis/kubeonecluster/types.go
xmudrii: Renamed to CloudProviderSpec

Issue #349, opened 2019-04-13
Use core as SSH user for CoreOS Container Linux


metalmatze:
**What happened**:
Used "coreos-stable" as image.

**What is the expected behavior**:
I expect that the output for `tf.json` uses `core` as ssh_user.

**How to reproduce the issue**:
Set `droplet_image = "coreos-stable"` in `terraform.tfvars` before running `terraform apply`.
Then with `terraform output -json > tf.json` you get the output with:

```json
    ...
    "kubeone_hosts": {
        "sensitive": false,
        "type": "map",
        "value": {
            "control_plane": [
                {
                    "cloud_provider": "digitalocean",
                    "cluster_name": "kubeone",
                    "private_address": [
                        "10.10.25.25",
                        "10.10.25.26",
                        "10.10.25.27"
                    ],
                    "public_address": [
                        "142.93.142.142",
                        "142.93.142.143",
                        "142.93.142.144"
                    ],
                    "ssh_agent_socket": "env:SSH_AUTH_SOCK",
                    "ssh_port": "22",
                    "ssh_private_key_file": "",
                    "ssh_user": "root"
                }
            ]
        }
    },
```

**Anything else we need to know?**

**Information about the environment**:
KubeOne version (`kubeone version`): master
Operating system: Arch Linux
Provider you're deploying cluster on: DigitalOcean
Operating system you're deploying on: CoreOS Container Linux

kron4eg:
Provided terraform configurations — are only examples, you’ve more then welcome copy and modify them for your needs.

metalmatze:
Sure. I have no problem with that. However, you also state that `Choice of Linux distributions between Ubuntu, CentOS and CoreOS` in the README. Therefore I'd say the example should at least support setting the ssh_user. I might work on a patch. :+1:

xmudrii:
I agree that the SSH user should be sourced from Terraform variables rather than predefining it in the `output.tf`.

If you want to create a patch, I would be happy to review it. :) You just need to move `ssh_user` from [`output.tf`](https://github.com/kubermatic/kubeone/blob/master/examples/terraform/aws/output.tf#L33) to `variables.tf` and then just source it from `output.tf`. KubeOne is able to parse the `ssh_user` variable, so no other change is required.

PR #355, opened 2019-04-17
aws private network example

### LINKED ISSUE #352: Add advanced AWS terraform configuration example
With all subnets private, and only 1 bastion host available for external ssh connections

kron4eg:
Fixes #352 

```release-note
NONE
```

### DIFF
--- a/examples/terraform/aws-private/output.tf
+++ b/examples/terraform/aws-private/output.tf
@@ -37,7 +37,7 @@ output "kubeone_workers" {
       instanceType     = "${var.worker_type}"
       diskSize         = 100
       sshPublicKeys    = ["${aws_key_pair.deployer.public_key}"]
-      replicas         = 3
+      replicas         = 1
       operatingSystem  = "ubuntu"
     }
 
@@ -52,7 +52,7 @@ output "kubeone_workers" {
       instanceType     = "${var.worker_type}"
       diskSize         = 100
       sshPublicKeys    = ["${aws_key_pair.deployer.public_key}"]
-      replicas         = 3
+      replicas         = 1
       operatingSystem  = "ubuntu"
     }
 
@@ -67,7 +67,7 @@ output "kubeone_workers" {
       instanceType     = "${var.worker_type}"
       diskSize         = 100
       sshPublicKeys    = ["${aws_key_pair.deployer.public_key}"]
-      replicas         = 3
+      replicas         = 1
       operatingSystem  = "ubuntu"
     }
   }
--- a/examples/terraform/aws-private/output.tf
+++ b/examples/terraform/aws-private/output.tf
@@ -1,0 +1,74 @@
+output "kubeone_api" {
+  value = {
+    endpoint = "${aws_lb.control_plane.dns_name}"
+  }
+}
+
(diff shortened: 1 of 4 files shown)


examples/terraform/aws-private/output.tf
```diff
@@ -0,0 +1,74 @@
...
+    pool-a = {
+      region           = "${var.aws_region}"
+      ami              = "${data.aws_ami.ubuntu.id}"
+      availabilityZone = "${data.aws_availability_zones.available.names[0]}"
+      instanceProfile  = "${aws_iam_instance_profile.workers.name}"
+      securityGroupIDs = ["${aws_security_group.common.id}"]
+      vpcId            = "${data.aws_vpc.selected.id}"
+      subnetId         = "${aws_subnet.private.0.id}"
+      instanceType     = "${var.worker_type}"
+      diskSize         = 100
```
xmudrii: Should we make this a variable?

examples/terraform/aws-private/output.tf
```diff
@@ -0,0 +1,74 @@
...
+      sshPublicKeys    = ["${aws_key_pair.deployer.public_key}"]
+      replicas         = 3
```
xmudrii: Isn't this going to produce 9 worker nodes? Is that too much?

examples/terraform/aws-private/output.tf
```diff
@@ -0,0 +1,74 @@
...
+      availabilityZone = "${data.aws_availability_zones.available.names[1]}"
+      instanceProfile  = "${aws_iam_instance_profile.workers.name}"
+      securityGroupIDs = ["${aws_security_group.common.id}"]
+      vpcId            = "${data.aws_vpc.selected.id}"
+      subnetId         = "${aws_subnet.private.1.id}"
+      instanceType     = "${var.worker_type}"
+      diskSize         = 100
+      sshPublicKeys    = ["${aws_key_pair.deployer.public_key}"]
+      replicas         = 3
+      operatingSystem  = "ubuntu"
```
xmudrii: Should we consider making this a variable?

kron4eg: control plain is ubuntu, naturally workers are ubuntu too

examples/terraform/aws-private/output.tf
kron4eg: I don't see why.

examples/terraform/aws-private/output.tf
kron4eg: changed to 1

examples/terraform/aws-private/output.tf
xmudrii: Control plane is Ubuntu by default, but the OS can be changed. Therefore, this should be made a variable or we should document this, so users know what they should replace in case of the OS change. A follow-up for this is okay.

examples/terraform/aws-private/output.tf
xmudrii: To avoid situations such as in #349, we should document what is where and what should be changed. A follow-up at some point is okay.

xmudrii:
/lgtm
/approve

Looks good :100:

PR #364, opened 2019-04-22
Deploy external CCMs

### LINKED ISSUE #255: Deploy external Cloud Controller Manager (CCM) if Provider.External is set
Currently if Provider.External is set to true it's user's responsibility to deploy the external CCM. We should consider deploying the external CCM ourselves because currently we're running into many edge cases:
* if `--cloud-provider=external` is set on Kubelet all nodes (including control plane nodes) will stay uninitialized until CCM is not deployed,
* this prevents components to be deployed, including system components such as CoreDNS,
* without CoreDNS, DNS functionality will not be available in the cluster and we can't create worker nodes,
* for cloud providers without in-tree CCMs, not all information about nodes is populated, for example IP addresses

When deploying CCM ourselves we deny user from choosing what CCM version they want to run, however we're gaining more control and we can ensure that such CCM version is working on KubeOne clusters.

At this point, this issue is blocked until machine-controller doesn't implement features needed for the external CCM to work as expected.

/kind feature
/lifecycle blocked

UPDATE:
Need to reference machine-controller credentials secret from CCM deployment.

kron4eg:
Fixes #255

There will be a follow up PR, once this and #362 merged that will introduce `external: true` flag to DO and hetzner

```release-note
Deploy Hetzner CCM when provider.name: 'hetzner' and provider.external: true
Deploy DigitalOcean CCM when provider.name: 'digitalocean' and provider.external: true
```

### DIFF
--- a/pkg/templates/externalccm/ccm.go
+++ b/pkg/templates/externalccm/ccm.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2019 The KubeOne Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package externalccm
 
 import (
--- a/pkg/templates/externalccm/ccm.go
+++ b/pkg/templates/externalccm/ccm.go
@@ -2,16 +2,21 @@ package externalccm
 
 import (
 	"context"
+	"strings"
+
+	"github.com/Masterminds/semver"
+	"github.com/pkg/errors"
 
 	"github.com/kubermatic/kubeone/pkg/config"
 	"github.com/kubermatic/kubeone/pkg/util"
 
+	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	dynclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
-// Ensure external CCM when
+// Ensure external CCM deployen if Provider.External
 func Ensure(ctx *util.Context) error {
 	if !ctx.Cluster.Provider.External {
 		return nil
@@ -22,6 +27,8 @@ func Ensure(ctx *util.Context) error {
 	switch ctx.Cluster.Provider.Name {
 	case config.ProviderNameHetzner:
 		return ensureHetzner(ctx)
+	case config.ProviderNameDigitalOcean:
+		return ensureDigitalOcean(ctx)
 	default:
 		ctx.Logger.Infof("External CCM for %q not yet supported, skipping", ctx.Cluster.Provider.Name)
 		return nil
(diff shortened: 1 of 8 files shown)


pkg/templates/externalccm/ccm.go
```diff
@@ -0,0 +1,93 @@
...
+
+	ctx.Logger.Info("Ensure external CCM is up to date")
+
+	switch ctx.Cluster.Provider.Name {
+	case config.ProviderNameHetzner:
+		return ensureHetzner(ctx)
+	case config.ProviderNameDigitalOcean:
+		return ensureDigitalOcean(ctx)
+	default:
+		ctx.Logger.Infof("External CCM for %q not yet supported, skipping", ctx.Cluster.Provider.Name)
```
xmudrii: I'd add a message that the user should deploy it manually, e.g. `External CCM for %q not yet supported, please deploy CCM manually`.

pkg/templates/externalccm/ccm.go
```diff
@@ -0,0 +1,93 @@
...
+			return errors.New("unable to grab CCM image version")
+		}
+
+		existing, err := semver.NewVersion(imageSpec[1])
+		if err != nil {
+			return errors.Wrap(err, "failed to parse deployed CCM version")
+		}
+
+		if !want.Check(existing) {
+			return errors.New("newer version deployed, skipping")
```
xmudrii: Why this returns an error instead of logging? Having a newer version deployed is okay IMO, but it should be logged, so the user knows about it.

pkg/templates/externalccm/digitalocean.go
```diff
@@ -0,0 +1,262 @@
...
+								"--cloud-provider=digitalocean",
+								"--leader-elect=false",
+							},
+							Env: []corev1.EnvVar{
+								{
+									Name: "DO_ACCESS_TOKEN",
+									ValueFrom: &corev1.EnvVarSource{
+										SecretKeyRef: &corev1.SecretKeySelector{
+											LocalObjectReference: corev1.LocalObjectReference{
+												Name: machinecontroller.MachineControllerCredentialsSecretName,
```
xmudrii: This is not going to work if the `machine-controller` isn't deployed. We should export that logic from the `machine-controller` and deploy it regardless.

pkg/templates/externalccm/ccm.go
kron4eg: errors from here are handled (and loggeg) at upper level, in caller

kron4eg: And returning an error from this callback is way to signal to `controllerutil.CreateOrUpdate` that create/update should not happen.

pkg/templates/externalccm/ccm.go
kron4eg: `please deploy CCM manually` is implied, since we can't install them 🤷‍♂️

pkg/templates/externalccm/digitalocean.go
kron4eg: We will handle this in separate PR: https://github.com/kubermatic/kubeone/issues/366

xmudrii: I'll take care of it with the API changes.

pkg/templates/externalccm/ccm.go
xmudrii: I prefer being explicit in the user-facing error messages.

kron4eg: I don't want to run pipeline again just for "notice"

xmudrii: Considering you have to rebase the PR, you can fix this. :wink:

kron4eg: too late, already rebased :P

PR #370, opened 2019-04-23
Parse all supported digitalocean params

### LINKED ISSUE #176: Make Terraform automation for workers able to source booleans from Terraform variables
In #153 we added initial support for creating DigitalOcean worker nodes using KubeOne. The initial implementation only supports sourcing region and droplet size from Terraform variables because `godo` expects boolean for variables such as private networking and monitoring.

Currently, as per [Terraform documentation](https://www.terraform.io/docs/configuration/variables.html#booleans), there is no proper support for booleans, so for now we have decided not to source those variables.

We want to find a good way to source booleans, but also to enforce backwards-compatibility and type-safety.

**Acceptance criteria:**

* Terraform automation for workers is able to source booleans from Terraform variables
* DigitalOcean worker creation is sourcing private networking, monitoring, backups and IPv6 form Terraform

kron4eg:
* cleanup testdata cluster configs
* remove what's already providede by terraform
* remove default options
* speed-up DO tests (by avoiding upgrade on boot)

```release-note
Digitalocean params are fully supported in terraform config
```

Fixes #176

### DIFF
--- a/examples/terraform/digitalocean/output.tf
+++ b/examples/terraform/digitalocean/output.tf
@@ -41,10 +41,19 @@ output "kubeone_workers" {
     # corresponding (by name) worker definition
     fra1-1 = {
       replicas        = 3
-      operatingSystem = "ubuntu"
-      droplet_size    = "${var.droplet_size}"
-      region          = "${var.region}"
       sshPublicKeys   = ["${digitalocean_ssh_key.deployer.public_key}"]
+      operatingSystem = "ubuntu"
+
+      operatingSystemSpec = {
+        distUpgradeOnBoot = false
+      }
+
+      region             = "${var.region}"
+      size               = "${var.droplet_size}"
+      private_networking = true
+      backups            = false
+      ipv6               = false
+      monitoring         = false
     }
   }
 }
--- a/pkg/terraform/config.go
+++ b/pkg/terraform/config.go
@@ -77,8 +77,13 @@ type awsWorkerConfig struct {
 
 // TODO: Add option for sourcing bool values (private networking, monitoring...)
 type doWorkerConfig struct {
-	DropletSize string `json:"droplet_size"`
-	Region      string `json:"region"`
+	Region            string   `json:"region"`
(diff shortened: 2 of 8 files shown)


kron4eg:
What boolean?
I specifically disabled distUpgadeOnBoot to speed up nodes join

xmudrii:
@kron4eg Private networking is a boolean, therefore it's false by default. Without private networking the cluster may not work correctly. Operator will mostly like always want it enabled.

examples/terraform/digitalocean/output.tf
```diff
@@ -41,10 +41,19 @@ output "kubeone_workers" {
     # corresponding (by name) worker definition
     fra1-1 = {
       replicas        = 3
-      operatingSystem = "ubuntu"
-      droplet_size    = "${var.droplet_size}"
-      region          = "${var.region}"
       sshPublicKeys   = ["${digitalocean_ssh_key.deployer.public_key}"]
+      operatingSystem = "ubuntu"
+
+      operatingSystemSpec = {
```
xmudrii: We should try to be as consistent as possible when doing such changes. Please apply a similar change to other Terraform scripts as well.

xmudrii:
This should fix #176

examples/terraform/digitalocean/output.tf
kron4eg: in followup PR, this one was sporadic experiment.

xmudrii: Please create a follow-up issue and then I'll lgtm.

kron4eg: we already have one: #176

xmudrii: This one is about making KubeOne able to parse Terraform booleans and not about making scripts consistent. With this PR merged that issue should be closed. I created #371 instead.

kron4eg:
/hold cancel

@xmudrii I've added external: true and rebased against master, PTAL.

PR #390, opened 2019-04-28
Migrate to the KubeOneCluster API

xmudrii:

Migrate to the KubeOneCluster API implemented in #353 and proposed in [the following proposal](https://github.com/kubermatic/kubeone/blob/master/docs/proposals/20190409-kubeonecluster-api.md).

This PR brings **breaking changes** as some fields are removed or renamed. The new API is a Kubernetes-style API and supports API versioning, which allows us to make changes to the API without affecting the end users and breaking the backwards compatibility.

The new API is called `KubeOneCluster`, the API group is `kubeone.io` and the initial version is `v1alpha1`.

Once this PR is merged it will **not** be possible to create KubeOne clusters using the old configuration file and API. The following changes must be made in order to migrate to the new API:

* `apiVersion` and `kind` must be added at the top of the configuration file:
```yaml
kind: KubeOneCluster
```
* The following fields at the root level are renamed, so they follow the Kubernetes-style of naming fields and structures:
  * `apiserver` -> `apiEndpoints`
  * `provider` -> `cloudProvider`
  * `network` -> `clusterNetwork`
  * `machine_controller` -> `machineController`
* All `hosts` fields must be changed to the camel case:
  * `public_address` -> `publicAddress`
  * `private_address` -> `privateAddress`
  * `ssh_port` -> `sshPort`
  * `ssh_username` -> `sshUsername`
  * `ssh_private_key_file` -> `sshPrivateKeyFile`
  * `ssh_agent_socket` -> `sshAgentSocket`
* The `apiserver` structure is replaced with the `apiEndpoint` structure which contains the following fields:
  * `host` - address or hostname of the API endpoint (by default load balancer DNS)
  * `port` - port of the API (by default `6443`)
* All `cloudProvider` (prev. `provider`) fields must be changed to the camel case:
  * `cloud_config` -> `cloudConfig`
* The `network` structure is renamed to `clusterNetwork`, all existing fields are renamed to follow the camel case and a new field is added:
  * `pod_subnet` -> `podSubnet`
  * `service_subnet` -> `serviceSubnet`
  * `node_port_range` -> `nodePortRange`
  * [NEW] `serviceDomainName` -> isn't used currently
* `http_proxy` and `https_proxy` fields in the `proxy` structure are renamed to drop the `_proxy` suffix and `no_proxy` is renamed to follow the camel case:
  * `http_proxy` -> `http`
  * `https_proxy` -> `https`
  * `no_proxy` -> `noProxy`
* The `credentials` field from `machineController` structure (prev. `machine_controller`) is moved to the root level.
* All fields in the `features` structure are renamed to follow the camel case:
  * `pod_security_policy` -> `podSecurityPolicy`
  * `dynamic_audit_log` -> `dynamicAuditLog`
  * `metrics_server` -> `metricsServer`
  * `openid_connect` -> `openidConnect`
* All `openidConnect` fields are renamed to follow the camel case.
* `workers.Config` is renamed to `workers.providerSpec`.

Changes affecting using KubeOne as a Go library:

* Structures are renamed to follow the camel case and some structures are changed or removed (see above points for more details),
* `WorkerConfig.Config.CloudProviderSpec` and `WorkerConfig.Config.OperatingSystemSpec` are taking `json.RawMessage` instead of `map[string]interface{}`,
* All fields in the `Features` structure are now pointers. All fields for enabling the feature are called `Enable` and are type of bool (prev. pointer on bool),
* `MachineController` is now a pointer of `MachineControllerConfig` (prev. non-pointer `MachineControllerConfig`),
* `MachineController.Deploy` is now bool instead of pointer on bool.

The following changes should be considered before merging the PR:

* [x] Make `serviceDomainName` working.
* [ ] Support more than one API endpoint (a follow-up).
* [x] Is `openidConnect` a correct JSON tag or should it be `openIdConnect`?
* [x] Fix a typo in `apiEndpoint` (should be `apiEndpoints`).
* [x] Make sure that the cluster is updated after the hostname or operating system is changed.
* [x] ~Remove the old API (`pkg/config`) (a follow-up).~
* [ ] Build a migration tool for migrating from the old API to the new KubeOneCluster API (a follow-up).
* [ ] Update documentation and examples (a follow-up).

xref #259
The issue will be closed once the old API is removed. 

**Documentation**:

Documentation is up to be updated.

**Release note**:
```release-note
[BREAKING CHANGES] Migrate to the new KubeOneCluster API. Disable the old API.
```

/assign @kron4eg 
cc @scheeles 

/hold
until we don't agree on the final API structure

### DIFF
--- a/config.yaml.dist
+++ b/config.yaml.dist
@@ -23,30 +23,28 @@ versions:
   kubernetes: "1.14.1"
 
 clusterNetwork:
-  # the subnet used for pods (flannel);
-  # leave it empty for default: 10.244.0.0/16
+  # the subnet used for pods (default: 10.244.0.0/16)
   podSubnet: ""
-  # the subnet used for services;
-  # leave it empty for default: 10.96.0.0/12
+  # the subnet used for services (default: 10.96.0.0/12)
   serviceSubnet: ""
-  # a nodePort range to reserve for services;
-  # leave it empty for default: 30000-32767
+  # the domain name used for services (default: cluster.local)
+  serviceDomainName: ""
+  # a nodePort range to reserve for services (default: 30000-32767)
   nodePortRange: ""
 
-provider:
+cloudProvider:
   # Supported cloud provider names:
   # * aws
   # * digitalocean
   # * hetzner
-  # * none
   # * openstack
   # * vsphere
+  # * none
   name: ""
-  # Set kubelet flag --cloud-provider=external, to be used with external
-  # Cloud Controller Managers (CCM).
+  # Set the kubelet flag '--cloud-provider=external' and deploy the external CCM for supported providers
   external: false
-  # Path to file that will be uploaded and used as custom --cloud-config file.
-  cloud_config: ""
+  # Path to file that will be uploaded and used as custom '--cloud-config' file.
+  cloudConfig: ""
 
 features:
   # Enables PodSecurityPolicy admission plugin in API server, as well as creates
@@ -122,17 +120,19 @@ features:
 #   sshAgentSocket: 'env:SSH_AUTH_SOCK'
 
 # The API server can also be overwritten by Terraform. Provide the
-# external address of your load balancer or the public address of
-# your first node.
+# external address of your load balancer or the public addresses of
+# your control plane nodes.
 # apiEndpoints:
-#   - host: '1.2.3.4'
-#     port: 6443
+# - host: '1.2.3.4'
+#   port: 6443
 
 # If the cluster runs on bare metal or an unsupported cloud provider,
 # you can disable the machine-controller deployment entirely. In this
 # case, anything you configure in your "workers" sections is ignored.
-#machineController:
+# machineController:
 #  deploy: false
+#  # Defines for what provider the machine-controller will be configured (defaults to cloudProvider.Name)
+#  provider: ""
 
 # Proxy is used to configure HTTP_PROXY, HTTPS_PROXY and NO_PROXY
 # for Docker daemon and kubelet, and to be used when provisioning cluster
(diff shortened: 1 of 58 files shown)


xmudrii:
@kron4eg Removed the hack and committed the appropriate fix. ServiceDomainName is now usable as well.

xmudrii:
/hold cancel
Issues will be created for other tasks once the PR is merged

config.yaml.dist
```diff
@@ -15,24 +15,23 @@
...
   # leave it empty for default: 10.96.0.0/12
-  service_subnet: ""
-
+  serviceSubnet: ""
   # a nodePort range to reserve for services;
   # leave it empty for default: 30000-32767
-  node_port_range: ""
+  nodePortRange: ""
 
 provider:
```
kron4eg: Did you forgot to update to `cloudProvider`? Also, why `provider` is bad?

config.yaml.dist
```diff
@@ -43,119 +42,105 @@ provider:
...
-#   ssh_agent_socket: 'env:SSH_AUTH_SOCK'
+#   sshPrivateKeyFile: '/home/me/.ssh/id_rsa'
+#   sshAgentSocket: 'env:SSH_AUTH_SOCK'
 
 # The API server can also be overwritten by Terraform. Provide the
 # external address of your load balancer or the public address of
 # your first node.
-# apiserver:
-#   address: '1.2.3.4'
+# apiEndpoints:
```
kron4eg: why is this an array if endpoint is a single point. And in kubeadm api it's also [a single string](https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/apis/kubeadm/v1beta1/types.go#L83)

config.yaml.dist
xmudrii: I forgot to update some fields including `cloudProvider`. Pushed a commit that updates everything needed.

The `cloudProvider` change was based on https://github.com/kubermatic/kubeone/pull/353#discussion_r276633959 and some discussion we had on the stand-up meeting.

One of the reasons I found `provider` bad is because it's mostly about the cloud provider features. `provider` wasn't descriptive enough and the idea was to rename it to something more clear.

config.yaml.dist
xmudrii: That was some initial effort to try to adopt Cluster-API or align with it. I decided not to do that, but `apiEndpoints` change was left on purpose.

The decision to rename it is because `apiEndpoints` is more descriptive than `apiserver`. It tells the user for what exactly is it used for.

As of why is this an array—is it always going to be a single point? For example, what if we don't have a load balancer? Can the user access all the nodes or just the first one? I believe that depends on the setup. For users using KubeOne with KubeOne CLI, that mostly likes means nothing, but if someone is using KubeOne as a Go library it could be helpful.

kron4eg: >is it always going to be a single point

yes, that's how [kubeconfig is arranged](https://github.com/kubernetes/client-go/blob/master/rest/config.go#L57) and the same is kubeadm.

xmudrii: After discussion on the daily meeting, we decided to use only one API endpoint. Pushed a commit that fixes this.

xmudrii:
One DNS flake and one DO API flake.
/retest

PR #414, opened 2019-05-07
Automatic LB provision for Hetzner

### LINKED ISSUE #409: Deploy and configure load balancer instance on Hetzner
As Hetzner doesn't have load balancer as a service, we should deploy and configure load balancer like we do for Packet and OpenStack.

kron4eg:
Fixes #409

```release-note
example of LB provisioned by terraform for hetzner
```


xmudrii:
@kron4eg Is this is something critical-urgent for v0.6.0? I'd postpone it for v0.7.0 as we don't have Hetzner E2E tests.

kron4eg:

@xmudrii nothing bad would happen if we include it into 0.6

Issue #254, opened 2019-03-12
Add a command for printing an example configuration file


xmudrii:
From the UX standpoint it might make sense to provide a CLI command that would print an example configuration file. The example configuration file could be `config.yaml.dist` for that version as it's well-documented and a good starting point.

This can be especially helpful for users using KubeOne from releases, as they don't have access to the sources. The configuration file in this case can only be found on GitHub. This makes harder to use the tool and is error prone as user must ensure to find the correct configuration file for such KubeOne version.

PR #415, opened 2019-05-07
Extend 'config print' command with additional options

xmudrii:

This is an **experimental** approach to the `config print` command. I tested and confirmed it works well. The `config print` command is now extended to support most of the options, so someone can build a configuration file without browsing GitHub and searching for the `config.yaml.dist` manifest.

The idea behind this is to improve the user experience, so the user can work with our API without having to browse GitHub to find `config.yaml.dist` or to browse the API types. Various tools have different approaches to this, so I can believe we can find a good one too.

However, I have several concerns regarding this topic:

* All those flags can potentially confuse the user. It might make more sense to leave `hosts` and `enableOpenIDConnect` (so we have all features supported) and drop others as they're mostly like not-commonly used.
  * Instead of all flags we can put the most common ones and have a `set` flag that changes some property not covered by flags. That way the command is not overloaded with flags, but it's still possible to build usable configuration file.
  * Negative side of the `set` flag is that you still have to know the API.
* Instead of a single `config print` command we can split them to two commands:
  * A `config create` command that works like one in this PR
  * A `config print` command that has two "modes":
    * Print full `config.yaml.dist`
    * Print just example manifest like we already have without this PR
    * In this case printing full `config.yaml.dist` might overlap with `config create`, so potentially we can drop one.
* Another problem is maintainability—we need to make sure that we incorporate every API change here. 
  * Using `yamled` is nice as we get a nice looking manifest without all the empty fields, but there is no strongly enforced type safety.
  * Using `kyaml.UnmarshalStrict` ensures that it'll not print invalid manifest, but in that case, there should be some unit test cases in the CI to check is it working.

xref #254 

**Release note**:
```release-note
Extend 'config print' command with additional options
```

/assign @kron4eg

### DIFF
--- a/pkg/cmd/config.go
+++ b/pkg/cmd/config.go
@@ -18,11 +18,11 @@ package cmd
 
 import (
 	"bytes"
+	"fmt"
 	"os"
 	"strconv"
 	"strings"
 	"text/template"
-	"fmt"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
--- a/pkg/cmd/config.go
+++ b/pkg/cmd/config.go
@@ -194,7 +194,9 @@ func runPrint(printOptions *printOptions) error {
 
 	// Hosts
 	if len(printOptions.Hosts) != 0 {
-		parseHosts(cfg, printOptions.Hosts)
+		if err := parseHosts(cfg, printOptions.Hosts); err != nil {
+			return errors.Wrap(err, "unable to parse provided hosts")
+		}
 	}
 
 	// API endpoint
--- a/pkg/cmd/config.go
+++ b/pkg/cmd/config.go
@@ -21,14 +21,15 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"text/template"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	yaml "gopkg.in/yaml.v2"
 
-	kubeoneapi "github.com/kubermatic/kubeone/pkg/apis/kubeone"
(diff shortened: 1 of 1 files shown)


kron4eg:
@xmudrii I just noticed that provider is not getting validated

```
$ kubeone config print -p unknown_provider
kind: KubeOneCluster
versions:
  kubernetes: 1.14.1
cloudProvider:
  name: unknown_provider
```

pkg/cmd/config.go
```diff
@@ -214,3 +370,206 @@ func validateAndPrintConfig(cfgYaml interface{}) error {
 
 	return nil
 }
+
+const exampleManifest = `
+# Copyright 2019 The KubeOne Authors.
```
kron4eg: Is there a need for boilerplate here?

pkg/cmd/config.go
```diff
@@ -214,3 +370,206 @@ func validateAndPrintConfig(cfgYaml interface{}) error {
...
+  # * aws
+  # * digitalocean
+  # * hetzner
+  # * none
+  # * openstack
+  # * packet
+  # * vsphere
+  name: "{{ .CloudProviderName }}"
+  # Set the kubelet flag '--cloud-provider=external' and deploy the external CCM for supported providers
+  external: false
```
kron4eg: external flag is statically `false`, we can do better.

xmudrii:
/override pull-kubeone-e2e-aws-conformance-1.13.5
/override pull-kubeone-e2e-digitalocean-conformance-1.13.5
/override pull-kubeone-e2e-digitalocean-conformance-1.14.1

PR #422, opened 2019-05-08
Update README to the release v0.6.0

xmudrii:

I've forgotten to update the README when doing a release, so here's a PR doing that now. :sweat_smile: 

**Release note**:
```release-note
NONE
```

/assign @kron4eg


xmudrii:
@kron4eg This also deprecates the release v0.4.1! Considering it's quite old now in terms of features and changes, I think we should deprecate it.

PR #424, opened 2019-05-09
Fix metrics-server for all providers

### LINKED ISSUE #423: metrics-server failing to grab node info
**What happened**:
Flag `--kubelet-preferred-address-types` is not provided and by default it's `[Hostname,InternalDNS,InternalIP,ExternalDNS,ExternalIP]`. This doesn't work on all supported providers. Examples:

* hetzner: need to have `ExternalIP` as other Address Types are unreliable
* digitalocean: `InternalIP,ExternalIP`

In most (but not all) cases InternalIP will work.

`--kubelet-preferred-address-types` should be provided and its value should depend on provider

KubeOne version: 0.6.0

kron4eg:
Fixes #423 

```release-note
Fix metrics-server for all providers
```

### DIFF
--- a/pkg/templates/metricsserver/deployment.go
+++ b/pkg/templates/metricsserver/deployment.go
@@ -208,6 +208,7 @@ func metricsServerDeployment() *appsv1.Deployment {
 							ImagePullPolicy: corev1.PullIfNotPresent,
 							Args: []string{
 								"--kubelet-insecure-tls",
+								"--kubelet-preferred-address-types=InternalIP,InternalDNS,ExternalDNS,ExternalIP",
 							},
 							VolumeMounts: []corev1.VolumeMount{
 								{


pkg/templates/metricsserver/deployment.go
```diff
@@ -208,6 +208,7 @@ func metricsServerDeployment() *appsv1.Deployment {
 							ImagePullPolicy: corev1.PullIfNotPresent,
 							Args: []string{
 								"--kubelet-insecure-tls",
+								"--kubelet-preferred-address-types=InternalIP,InternalDNS,ExternalDNS,ExternalIP",
```
kron4eg: metrics-server will use first accessible address type.

on hetzner it will be `ExternalIP`, others do support `InternalIP`. With provider `none` real external IP will be indicated as `InternalIP` so will work ether way.

PR #431, opened 2019-05-10
Improve aws terraform configurations

### LINKED ISSUE #429: Ensure AWS resources in terraform examples are tagged
Currently not all AWS resources are tagged with special `kubernetes.io/cluster/${cluster.name} = ""` tag, that should be fixed

kron4eg:
Fixes #429

```release-note
Improve aws terraform:
* apply kubernetes tags
* allow fixating AMI
```

### DIFF
--- a/examples/terraform/aws/output.tf
+++ b/examples/terraform/aws/output.tf
@@ -45,9 +45,9 @@ output "kubeone_workers" {
   value = {
     # following outputs will be parsed by kubeone and automatically merged into
     # corresponding (by name) worker definition
-    fra1-a = {
+    pool-a = {
       region           = "${var.aws_region}"
-      ami              = "${data.aws_ami.ubuntu.id}"
+      ami              = "${local.ami}"
       availabilityZone = "${local.az_a}"
       instanceProfile  = "${aws_iam_instance_profile.profile.name}"
       securityGroupIDs = ["${aws_security_group.common.id}"]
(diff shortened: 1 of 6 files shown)


examples/terraform/aws/variables.tf
```diff
@@ -19,7 +19,7 @@ variable "cluster_name" {
 }
 
 variable "aws_region" {
-  default     = "eu-west-3"
+  default     = "eu-central-1"
```
xmudrii: Why we changed back to `eu-central-1`? I think we had less number of flakes on `eu-west-3`.

examples/terraform/aws/output.tf
```diff
@@ -45,9 +45,9 @@ output "kubeone_workers" {
   value = {
     # following outputs will be parsed by kubeone and automatically merged into
     # corresponding (by name) worker definition
-    fra1-a = {
+    pool-a = {
```
xmudrii: Can this be potentially backwards incompatible? What if the user provisioned infrastructure using old scripts? Also, if we're doing this change, we should do it for all providers.

kron4eg: We don't cover *example* terraform configurations with backward compatibility, and even if we do, we should cancel it. Those are only example show-cases, nothing more.

examples/terraform/aws/variables.tf
kron4eg: The point was that automated clean-up of leftovers job is working in `eu-central-1`, but OK, reverted back to Paris.

PR #450, opened 2019-05-15
Delete worker nodes using dynamic client and add an option to remove Kubernetes packages when resetting the cluster

### LINKED ISSUE #397: Reset doesn't remove Kubernetes binaries
**What happened**:

`kubeone reset` doesn't remove Kubernetes binaries (kubeadm, kubelet..).

Kubernetes binaries are removed from the instances.

Run `kubeone reset` and verify are `kubeadm` and `kubelet` installed.

This makes it impossible to install the older Kubernetes version, as the `kubelet` binary is newer.

### LINKED ISSUE #447: error: the server doesn't have a resource type "machinedeployment"
**What happened**:
I tried to delete a cluster with the following config:

```
kind: KubeOneCluster
name: demo
versions:
  kubernetes: "1.13.2"
cloudProvider:
  name: "none"
hosts:
- publicAddress: '1.2.3.4'
  privateAddress: '172.18.0.1'
  sshPort: 22 # can be left out if using the default (22)
  sshUsername: ubuntu
machineController:
  deploy: false
  provider: "none"
```

This failed with the following issue:
```
~/go/bin/kubeone reset config.yaml -v
INFO[08:22:07 CEST] Resetting kubeadm…
INFO[08:22:07 CEST] Destroying worker nodes…                      node=172.18.0.1
+ kubectl cluster-info
+ 1> /dev/null
+ kubectl annotate --all --overwrite node 'kubermatic.io/skip-eviction=true'
node/master0 annotated
+ kubectl delete machinedeployment -n kube-system --all
error: the server doesn't have a resource type "machinedeployment"
Error: failed to exec command: export "PATH=$PATH:/sbin:/usr/local/bin:/opt/bin"

set -xeu pipefail


if kubectl cluster-info > /dev/null; then
  kubectl annotate --all --overwrite node kubermatic.io/skip-eviction=true
  kubectl delete machinedeployment -n "kube-system" --all
  kubectl delete machineset -n "kube-system" --all
  kubectl delete machine -n "kube-system" --all

  for try in {1..30}; do
    if kubectl get machine -n "kube-system" 2>&1 | grep -q  'No resources found.'; then
      exit 0
    fi
    sleep 10s
  done

  echo "Error: Couldn't delete all machines!"
  exit 1
fi
```

Deletion proceeds.

Try to delete a cluster when the following is set in config.yaml
```
machineController:
  deploy: false
  provider: "none"
```

KubeOne version (kubeone version): v0.6.0-6-g6407139
Operating system: CentOS Linux release 7.6.1810 (Core)
Provider you're deploying cluster on: Bare-metal
Operating system you're deploying on: CentOS Linux release 7.6.1810 (Core)

xmudrii:

This PR refactors the `reset` command to use dynamic client instead of kubectl over SSH tunnel for deleting worker nodes.

Fixes #447 
Fixes #397

**Release note**:
```release-note
Fix reset failures if there are no Machine objects or Cluster-API CRDs are not deployed
Add the --remove-binaries flag which removes Kubernetes binaries after resetting the cluster
```

/assign @kron4eg

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,6 @@ test:
 dep:
 	dep ensure -v
 
-
 .PHONY: licence-check
 licence-check:
 	wwhrd check
@@ -67,7 +66,7 @@ docker-make-install:
 		make install
 
 .PHONY: e2e_test
-e2e_test:
+e2e_test: build lint test dep
 	./hack/run_ci_e2e_test.sh
 
 dist/kubeone: $(shell find . -name '*.go')
--- a/pkg/cmd/reset.go
+++ b/pkg/cmd/reset.go
@@ -29,7 +29,7 @@ type resetOptions struct {
 	globalOptions
 	Manifest       string
 	DestroyWorkers bool
-	RemovePackages bool
+	RemoveBinaries bool
 }
 
 // resetCmd setups reset command
@@ -64,7 +64,7 @@ It's possible to source information about hosts from Terraform output, using the
 	}
 
 	cmd.Flags().BoolVarP(&ropts.DestroyWorkers, "destroy-workers", "", true, "destroy all worker machines before resetting the cluster")
-	cmd.Flags().BoolVarP(&ropts.RemovePackages, "remove-packages", "", false, "remove kubernetes packages after resetting the cluster")
+	cmd.Flags().BoolVarP(&ropts.RemoveBinaries, "remove-binaries", "", false, "remove kubernetes binaries after resetting the cluster")
 
 	return cmd
 }
@@ -83,7 +83,7 @@ func runReset(logger *logrus.Logger, resetOptions *resetOptions) error {
 	options := &installer.Options{
 		Verbose:        resetOptions.Verbose,
 		DestroyWorkers: resetOptions.DestroyWorkers,
-		RemovePackages: resetOptions.RemovePackages,
+		RemoveBinaries: resetOptions.RemoveBinaries,
 	}
 
 	return installer.NewInstaller(cluster, logger).Reset(options)
--- a/pkg/cmd/reset.go
+++ b/pkg/cmd/reset.go
@@ -29,6 +29,7 @@ type resetOptions struct {
 	globalOptions
 	Manifest       string
 	DestroyWorkers bool
+	RemovePackages bool
 }
 
 // resetCmd setups reset command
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -67,7 +66,7 @@ docker-make-install:
 		make install
 
 .PHONY: e2e_test
-e2e_test:
+e2e_test: build lint test dep
```
kron4eg: `s/dep/check-dependencies/` — we'd like to know when deps are updated but not checked-in
also add `licence-check` to this step.

xmudrii:
This is a `yamled` flake.
/retest

pkg/templates/machinecontroller/helper.go
```diff
@@ -75,3 +79,80 @@ func WaitReady(ctx *util.Context) error {
...
+	nodes := &corev1.NodeList{}
+	if err := ctx.DynamicClient.List(bgCtx, &dynclient.ListOptions{}, nodes); err != nil {
+		return errors.Wrap(err, "unable to list nodes")
+	}
+	for i := range nodes.Items {
+		if nodes.Items[i].Annotations == nil {
+			nodes.Items[i].Annotations = map[string]string{}
+		}
+		nodes.Items[i].Annotations["kubermatic.io/skip-eviction"] = "true"
+		if err := ctx.DynamicClient.Update(bgCtx, &nodes.Items[i]); err != nil {
```
kron4eg: This is part should contain retry logic, as it's likely that you'll hit a conflict. There is `k8s.io/client-go/util/retry` helper package for retries.

Makefile
kron4eg: OK, probably no `licence-check`, sorry, it's different image.

pkg/templates/machinecontroller/helper.go
xmudrii: Check now. :)

Makefile
xmudrii: Regarding changing `dep` to checking `check-dependencies`, it would require a change to the presubmit, so let's leave it for a follow-up.

xmudrii:
/hold
I'll take care of #397 as well

xmudrii:
/hold
This is a second flake related to reset not being able to delete the MachineSet. It should be fixed before merging the PR.

pkg/cmd/reset.go
```diff
@@ -63,6 +64,7 @@ It's possible to source information about hosts from Terraform output, using the
 	}
 
 	cmd.Flags().BoolVarP(&ropts.DestroyWorkers, "destroy-workers", "", true, "destroy all worker machines before resetting the cluster")
+	cmd.Flags().BoolVarP(&ropts.RemovePackages, "remove-packages", "", false, "remove kubernetes packages after resetting the cluster")
```
xmudrii: I'm concerned about the flag name. Packages may be a bad word here because we don't use package management solution on all platforms. For example, on CoreOS we download and install binaries manually. Technically, `remove-binaries` could work better. Any opinions?

kron4eg: `remove-binaries` sgtm

xmudrii: Done. :100:

PR #486, opened 2019-06-11
kubeadm.KubeADM interface

### LINKED ISSUE #505: Revisit default admission plugins
It seems like Kubernetes 1.15 has different set of enabled by default admission plugins. We should revisit this and update if needed.

/kind feature
/priority critical-urgent
Potentially 0.9 release blocking

kron4eg:
A way to abstract different kubeadm API version.

Fixes #505 

```release-note
NONE
```


pkg/templates/kubeadm/kubeadmv1beta1.go
```diff
@@ -17,31 +17,28 @@ limitations under the License.
...
-	//TODO: Change KubernetesToYAML to accept runtime.Object instead of empty interface
-	var kubernetesToYAMLInput []interface{}
-	for _, config := range configs {
-		kubernetesToYAMLInput = append(kubernetesToYAMLInput, interface{}(config))
-	}
-	return templates.KubernetesToYAML(kubernetesToYAMLInput)
+	return templates.KubernetesToYAML(config)
+}
+
+func (*kubeadmv1beta1) UpgradeLeaderCMD() string {
```
xmudrii: I personally find `CMD` suffix ugly. `UpgradeLeaderCommand` would probably look much better and it's clear what it is about.

pkg/templates/kubeadm/provider.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+var (
+	v13x = mustParseConstraint("1.13.x")
+	v14x = mustParseConstraint("1.14.x")
+	v15x = mustParseConstraint("1.15.x")
+)
+
+// KubeADM interface abstract differences between different kubeadm versions
+type KubeADM interface {
```
xmudrii: Why is this called `KubeADM` and not just `Kubeadm`?

pkg/templates/kubeadm/provider.go
```diff
@@ -0,0 +1,66 @@
...
+	UpgradeFollowerCMD() string
+}
+
+// New constructor
+func New(ver string) (KubeADM, error) {
+	sver, err := semver.NewVersion(ver)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to parse version")
+	}
+	switch {
```
xmudrii: I'd like to enforce this check at the validation level too. If you provide not supported version, e.g. `1.16`, it's going to pass validation but fail on install.

pkg/templates/kubeadm/provider.go
```diff
@@ -0,0 +1,66 @@
...
+	case v13x.Check(sver):
+		return &kubeadmv1beta1{}, nil
+	case v14x.Check(sver):
+		return &kubeadmv1beta1{}, nil
+	case v15x.Check(sver):
+		return &kubeadmv1beta2{}, nil
+	}
+
+	// By default use latest known kubeadm API version
+	return &kubeadmv1beta2{}, nil
```
xmudrii: Related to the previous comment: should we strong enforce supported versions or not? For example, if `1.16` introduces any breaking change in kubeadm (like drop `--experimental`), that would render KubeOne unusable.

pkg/templates/util.go
```diff
@@ -40,18 +41,15 @@ func KubernetesToYAML(data []interface{}) (string, error) {
 			err         error
 		)
 
-		if str, ok := item.(string); ok {
-			encodedItem = []byte(strings.TrimSpace(str))
-		} else {
-			encodedItem, err = yaml.Marshal(item)
-		}
-
+		encodedItem, err = yaml.Marshal(item)
```
xmudrii: I don't understand what was the old code doing? Can you explain to me what's the difference here?

kron4eg: because input was `[]inetrface`, under the hook it can be anything including `string`, which was type-asserted. Since string doesn't need to be marshalred — no marshaling actually happened. So, basically leftover from old-times when we were running on static yaml files.

xmudrii: Okay, that makes sense. Thanks for explaining!

pkg/templates/kubeadm/provider.go
kron4eg: We don't use `--experimental` btw ;)

1.14 is using `kubeadm upgrade node experimental-control-plane` that's true. But starting from kubeadm 1.15 onward it will became stable `kubeadm upgrade node control-plane`. So `v1beta2` is forward compatible up to the kubeadm version that will drop v1beta2. That's why it's default.

When new versions will be released or at least close to it, we will insert them here also.

pkg/templates/kubeadm/provider.go
kron4eg: >but fail on install

Not necessary. As kubeadm interfaces are stabilizing they are becoming more and more forward compatible. In order to avoid failing with 1.15 I'll add v1beta2 kubeadm APIs to this PR.

pkg/templates/kubeadm/kubeadmv1beta1.go
kron4eg: changed to Command

pkg/templates/kubeadm/provider.go
kron4eg: changed to Kubeadm

xmudrii:
This might be required for the v0.9 release and for 1.15 to work properly with all features
/priority critical-urgent

Issue #372, opened 2019-04-25
Refactor tests


xmudrii:
## Refactor tests

We should refactor all tests suites so we improve stability of KubeOne and the CI pipeline. There are multiple problems that should be addressed by this issue:

* Unit tests coverage is very bad,
* E2E tests are flaking a lot,
* We don't run E2E tests for all providers,
* Our E2E tests are potentially too repetitive,
* It takes a lot of time to merge a PR,
* It's hard to test is everything working locally without spending money,
* It often happens that we leave plenty of hanging cloud resources on all providers.

### E2E provider-coverage

* [ ] [Add E2E tests for GCP (#388)](https://github.com/kubermatic/kubeone/issues/388)
* [ ] [Add E2E tests for OpenStack (#404)](https://github.com/kubermatic/kubeone/issues/404)
* [ ] [Add E2E tests for Packet (#403)](https://github.com/kubermatic/kubeone/issues/403)
* [ ] Consider testing are all CNI plugins working

### E2E general

* [ ] [Run full "conformance" test suite (#158)](https://github.com/kubermatic/kubeone/issues/158)
* [ ] [Use Ginkgo directly in E2E tests (#162)](https://github.com/kubermatic/kubeone/issues/162)
* [ ] Add `kubeone` prefix to E2E resources

PR #487, opened 2019-06-11
Make ams3 default DigitalOcean region

xmudrii:

After some discussion, we decided to try to switch to another DO region to see is that going to decrease the number of E2E failures and flakes. For now, we'll try Amsterdam (`ams3`) region.

xref #372

**Release note**:
```release-note
NONE
```

/assign @kron4eg


xmudrii:
It seems like this doesn't improve the number of flakes at all, so I'll close this PR.

/close

PR #489, opened 2019-06-12
Add E2E tests for Packet and GCE. Drop 1.13 tests and use 1.14.3 instead of 1.14.1

### LINKED ISSUE #388: Add E2E tests for GCP
In #317 we added support for GCP/GCE, but E2E tests were postponed for a follow-up. The following test suites should be implemented:

* [ ] Kubernetes conformance on 1.13
* [ ] Kubernetes conformance on 1.14
* [ ] Kubernetes conformance for 1.13 to 1.14 upgraded cluster

### LINKED ISSUE #403: Add E2E tests for Packet
In #384 we added support for Packet, but E2E tests were postponed for a follow-up. The following test suites should be implemented:

* [ ] Kubernetes conformance on 1.13
* [ ] Kubernetes conformance on 1.14
* [ ] Kubernetes conformance for 1.13 to 1.14 upgraded cluster

xmudrii:

This PR adds E2E tests for Packet and GCE.

xref #403
xref #388
xref #372

**Release note**:
```release-note
NONE
```

/assign @kron4eg


hack/run_ci_e2e_test.sh
```diff
@@ -89,6 +89,12 @@ if [ -n "${RUNNING_IN_CI}" ]; then
   "hetzner")
     export HCLOUD_TOKEN=${HZ_E2E_TOKEN}
     ;;
+  "packet")
+    export PACKET_AUTH_TOKEN=${PACKET_API_KEY}
```
kron4eg: what about project id?

test/e2e/conformance_test.go
```diff
@@ -166,6 +157,14 @@ func TestClusterConformance(t *testing.T) {
 				t.Fatalf("failed to download kubeconfig failed ('kubeone kubeconfig'): %v", err)
 			}
 
+			// Run Terraform again for GCE to add nodes to the load balancer
```
kron4eg: well, for E2E it will be perfectly find to run cluster only on 1 node behind the LB.

test/e2e/provisioner/gce.go
```diff
@@ -0,0 +1,80 @@
...
+
+		init: false,
+	}, nil
+}
+
+// Provision provisions a cluster using Terraform
+func (p *GCEProvisioner) Provision() (string, error) {
+	args := []string{}
+	if !p.init {
+		args = []string{"-var", "control_plane_target_pool_members_count=1"}
```
kron4eg: is new type GCEProvisioner needed only for this line?

hack/run_ci_e2e_test.sh
xmudrii: Project ID environment variable name exposed by CI already matches what we need, so I decided to skip it.

xmudrii: Should I add it?

test/e2e/conformance_test.go
xmudrii: I think we should test as we would do in the real environment. This adds some more lines, but it's a proper way.

test/e2e/provisioner/gce.go
xmudrii: /shrug

Basically. Another way was to extend DefaultProvisioner to take provider and add an init field, and then if provider GCE and if not init, add args and set init to true.

I don't have strong opinions, but provisioner is already small and mostly about that, so I went with a new one over extending default one.

hack/run_ci_e2e_test.sh
kron4eg: oh, well... probably not then.

test/e2e/provisioner/default.go
```diff
@@ -49,7 +49,7 @@ func NewDefaultProvisioner(creds func() error, testPath, identifier, provider st
 
 // Provision provisions a cluster using Terraform
 func (p *DefaultProvisioner) Provision() (string, error) {
```
kron4eg:
```suggestion
func (p *DefaultProvisioner) Provision(args ...string) (string, error) {
```

test/e2e/provisioner/default.go
```diff
@@ -49,7 +49,7 @@ func NewDefaultProvisioner(creds func() error, testPath, identifier, provider st
...
-	tf, err := p.terraform.initAndApply()
+	tf, err := p.terraform.initAndApply(nil)
```
kron4eg:
```suggestion
	tf, err := p.terraform.initAndApply(args)
```

test/e2e/provisioner/gce.go
xmudrii: The GCEProvisioner is removed in favor of the proposed approach.

/unshrug

test/e2e/provisioner/default.go
xmudrii: Fixed

test/e2e/provisioner/default.go
xmudrii: Fixed

xmudrii:
Required tests are passing. In order to re-enable the tests, I'm going to merge this and fix problems in a follow up.

/hold cancel

PR #494, opened 2019-06-13
Make gobetween.sh universal across different OS

kron4eg:
To handle all supported OS unconditionally (if ubuntu? if coreos?)

```release-note
NONE
```


kron4eg:
/override pull-kubeone-e2e-aws-conformance-1.14.1

As it has nothing to do with gobetwee

PR #495, opened 2019-06-14
Allow to automatically read terraform state from directory

### LINKED ISSUE #421: Read terraform outputs directly from terraform state file (if it's accessible)
State file is not always accessible (since state could be configured to remote), but when it does it already contains outputs and we can ready it directly. 

When state configured to some remote, old `terraform output -json` should be used.

eqrx:
Solves #421

Fixes #421

**Documentation**:
None

**Release note**:
```release-note
Terraform state can now be read from stdin, file or be collected from a directory
```

### DIFF
--- a/pkg/cmd/install.go
+++ b/pkg/cmd/install.go
@@ -76,7 +76,7 @@ It's possible to source information about hosts from Terraform output, using the
 
 // runInstall provisions Kubernetes on the provided machines
 func runInstall(logger *logrus.Logger, installOptions *installOptions) error {
-	cluster, err := loadClusterConfig(installOptions.Manifest, installOptions.TerraformState)
+	cluster, err := loadClusterConfig(installOptions.Manifest, installOptions.TerraformState, logger)
 	if err != nil {
 		return errors.Wrap(err, "failed to load cluster")
 	}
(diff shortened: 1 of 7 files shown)


kron4eg:
/assign @xmudrii 

PTAL

kron4eg:
@eqrx probably some log message is needed. Something like "exec terraform output".

kron4eg:
I've testes this with
```
$ kubeone install config.yaml -t .
$ kubeone reset config.yaml -t .
```

works quite good!

eqrx:
I could include the output, should I just write it to stdout or use some logger?

kron4eg:
oh, the output should be output only when `-v` flag is used (Verbose).

eqrx:
Should I pass the logger down to loadClusterConfig then?

kron4eg:
`Context.Logger` it is

xmudrii:
Looks great to me :100: As @kron4eg mentioned, it would be nice to add some output, but besides that lgtm.

eqrx:
util.Context is created after that function runs and is not available :(

kron4eg:
@eqrx then... call logger creation before running terraform output :)

eqrx:
@kron4eg alright :D on it

PR #496, opened 2019-06-14
Update machinecontroller to 1.2.1 and use proxy for kubelets

kron4eg:
```release-note
kubelets will have proxy support if it's specified in config
```

### DIFF
--- a/pkg/installer/installation/prerequisites.go
+++ b/pkg/installer/installation/prerequisites.go
@@ -26,7 +26,174 @@ import (
 	"github.com/kubermatic/kubeone/pkg/util"
 )
 
-const dockerVersion = "18.09.2"
+const (
+	dockerVersion = "18.09.6"
+)
+
+const (
+	daemonsProxyScript = `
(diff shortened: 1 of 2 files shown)


pkg/installer/installation/prerequisites.go
```diff
@@ -26,7 +26,174 @@ import (
 	"github.com/kubermatic/kubeone/pkg/util"
 )
 
-const dockerVersion = "18.09.2"
+const (
+	dockerVersion = "18.09.6"
+)
+
+const (
```
xmudrii: Why do you repeat `const` two times?

kron4eg: it's two blocks of consts, 1 for "constants" and 1 for "scripts"

xmudrii: Usually, that doesn't matter at all, but I'm not going to block on this.

kron4eg:

PR #506, opened 2019-06-20
Make E2E tests version-agnostic

xmudrii:

Considering steps to run E2E tests are the same for each Kubernetes version, I think we should not create a test case for each version, but instead, just provide the version to be used as a flag.

This PR updates the E2E tests and the script used to start the tests.

xref #372

**Release note**:
```release-note
NONE
```

/assign @kron4eg


xmudrii:
/hold
This requires an update to the presubmit jobs

PR #515, opened 2019-06-27
node local dns cache

### LINKED ISSUE #389: Add nodelocal dns cache

The [nodelocal dns cache](https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/0030-nodelocal-dns-cache.md) as it greatly improves resiliency of DNS by using a Daemonset to deploy a DNS cache on all nodes, which creates a local interface, adds an IP address and binds to it. After reconfiguring the Kubelet with the address of the interface, all DNS requests can be made without requiring NAT, avoiding [known races there](https://www.weave.works/blog/racy-conntrack-and-dns-lookup-timeouts). The connection from the dns cache to the upstream DNS is made via TCP where the race doesn't occur.

Sample manifest using `169.254.20.10` as address to bind to on nodes: https://github.com/kubermatic/infra/blob/master/prow/build-cluster-manifests/dns.yaml

kron4eg:
Fixes #389 

```release-note
new optional feature: node local DNS cache
```


pkg/cmd/config.go
```diff
@@ -447,6 +449,14 @@ features:
   # 'kube-system' namespace pods to 'use' it.
   podSecurityPolicy:
     enable: {{ .EnablePodSecurityPolicy }}
+  # Enables initialization of local DNS cache on every node in the cluster so
+  # known problematic race conditions could be avoided. This only needed on
```
xmudrii:
```suggestion
  # known problematic race conditions could be avoided. This is only needed on
```

pkg/features/activate.go
```diff
@@ -35,6 +35,10 @@ func Activate(ctx *util.Context) error {
 		return errors.Wrap(err, "failed to install metrics-server")
 	}
 
+	if err := installNodeLocalDNSCache(ctx.Cluster.Features.NodeLocalDNSCache.Enable, ctx); err != nil {
```
xmudrii: Is this NPE safe? What if `ctx.Cluster.Features.NodeLocalDNSCache` is nil? Do we check that somewhere before running this?

pkg/templates/dnscache/deployment.go
```diff
@@ -0,0 +1,272 @@
...
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+const (
+	imagetag         = "k8s.gcr.io/k8s-dns-node-cache:1.15.2"
+	dnscacheCorefile = `
+cluster.local:53 {
```
xmudrii: What if services DNS name is not `cluster.local`? Is it relevant and still going to work as expected?

pkg/templates/dnscache/deployment.go
```diff
@@ -0,0 +1,272 @@
...
+	errors
+	cache {
+		success 9984 30
+		denial 9984 5
+	}
+	reload
+	loop
+	bind 169.254.20.10
```
xmudrii: Are all those IP addresses same regardless of the setup?

xmudrii: To me, it seems like this is related to the Cluster DNS IP, so I believe it should work, but just checking.

pkg/templates/machinecontroller/deployment.go
```diff
@@ -715,16 +715,23 @@ func machineControllerMachineDeploymentCRD() *apiextensions.CustomResourceDefini
...
+		}
+		clusterDNSstring = clusterDNS.String()
 	}
 
 	args := []string{
 		"-logtostderr",
 		"-v", "4",
 		"-internal-listen-address", "0.0.0.0:8085",
-		"-cluster-dns", clusterDNS.String(),
+		"-cluster-dns", clusterDNSstring,
```
xmudrii: Do we need to set anything similar on the control plane nodes? As far as I know, this is a kubelet setting, so eventually, we maybe need to have the same value on all kubelets.

pkg/cmd/config.go
kron4eg: fixed

pkg/templates/dnscache/deployment.go
kron4eg: fixed

pkg/templates/dnscache/deployment.go
kron4eg: yep, it's always the same.

pkg/templates/machinecontroller/deployment.go
kron4eg: eventually yes

pkg/features/activate.go
kron4eg: yeah, not it wasn't. now fixed.

pkg/templates/machinecontroller/deployment.go
kron4eg: done

pkg/features/node_local_dns_cache.go
```diff
@@ -0,0 +1,49 @@
...
+func updateNodeLocalDNSCacheKubeadmConfig(feature *kubeoneapi.NodeLocalDNSCache, args *kubeadmargs.Args) {
+	if feature == nil {
+		return
+	}
+
+	if !feature.Enable {
+		return
+	}
+
+	args.Kubelet.ExtraArgs["cluster-dns"] = dnscache.VirtualIP
```
xmudrii: Do we initialize the ExtraArgs structure before using it here?

kron4eg: yes, `kubeadmargs.New()` doing this

PR #522, opened 2019-07-01
Enable NodeLocalDNSCache by default

xmudrii:

This PR enables NodeLocalDNSCache by default.

**Release note**:
```release-note
Enable NodeLocalDNSCache feature by default
```

/assign @kron4eg @alvaroaleman

### DIFF
--- a/pkg/apis/kubeone/v1alpha1/defaults.go
+++ b/pkg/apis/kubeone/v1alpha1/defaults.go
@@ -111,6 +111,11 @@ func SetDefaults_Features(obj *KubeOneCluster) {
 			Enable: true,
 		}
 	}
+	if obj.Features.NodeLocalDNSCache == nil {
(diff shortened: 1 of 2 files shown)


xmudrii:
/hold
until we don't test this and make the final decision

xmudrii:
@kron4eg Done, PTAL

xmudrii:
After a very long discussion on the KubeOne internal channel and on the planning meeting, we've decided to not continue with this. There are some problems involved with NodeLocalDNSCache, user experience, as well as that we had no chance to test this for some time before the release.

We may reconsider how we want to work with NodeLocalDNSCache in the future or for the v0.10 release cycle.

/close

kron4eg:
Currently it's not possible to automate disabling this at later stage of cluster life. As feature is regulated by the flag, which will be changed in the future and all kubelet specific configs will use https://godoc.org/k8s.io/kubelet/config/v1beta1 API instead of flags.

PR #534, opened 2019-07-02
Refactor the pkg/util package

### LINKED ISSUE #386: Refactor the pkg/util package
We have [the `pkg/util` package](https://github.com/kubermatic/kubeone/tree/8b35b17876dd2f547205a3ab8468cf3b5d37d95c/pkg/util) which contains various functions. This is an anti-pattern in Go and we should refactor the package.

xmudrii:

The `pkg/util` package we have right now is considered as an anti-pattern in Go and therefore it has been decided to refactor it. This PR moves all functions from the `pkg/util` package to new subpackages under `pkg/util/`.

Fixes #386 

```release-note
NONE
```

/assign @kron4eg


pkg/certificate/ca.go
```diff
@@ -21,12 +21,13 @@ import (
...
 	"github.com/kubermatic/kubeone/pkg/ssh"
-	"github.com/kubermatic/kubeone/pkg/util"
+	"github.com/kubermatic/kubeone/pkg/util/context"
+	"github.com/kubermatic/kubeone/pkg/util/runner"
 )
 
 // DownloadCA grabs CA certs/keys from leader host
-func DownloadCA(ctx *util.Context) error {
-	return ctx.RunTaskOnLeader(func(ctx *util.Context, _ *kubeoneapi.HostConfig, conn ssh.Connection) error {
+func DownloadCA(ctx *context.Context) error {
```
kron4eg: I really don't like the name `context.Context`, since it exact clash with https://golang.org/pkg/context/#Context

xmudrii: Do you have idea what name should we use instead? In some places I had the `kubeonecontext` import alias, but not sure I'd name the package like that

pkg/certificate/ca.go
```diff
@@ -21,12 +21,13 @@ import (
 
 	kubeoneapi "github.com/kubermatic/kubeone/pkg/apis/kubeone"
 	"github.com/kubermatic/kubeone/pkg/ssh"
-	"github.com/kubermatic/kubeone/pkg/util"
+	"github.com/kubermatic/kubeone/pkg/util/context"
+	"github.com/kubermatic/kubeone/pkg/util/runner"
```
kron4eg: I suggest to move `runner` out of `util`.

pkg/certificate/ca.go
xmudrii: Regarding your edit
> Ideally we should use it in code as kubeone.Context.

I'm against having a package called `kubeone` because I don't really like the idea of having `github.com/kubermatic/kubeone/pkg/util/kubeone` and we already use `kubeone` package for the API, so imports will clash again.

pkg/certificate/ca.go
xmudrii: I've thought about that as well, but I wasn't sure. `runner` is something you're not going to use standalone—you're going to use it in some package to run something, so it may make sense to be a utility (`util`) package.

xmudrii: Done.

pkg/cmd/kubeconfig.go
```diff
@@ -77,12 +77,12 @@ func runKubeconfig(logger *logrus.Logger, kubeconfigOptions *kubeconfigOptions)
 		return errors.Wrap(err, "failed to load cluster")
 	}
 
-	kubeconfig, err := util.DownloadKubeconfig(cluster)
+	kc, err := kubeconfig.DownloadKubeconfig(cluster)
```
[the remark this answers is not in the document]
xmudrii: Done :)

PR #537, opened 2019-07-03
Add Dockerfile and helper scripts for easier distribution

### LINKED ISSUE #473: Distribute a Docker image of KubeOne
Distribute a Docker image for users to retrieve to elimate the need to install packages.
This requires mounting of configuration files and integration into the SSH agent running in the user environment.

eqrx:
Adds build instructions for a docker image that can be used to execute kubeone and terraform independently of the underlying distribution.

Fixes #473 

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ PROVIDER=$(notdir $(wildcard ./terraform/*))
 CREATE_TARGETS=$(addsuffix -env,$(PROVIDER))
 DESTROY_TARGETS=$(addsuffix -env-cleanup,$(PROVIDER))
 
-.PHONY: all install-via-docker install build vendor download-dependencies
+.PHONY: all install-via-docker docker install build vendor download-dependencies
 .PHONY: generate-internal-groups verify-dependencies
 all: install
 
@@ -40,6 +40,10 @@ install-via-docker: docker-make-install
 		-w /go/src/github.com/kubermatic/kubeone \
 		$(BUILD_IMAGE) \
 		make install
+
+.PHONY: docker
+docker: dist/kubeone
+	docker build -t quay.io/kubermatic/kubeone:$(GITTAG) -f build/docker/Dockerfile .
 
 install:
 	go install $(GOBUILDFLAGS) -ldflags='$(GOLDFLAGS)' -v .
--- a/README.md
+++ b/README.md
@@ -106,6 +106,25 @@ Use your favorite method to build it on your system, for example by using
 `aurutils`:
 ```bash
 aur sync kubeone && pacman -S kubeone
+#### Docker
+
+We distribute a docker image that contains KubeOne and Terraform to use for your
+setup.
+
+##### Building
+
+```bash
+docker build -t kubeone:<Tag> .
+```
+
+##### Running
+
+To receive SSH credentials KubeOne needs to connect to the local SSH
+agent. This is done by mounting its UNIX socket. You can execute KubeOne and
+Terraform that way by running the following command line:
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -36,6 +36,10 @@ install:
 kubeone: build
 build: dist/kubeone
 
+.PHONY: docker
+docker: dist/kubeone
+	docker build -t quay.io/kubermatic/kubeone:$(GITTAG) -f build/docker/Dockerfile .
```
xmudrii: The `$(GITTAG)` is going to be something like `v0.9.0-1-<hash>` I believe. Or that's at least what I get when I run the command sourced by variable.

Do we want to use that or just version like `0.9.0`?

build/docker/kubeone.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
xmudrii: I'd prefer having this in the README instead of the script or maybe having both. The scripts are also not platform-agnostic, i.e. those are going to work on Linux and macOS but not on Windows.

Makefile
eqrx: I am unsure about that. Do we want some sort of "nightly" image? or do we want to build it on releases?

build/docker/kubeone.sh
eqrx: Right you are. On it

README.md
```diff
@@ -103,6 +103,25 @@ Use your favorite method to build it on your system, for example by using
 `aurutils`:
 ```bash
 aur sync kubeone && pacman -S kubeone
+#### Docker
```
xmudrii: You're missing ``` causing markdown to render incorrectly

README.md
```diff
@@ -103,6 +103,25 @@ Use your favorite method to build it on your system, for example by using
...
+
+We distribute a docker image that contains KubeOne and Terraform to use for your
```
xmudrii: Nit: for now, saying something along `We provide a Dockerfile that you can use to build an image containing KubeOne and Terraform...` could sound better, as we still don't push images.

README.md
```diff
@@ -103,6 +103,25 @@ Use your favorite method to build it on your system, for example by using
...
+
+```bash
+docker build -t kubeone:<Tag> .
+```
+
+##### Running
+
+To receive SSH credentials KubeOne needs to connect to the local SSH
+agent. This is done by mounting its UNIX socket. You can execute KubeOne and
+Terraform that way by running the following command line:
```
xmudrii: s/command line/command

README.md
```diff
@@ -106,6 +106,25 @@ Use your favorite method to build it on your system, for example by using
...
+
+```bash
+docker run --rm --user $(id -u) --name kubeone --mount type=bind,source="$SSH_AUTH_SOCK",target=/run/ssh.sock --mount type=bind,source="$(pwd)",target=/mnt -t kubeone:<Version> <Command>
```
kron4eg: oh well.. all the ENV with credentials is missing, and I suppose before we get "secrets in file" it will be very cumbersome to list every single ENV in CLI

eqrx: This is my problem as well .... kubeone in a container seems rather useless since you need to work around the sandboxing.

eqrx:
I do not see how to make the image having a good UX. To use kubeone in it you need to

- have the same SSH version inside and outside of the container
- use a really long command line for mounts, envs etc
- get all the tokens into the container, either by specifying all of them in the command line or put them in a file and source that

PR #550, opened 2019-07-04
Gomod

### LINKED ISSUE #524: Switch to Go modules

With Go modules being more and more stable and becoming default soon, we should consider switching to Go modules. Kubernetes has also switched to Go modules as of 1.15.

AC

* [x] Switch to Go modules
* [x] Make code generation works after switching to Go modules

Resources

https://github.com/kubernetes/kubernetes/pull/74877
[Introduction to Go modules by Jordan Liggitt](https://www.youtube.com/watch?v=OrMA16ASz2U)

eqrx:
Makes KubeOne a go module

Fixes #524 

None

```release-note
KubeOne is now a go module
```


Makefile
```diff
@@ -45,18 +45,10 @@ lint:
 test:
 	CGO_ENABLED=1 go test -race ./...
 
-.PHONY: dep
-dep:
-	dep ensure -v
```
kron4eg: I think we can replace this with `go mod download`

Makefile
```diff
@@ -45,18 +45,10 @@ lint:
...
-
 .PHONY: licence-check
 licence-check:
 	wwhrd check
 
-.PHONY: check-dependencies
-check-dependencies:
-	dep check
```
kron4eg: and this with `go mod verify`

Makefile
eqrx: Should we still call that target "dep"?

kron4eg: yeah, SGTM, since it's from "dependency" so whatever 🤷‍♂️

eqrx:
Currently not possible because of the kubernetes code-generator. See issue for more details.

PR #551, opened 2019-07-04
Update machine-controller to v1.4.1 and add credentials validation

xmudrii:

This PR updates `machine-controller` to v1.4.1. The default root disk size in the Terraform script is increased to 50 GB.

```release-note
Update machine-controller to v1.4.1
Add OS_TENANT_ID environment variable
```

/assign @kron4eg

### DIFF
--- a/docs/environment_variables.md
+++ b/docs/environment_variables.md
@@ -27,6 +27,7 @@ In the following table you can find all configuration variables with support for
 | `OS_USERNAME` | The username of the OpenStack user |
 | `OS_PASSWORD` | The password of the OpenStack user |
 | `OS_DOMAIN_NAME` | The name of the OpenStack domain |
+| `OS_TENANT_ID` | The ID of the OpenStack tenant |
 | `OS_TENANT_NAME` | The name of the OpenStack tenant |
 | | |
 | `PACKET_AUTH_TOKEN` | Packet auth token |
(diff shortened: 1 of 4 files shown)


kron4eg:
@xmudrii but about about TenantID?

xmudrii:
@kron4eg Fixed, PTAL and let me know if anything else is needed.

pkg/util/credentials/credentials.go
```diff
@@ -101,6 +102,7 @@ func ProviderCredentials(p kubeone.CloudProviderName) (map[string]string, error)
 			{Name: "OS_PASSWORD"},
 			{Name: "OS_DOMAIN_NAME"},
 			{Name: "OS_REGION_NAME"},
+			{Name: "OS_TENANT_ID"},
```
kron4eg: The problem here is that only ether of two (`OS_TENANT_ID` or `OS_TENANT_NAME`) is required, but not both simultaneously. But this line makes both of them always required, which is not what MC expect.

xmudrii: Both are not required to be present, but what happens in such case? Should we throw an error or have higher priority for one of them?

kron4eg: That's the trick, we need to make sure that one of ether is present, we don't need both of them, [here's MC error check](https://github.com/kubermatic/machine-controller/pull/580/files#diff-e07452860a04f7019cbb8dfb86d357c1R353)

PR #555, opened 2019-07-04
[WIP] Support sourcing AWS credentials from the AWS profiles file

### LINKED ISSUE #501: Ensure AWS credentials are handled properly
In #623, we have added an option to source credentials from a credentials file. This works well for other providers, but for AWS can be further improved.

For AWS, there is a default file called `~/.aws/credentials`, where users can store profiles and credentials for each profile. Eventually, we should support sourcing from that file, instead of requiring users to put credentials in another file or the environment.

Besides that, AWS instances can communicate with the API without the token. If `machine-controller` is running on AWS instance, `machine-controller` should be able to obtain the permission using IAM. Currently, this is not possible because KubeOne requires credentials to be present in the form of the environment variables or credentials file.

xmudrii:

Earlier we have added a function that sources AWS credentials from the profiles file, but actually, that function never worked. This PR fixes the behavior and ensures credentials can be sourced from the AWS profiles file.

Two new flags are added:

* `--aws-profile-path` - a path to the file where AWS profiles are located (default `~/.aws/credentials`)
* `--aws-profile-name` - a name of the profile that will be deployed on the cluster to be used by `machine-controller` (default `default`)

Fixes #501

```release-note
Fix sourcing AWS credentials from the AWS profiles file
Add --aws-profile-path and --aws-profile-name flags
```

/assign @kron4eg

### DIFF
--- a/pkg/cmd/install.go
+++ b/pkg/cmd/install.go
@@ -34,6 +34,9 @@ type installOptions struct {
 	globalOptions
 	Manifest   string
 	BackupFile string
+
+	AWSProfilePath string
+	AWSProfileName string
 }
 
 // installCmd setups install command
@@ -70,6 +73,9 @@ It's possible to source information about hosts from Terraform output, using the
 	}
 
 	cmd.Flags().StringVarP(&iopts.BackupFile, "backup", "b", "", "path to where the PKI backup .tar.gz file should be placed (default: location of cluster config file)")
+
+	cmd.Flags().StringVarP(&iopts.AWSProfilePath, "aws-profile-path", "", "~/.aws/credentials", "path to the file where aws credentials are located")
+	cmd.Flags().StringVarP(&iopts.AWSProfileName, "aws-profile-name", "", "default", "name of the aws profile to be deployed on the cluster for machine-controller")
 
 	return cmd
 }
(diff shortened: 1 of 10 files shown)


xmudrii:
`yamled` flake
/retest

pkg/cmd/install.go
```diff
@@ -71,12 +74,15 @@ It's possible to source information about hosts from Terraform output, using the
 
 	cmd.Flags().StringVarP(&iopts.BackupFile, "backup", "b", "", "path to where the PKI backup .tar.gz file should be placed (default: location of cluster config file)")
 
+	cmd.Flags().StringVarP(&iopts.AWSProfilePath, "aws-profile-path", "", "~/.aws/credentials", "path to the file where aws credentials are located")
+	cmd.Flags().StringVarP(&iopts.AWSProfileName, "aws-profile-name", "", "default", "name of the aws profile to be deployed on the cluster for machine-controller")
```
kron4eg: I suppose this flag should be empty by default, so people would have to explicitly opt-in and provide `--aws-profile-name=whatever` in order to give-up credentials from `~/.aws/credentials`.

xmudrii:
/close
Starting over in a new PR

PR #556, opened 2019-07-04
Fix kubelet upgrade on Container Linux

ainmosni:

On Container Linux, kubelet locks itself, this patch stops kubelet
before downloading the new binaries and then starts it back up once
everything is ready.

```release-note
* Fix issue where kubelet wouldn't properly upgrade on Container Linux
```

### DIFF
--- a/pkg/upgrader/upgrade/kubernetes_binaries.go
+++ b/pkg/upgrader/upgrade/kubernetes_binaries.go
@@ -59,14 +59,18 @@ curl -L "https://github.com/containernetworking/plugins/releases/download/{{ .CN
 
 RELEASE="v{{ .KUBERNETES_VERSION }}"
 
-sudo systemctl stop kubelet
+
+
+sudo mkdir -p /var/tmp/kube-binaries
+cd /var/tmp/kube-binaries
+sudo curl -L --remote-name-all \
+     https://storage.googleapis.com/kubernetes-release/release/${RELEASE}/bin/linux/amd64/{kubeadm,kubelet,kubectl}
+
 
 sudo mkdir -p /opt/bin
 cd /opt/bin
-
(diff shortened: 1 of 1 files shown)


pkg/upgrader/upgrade/kubernetes_binaries.go
```diff
@@ -59,8 +59,12 @@ curl -L "https://github.com/containernetworking/plugins/releases/download/{{ .CN
 
 RELEASE="v{{ .KUBERNETES_VERSION }}"
 
+sudo systemctl stop kubelet
```
kron4eg: I'd prefer stopping kubelet only after download already finished and we know for real it's OK disk

ainmosni: Done :)

pkg/upgrader/upgrade/kubernetes_binaries.go
```diff
@@ -59,10 +59,18 @@ curl -L "https://github.com/containernetworking/plugins/releases/download/{{ .CN
 
 RELEASE="v{{ .KUBERNETES_VERSION }}"
 
-sudo mkdir -p /opt/bin
-cd /opt/bin
+
+
+sudo mkdir -p /var/tmp/kube-binaries
```
xmudrii: Can we maybe use `/tmp`? Does it exist on CoreOS out of box?

kron4eg:
/hold

@xmudrii please unhold once done

pkg/upgrader/upgrade/kubernetes_binaries.go
ainmosni: I tested it on our machines, and yes it does.

PR #561, opened 2019-07-04
WIP Node local DNS cache

### LINKED ISSUE #389: Add nodelocal dns cache

kron4eg:

Fixes #389 

```release-note
Node local DNS cache
```


kron4eg:
this one is replaced by #704

PR #575, opened 2019-07-12
Fix that install subcommand requires credentials for none provider

eqrx:
Fix that install subcommand requires credentials for none provider.

None

None

```release-note
NONE
```

### DIFF
--- a/pkg/config/cluster.go
+++ b/pkg/config/cluster.go
@@ -542,6 +542,8 @@ func (p ProviderName) ProviderCredentials() (map[string]string, error) {
 			{Name: VSphereUsername},
 			{Name: VSpherePasswords},
 		})
+	case ProviderNameNone:
+		return parseCredentialVariables([]ProviderEnvironmentVariable{})
 	}
 
 	return nil, errors.New("no provider matched")
--- a/pkg/credentials/credentials.go
+++ b/pkg/credentials/credentials.go
@@ -142,6 +142,8 @@ func ProviderCredentials(p kubeone.CloudProviderName) (map[string]string, error)
 		// force scheme, as machine-controller requires it while terraform does not
 		vscreds[VSphereAddress] = "https://" + vscreds[VSphereAddress]
 		return vscreds, nil
+	case kubeone.CloudProviderNameNone:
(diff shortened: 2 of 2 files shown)


xmudrii:
/hold

I want to check something

xmudrii:
We validate and deploy credentials in two cases:
* If machine-controller is deployed (`.machineController.deploy` is set to `true`)
* If external CCM is deployed (`.provider.external` is set to `true`)

When deploying the external CCM, there are two bugs:
1) Validation is going to check for credentials, but as the provider name is `none`, the validation is going to fail
2) In a case when `.machineController.Deploy` is `false`, but `.provider.Name` is a provider that we don't deploy CCM for, the validation is going to fail as well. This is not expected as we don't deploy credentials, so we don't need them. This case is currently very rare and might only happen if someone tries to deploy external CCM for providers like AWS.

The machine-controller case is tricky. If we don't deploy machine-controller (and we don't deploy external CCM), we should just not check for credentials, as we don't need the credentials. If this is the case you're encountering, I think we should just change the validation rule instead. However, I was not able to reproduce this case locally, so if you share the config you used, maybe I can check is it really related to this. :slightly_smiling_face: 

Another case is that if you want to deploy machine-controller but with `.provider.Name` and `.machineController.Provider` set to `none`. My opinion is that we should still reject that, but if there are requests and good use cases, I'm okay with changing this. The problem with this case is that if `.machineController.Provider` is `none`, it's not going to deploy credentials, causing:

* Machines not to be created, although machines are going to be created as soon as the credentials are deployed,
* The user has to manually deploy the secret with credentials, but also change the machine-controller deployment to create a binding to the secret

This PR is going to allow this and the question is do we want to allow this case to happen? If we do, I would like to see a log message or a document explaining what the user should do and what is going to happen.

eqrx:
I used the ansible example for that. Check out the branch [ansible](https://github.com/kubermatic/kubeone/tree/ansible) and just run the install playbook. kubeone will explode before anything real happens :)

If we a stuck a little bit longer with that.... is there are workaround for this since I need it to work on wednesday

eqrx:
Error was on my side

PR #595, opened 2019-07-19
Add reference to private key file, as it's not configured by default

fletchgqc:
Signed-off-by: John Fletcher <john.fletcher@codecentric.de>

The getting started with AWS guide doesn't work as it currently stands.

```release-note
```NONE

### DIFF
--- a/docs/quickstart-aws.md
+++ b/docs/quickstart-aws.md
@@ -59,6 +59,7 @@ For the list of available settings along with their names please see the [`varia
 * `cluster_name` (required) - prefix for cloud resources
 * `aws_region` (default: eu-west-3)
 * `ssh_public_key_file` (default: `~/.ssh/id_rsa.pub`) - path to your SSH public key that's deployed on instances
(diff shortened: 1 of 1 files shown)


kron4eg:
@fletchgqc hey and thanks for PR.

We deliberately left out `ssh_private_key_file` from examples, as it means that SSH private key should be unencrypted. Kubeone can only work with SSH unencrypted private key files. Which is actually sounds quite bad :). Kubeone delegates handling of encrypted SSH key files to `ssh-agent`. You can find more information here: https://github.com/kubermatic/kubeone/blob/master/docs/ssh.md

I'll close this PR as it's against our original intention.

P.S.
Please don't use `ssh_private_key_file` — in order to use it you need to store it unencrypted, which is bad for security.

fletchgqc:
OK, I understand, I would recommend that you nevertheless change your docs to explain this. In the end you want to encourage security, but not scare away beginners by forcing them to change their key management practices when they are just trying to test out your system.

PR #596, opened 2019-07-19
Add support for Hetzner Private Networking

### LINKED ISSUE #661: Support private hetzner networking

kron4eg:
Fixes #661

```release-note
Add support for Hetzner Private Networking
```

### DIFF
--- a/examples/terraform/hetzner/main.tf
+++ b/examples/terraform/hetzner/main.tf
@@ -21,14 +21,14 @@ resource "hcloud_ssh_key" "kubeone" {
 
 resource "hcloud_network" "net" {
   name     = var.cluster_name
-  ip_range = "192.168.0.0/16"
+  ip_range = var.ip_range
 }
 
 resource "hcloud_network_subnet" "kubeone" {
   network_id   = hcloud_network.net.id
   type         = "server"
-  network_zone = "eu-central"
-  ip_range     = "192.168.0.0/24"
+  network_zone = var.network_zone
+  ip_range     = var.ip_range
 }
 
 resource "hcloud_server_network" "control_plane" {
--- a/examples/terraform/hetzner/main.tf
+++ b/examples/terraform/hetzner/main.tf
@@ -19,6 +19,24 @@ resource "hcloud_ssh_key" "kubeone" {
   public_key = file(var.ssh_public_key_file)
 }
 
+resource "hcloud_network" "net" {
+  name     = var.cluster_name
+  ip_range = "192.168.0.0/16"
+}
+
+resource "hcloud_network_subnet" "kubeone" {
+  network_id   = hcloud_network.net.id
+  type         = "server"
+  network_zone = "eu-central"
+  ip_range     = "192.168.0.0/24"
+}
+
+resource "hcloud_server_network" "control_plane" {
+  count       = 3
+  server_id  = element(hcloud_server.control_plane.*.id, count.index)
+  network_id = hcloud_network.net.id
+}
+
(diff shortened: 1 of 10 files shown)


kron4eg:
WIP because we also need to add hetzner private networking support to machine-controller

xmudrii:
This PR is currently blocked as Hetzner CCM doesn't support private networks

https://github.com/hetznercloud/hcloud-cloud-controller-manager/issues/19
https://github.com/hetznercloud/hcloud-cloud-controller-manager/pull/20

There is a WIP PR to implement Hetzner Networks for machine-controller

https://github.com/kubermatic/machine-controller/pull/627

xmudrii:
This issue is not anymore blocked as we have a new release of both machine-controller and Hetzner CCM.

/remove-lifecycle blocked

pkg/apis/kubeone/types.go
```diff
@@ -119,6 +119,10 @@ type ClusterNetworkConfig struct {
 	ServiceDomainName string `json:"serviceDomainName"`
 	NodePortRange     string `json:"nodePortRange"`
 	CNI               *CNI   `json:"cni,omitempty"`
+
+	// TODO(xmudrii): Revisit do we want to put this here.
+	// +optional
+	NetworkID string `json:"networkID,omitempty"`
```
xmudrii: Do we want this field to be here? If not, where should I put it?

pkg/apis/kubeone/types.go
```diff
@@ -119,6 +119,10 @@ type ClusterNetworkConfig struct {
 	ServiceDomainName string `json:"serviceDomainName"`
 	NodePortRange     string `json:"nodePortRange"`
 	CNI               *CNI   `json:"cni,omitempty"`
+
+	// TODO(xmudrii): Revisit do we want to put this here.
+	// +optional
```
xmudrii: Should I keep the `+optional` tag? If yes, should I make this a pointer to string?

pkg/templates/externalccm/hetzner.go
```diff
@@ -183,6 +193,11 @@ func hetznerDeployment() *appsv1.Deployment {
 										},
 									},
 								},
+								{
+									Name: "HCLOUD_NETWORK",
+									// TODO(xmudrii): Should we keep the network ID/name in the secret?
+									Value: networkID,
```
xmudrii: Should this be in a secret or this approach is okay?

xmudrii:
/hold
I haven't pushed and tested a fix for CCM flags

examples/terraform/hetzner/main.tf
```diff
@@ -19,6 +19,24 @@ resource "hcloud_ssh_key" "kubeone" {
...
+resource "hcloud_network" "net" {
+  name     = var.cluster_name
+  ip_range = "192.168.0.0/16"
+}
+
+resource "hcloud_network_subnet" "kubeone" {
+  network_id   = hcloud_network.net.id
+  type         = "server"
+  network_zone = "eu-central"
+  ip_range     = "192.168.0.0/24"
```
kron4eg: and this

pkg/templates/externalccm/hetzner.go
kron4eg: I don't know why networkID should be treated as a secret. So it's OK like this.

pkg/apis/kubeone/types.go
kron4eg: We should follow here what machine-controller uses (which is []string)

PR #597, opened 2019-07-19
Cleanup Makefile

kron4eg:
```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,86 +14,69 @@
 
 export GOPATH?=$(shell go env GOPATH)
 export CGO_ENABLED=0
-export KUBERNETES_VERSION=1.14.1
 export GOPROXY=https://proxy.golang.org
 export GO111MODULE=on
 export GOFLAGS?=-mod=readonly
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -14,86 +14,69 @@
...
 
-e2e-test: dist/kubeone
-	CGO_ENABLED=1 ./hack/run-ci-e2e-test.sh
+.PHONY: e2e-test
+e2e-test: install
+	./hack/run-ci-e2e-test.sh
 
+.PHONY: buildenv
+buildenv:
+	@go version
```
kron4eg: This line reports: `go version go1.11.4 linux/amd64`

Makefile
```diff
@@ -14,86 +14,69 @@
...
-verify-dependencies:
+.PHONY: verify-dependencies
+verify-dependencies: buildenv
 	go mod verify
 
-.PHONY: test e2e-test lint verify-licence verify-codegen verify-boilerplate
+.PHONY: test
 test:
-	CGO_ENABLED=1 go test $(GOBUILDFLAGS) -race ./pkg/... ./test/...
+	CGO_ENABLED=1 go test -race ./pkg/... ./test/...
```
xmudrii: Why do we enable cgo here?

kron4eg: Because of -race requesting it

Issue #533, opened 2019-07-01
Drop permissions in AWS policies 


kron4eg:
AWS terraform examples are using too liberal policies.

* [ ] lock down worker instance profiles to the very minimal permissions 
* [ ] lock down control plane to possible minimal 
* [ ] lock both on cluster tag

PR #602, opened 2019-07-25
Upgrade machinecontroller to v1.5.1

kron4eg:
* Upgrade machinecontroller to v1.5.1
* Use new assignPublicIP field for completely private clusters

```release-note
* Upgrade machinecontroller to v1.5.1
* Use new assignPublicIP field for completely private clusters
```


kron4eg:
/hold

there is an error from machine-controller

```
E0725 14:22:17.918795       1 reflector.go:134] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:126: Failed to list *v1beta1.CustomResourceDefinition: customresourcedefinitions.apiextensions.k8s.io is forbidden: User "system:serviceaccount:kube-system:machine-controller" cannot list resource "customresourcedefinitions" in API group "apiextensions.k8s.io" at the cluster scope
```

examples/terraform/aws-private/output.tf
```diff
@@ -47,32 +52,95 @@ output "kubeone_workers" {
   value = {
     # following outputs will be parsed by kubeone and automatically merged into
     # corresponding (by name) worker definition
-    "${var.cluster_name}-pool1" = {
+    "${var.cluster_name}-${local.zoneA}" = {
```
xmudrii: Should we create a node in each AZ for non-private AWS setup too? There we create only one MachineDeployment, with one replica, in the first AZ.

kron4eg:
> If this setup is now working as intended, can you please remove the experimental warning from the README

@xmudrii it's still experimental. As it still lacks the proper policies definitions (see #533).

examples/terraform/aws-private/output.tf
kron4eg: yeah, spreading control plane nodes across AZs but having workers in only 1 AZ is kinda strange. But this will require changes in e2e tests (the place where we count ready nodes).

xmudrii: > But this will require changes in e2e tests (the place where we count ready nodes).

That should be easy, you just have to increase numbers here:
* https://github.com/kubermatic/kubeone/blob/master/test/e2e/upgrade_test.go#L62
* https://github.com/kubermatic/kubeone/blob/master/test/e2e/conformance_test.go#L50

Do you want to do that now or in a follow-up?

kron4eg: someday ©

PR #606, opened 2019-07-29
Take static networking into account when contructing machineDeployment

kron4eg:
We had Network config structures already, but never used them, so this PR fixes this unfortunate incident.

```release-note
Configure static network
```

### DIFF
--- a/pkg/templates/machinecontroller/machines.go
+++ b/pkg/templates/machinecontroller/machines.go
@@ -35,14 +35,6 @@ import (
 	clusterv1alpha1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
 )
 
-type providerSpec struct {
-	SSHPublicKeys       []string                     `json:"sshPublicKeys"`
-	CloudProvider       kubeoneapi.CloudProviderName `json:"cloudProvider"`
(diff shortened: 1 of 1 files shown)


pkg/templates/machinecontroller/machines.go
```diff
@@ -75,15 +67,14 @@ func createMachineDeployment(cluster *kubeoneapi.KubeOneCluster, workerset kubeo
 		return nil, errors.Wrap(err, "failed to generate machineSpec")
 	}
 
-	config := providerSpec{
```
kron4eg: we don't need special `providerSpec` struct, we already have everything in put API, we this change introduce using existing structs.

pkg/templates/machinecontroller/machines.go
```diff
@@ -96,6 +87,12 @@ func createMachineDeployment(cluster *kubeoneapi.KubeOneCluster, workerset kubeo
 		"workerset": workerset.Name,
 	}
 
+	if workerset.Config.Network != nil {
+		// we have static network config
```
kron4eg: static network config is kinda special care and we need to flip those `maxSurge` and `maxUnavailable`

PR #607, opened 2019-07-30
Set hostname for GCE instances in Terraform scripts

xmudrii:
GCE now uses the instance name as a hostname, like AWS.

```release-note
NONE
```

/hold


xmudrii:
/hold cancel
The E2E tests are going to fail but this is as expected because we need to make prefixing E2E resources work in a follow-up.

PR #611, opened 2019-08-01
[WIP] Add ability to source secrets from a secrets manifest

### LINKED ISSUE #492: Allow providing secrets and cloudConfig in a form of a file

We should consider adding a way to provide secrets (like API tokens and credentials) and CloudConfig in a form of a file that would be provided on the install time.

Example:
```
kubeone install config.yaml -t tf.json --secrets secrets.yaml
```

* Users can choose to deploy other credentials to the cluster than those that are already in the environment
* Users can put cloudConfig which contains many secrets and credentials in another file and source it on the install time

xmudrii:
**Status**: Proposal/WIP

This PR proposes an additional way for sourcing secrets and credentials. Before, it was only possible to source secrets from environment variables, but with this PR, users can provide a "secrets manifest".

When a secrets manifest is provided, it has the highest priority, i.e. KubeOne will take values from the manifest instead from the environment. If a required secret is missing from the manifest, KubeOne will resort to the environment, giving you more flexibility to securely store and source secrets.

To make this possible, this implements a type called `KubeOneSecrets`. A secrets manifest would look like this:
```
kind: KubeOneSecrets
secrets:
  DIGITALOCEAN_TOKEN: "..."
```

With such manifest, KubeOne is invoked as:
```
kubeone install config.yaml -t . -s secrets.yaml
```

Fixes #492

This PR is only a proposal and we should eventually discuss this more. Under the hood, `secrets` is a `map[string]string`. This has some drawbacks that I'll mention in the review comments.

This PR is not complete. I'd like to ensure that this is a good way of doing this before continuing.

Missing features include:

* [ ] Ability to provide cloudConfig
* [ ] Proper credentials validation

```release-note
Add KubeOneSecrets API type
Add ability to source secrets from a secrets manifest
```


pkg/apis/kubeone/config/config.go
```diff
@@ -28,36 +28,11 @@ import (
...
 	kubeonev1alpha1 "github.com/kubermatic/kubeone/pkg/apis/kubeone/v1alpha1"
 	"github.com/kubermatic/kubeone/pkg/apis/kubeone/validation"
-	"github.com/kubermatic/kubeone/pkg/credentials"
 	"github.com/kubermatic/kubeone/pkg/terraform"
 
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
-// SetKubeOneClusterDynamicDefaults sets the dynamic defaults for a given KubeOneCluster object
-func SetKubeOneClusterDynamicDefaults(cfg *kubeoneapi.KubeOneCluster) error {
```
xmudrii: * We will not store credentials in the API anymore to increase security and make it easier to work with credentials and the API
* There will be RuntimeValidation that checks are credentials present
* This might require promoting the API to v1alpha2

pkg/apis/kubeone/types.go
```diff
@@ -52,6 +52,23 @@ type KubeOneCluster struct {
...
 
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// KubeOneSecrets hold secrets needed to provision the cluster
+type KubeOneSecrets struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// Secrets contains list of secrets to be used by KubeOne instead of
+	// environment variables
+	Secrets map[string]string `json:"secrets"`
```
xmudrii: This has both positive and negative sides.

From the positive side, we don't have a very complex type and users can just provide secrets similar to environment variables.

From the negative side, we don't have strong type validation, i.e. everything is a string and we can't validate names out-of-box without some more complex function (e.g. `credentials.ProviderCredentials`).

pkg/apis/kubeone/types.go
```diff
@@ -52,6 +52,23 @@ type KubeOneCluster struct {
...
+}
+
+// Secret contains a single secret
```
xmudrii: This is a leftover. To be removed.

pkg/credentials/credentials.go
```diff
@@ -147,11 +147,18 @@ func ProviderCredentials(p kubeone.CloudProviderName) (map[string]string, error)
 	return nil, errors.New("no provider matched")
 }
 
-func parseCredentialVariables(envVars []ProviderEnvironmentVariable, validationFunc func(map[string]string) error) (map[string]string, error) {
+func parseCredentialVariables(envVars []ProviderEnvironmentVariable, secrets *kubeone.KubeOneSecrets, validationFunc func(map[string]string) error) (map[string]string, error) {
 	// Validate credentials using given validation function
 	creds := make(map[string]string)
 	for _, env := range envVars {
-		creds[env.Name] = strings.TrimSpace(os.Getenv(env.Name))
+		if secrets != nil {
```
xmudrii: As mentioned in the PR description, this allows both secrets and environment variables.

pkg/installer/installer.go
```diff
@@ -38,20 +39,27 @@ type Options struct {
...
 	return &Installer{
 		cluster: cluster,
+		secrets: secrets,
 		logger:  logger,
 	}
 }
 
 // Install run the installation process
 func (i *Installer) Install(options *Options) error {
+	_, err := credentials.ProviderCredentials(i.cluster.CloudProvider.Name, i.secrets)
```
xmudrii: This is a hack to ensure validation would work. To be replaced with a proper runtime validation.

xmudrii:
/hold
Proposal/WIP

/assign @kron4eg 
For an initial review

pkg/apis/kubeone/types.go
```diff
@@ -52,6 +52,23 @@ type KubeOneCluster struct {
 	Credentials map[string]string `json:"credentials,omitempty"`
 }
 
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// KubeOneSecrets hold secrets needed to provision the cluster
+type KubeOneSecrets struct {
```
xmudrii: This adds a new API type. Eventually, we don't have to use API type like this, but have some package like the old API (`pkg/config`), although I prefer this way.

toschneck:
@xmudrii one question - why we introduce a separate `KubeOneSecret` object and not using the generic Kubernetes `Secret`? 
I think to be compatible with some KMS provider plugins, this would be needed :man_shrugging:

xmudrii:
/close
Starting over in a new PR

PR #615, opened 2019-08-04
Initial OpenStack external CCM setup.

ainmosni:
Adds support for the external OpenStack cloud controll manager.

As the internal OpenStack code in the controller manager
will be gone soon, we'll need to switch to the external cloud controller
manager, this PR adds the configuration for it.

```release-note
* Adds support for the external OpenStack Cloud Control Manager.
```

### DIFF
--- a/pkg/templates/externalccm/openstack.go
+++ b/pkg/templates/externalccm/openstack.go
@@ -17,19 +17,22 @@ limitations under the License.
 package externalccm
 
 import (
-	"github.com/Masterminds/semver"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
-	"strings"
 	"context"
+	"strings"
+
+	"github.com/Masterminds/semver"
 	"github.com/pkg/errors"
+
 	"github.com/kubermatic/kubeone/pkg/clientutil"
-	"k8s.io/apimachinery/pkg/runtime"
 	"github.com/kubermatic/kubeone/pkg/state"
-	"k8s.io/apimachinery/pkg/api/resource"
+
 	appsv1 "k8s.io/api/apps/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


xmudrii:
Hello @ainmosni,

thank you for the PR! However, before we can merge this, there are several things we need to take care of and decide about. The switch to the external CCMs is a little bit tough topic because there are some points that are still unclear.

You can find some of the discussion in #616

ainmosni:
@xmudrii Yeah the migration will be painful but as there's no real development on the in-tree one and the in-tree code will be removed sooner rather later, it might be worth to pull the bandaid off sooner rather than later. I can definitely take a look at the OpenStack side to help with the migration options.

kron4eg:
> @xmudrii Yeah the migration will be painful but as there's no real development on the in-tree one and the in-tree code will be removed sooner rather later

AFAIK, timeline is set to 1.17.x

>, it might be worth to pull the bandaid off sooner rather than later.

Indeed, that's why we create #616.

>I can definitely take a look at the OpenStack side to help with the migration options.

There is a problem with openstack CCM. Its releases are bound to specific minor kubernetes versions. So it's not that easy to jump on it and we need to take into account older version (since we support k8s starting from 1.13).

ainmosni:
Yeah I kept that in mind in this patch, AFAICS they only started to tag their images since 1.15, they used `latest` before that.

One thing that I didn't put in this patch but we should probably add is support for the cinder CSI plugin, so that PVCs keep working.

kron4eg:
OK, I've got a confirmation that it's safe to use OS CCM 1.15 in 1.13, 1.14 and 1.15 clusters (but no 1.12, but since we don't support 1.12 anyway, we are safe).

pkg/templates/externalccm/openstack.go
```diff
@@ -0,0 +1,342 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package externalccm
+
+import (
```
kron4eg: please take a look at https://github.com/kubermatic/kubeone/blob/master/CONTRIBUTING.md#import-order

ainmosni:
@kron4eg Yeah that's what I gathered from their examples as well. It's a bit more complicated with cinder-csi (which isn't in this PR) as the CSI sidecar containers are k8s version specific. I was thinking about adding those, but changed my mind to keep the PR smaller.

pkg/templates/externalccm/openstack.go
ainmosni: Sorted. :)

ivomarino:
Hi all, is this PR going to happen "soon" or what's the best actual solution on this topic? Thanks

kron4eg:
@ivomarino yes, in the nearest future, I'd say. I'm working on integrating changes from this PR.

kron4eg:
This PR is replaced by https://github.com/kubermatic/kubeone/pull/820

@ainmosni thanks for initiating this!

PR #619, opened 2019-08-14
Add documentation on how to configure proxies

eqrx:
Improve documentation

None

None

```release-note
NONE
```

### DIFF
--- a/docs/proxy.md
+++ b/docs/proxy.md
@@ -1,0 +1,16 @@
+# How to use KubeOne in a proxy environment
+
+Create a KubeOne configuration as usual and add the following section:
+
+```yaml
+proxy:
+  http_proxy: 'http://proxy.example.com'
+  https_proxy: 'https://proxy.example.com'
+  no_proxy: '127.0.0.1/8,localhost,*.local,10.10.10.0/24,10.20.0.0/16,10.254.0.0/16,172.25.0.0/16'
+```
+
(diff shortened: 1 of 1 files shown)


docs/proxy.md
```diff
@@ -0,0 +1,16 @@
+# How to use KubeOne in a proxy environment
+
+Create a KubeOne configuration as usual and add the following section:
+
+```yaml
+proxy:
+  http_proxy: 'http://proxy.example.com'
```
chrkl: This doesn't match the `kubeone config print` output:
proxy:
 http: ''
 https: ''
 noProxy: ''

kron4eg: indeed, this is pre-0.6 config format. it's time to forget about it once and for all.

toschneck:
@eqrx can you follow up on this?

xmudrii:
Agreed with @eqrx to close this PR since it's old and out-of-sync.

toschneck:
@xmudrii did we finally documented it somewhere. This could cost a lot of time at customer installation, so we really really should have a documented working example.
The best would also to have an e2e test for the proxy support.

xmudrii:
@toschneck No, we don't have any proxy-related document currently, but we would like to. I think it would be best that someone who has tested and used the proxy feature create a PR to document it.

xmudrii:
@toschneck I've created a follow-up issue #739.

PR #625, opened 2019-08-20
Run gofmt for hack/tools.go

xmudrii:

This PR runs gofmt for `hack/tools.go`.

```release-note
NONE
```

/assign @kron4eg

### DIFF
--- a/hack/tools.go
+++ b/hack/tools.go
@@ -14,14 +14,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-
 // +build tools
 
 package tools
 
 import (
-	_ "k8s.io/code-generator/cmd/deepcopy-gen"
-	_ "k8s.io/code-generator/cmd/conversion-gen"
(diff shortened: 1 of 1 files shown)


kron4eg:
heh, I see why it wasn't caught previously by linter

` golangci-lint run ./pkg/... ./test/... ` 🔥

xmudrii:
@kron4eg What would happen if we used `golangci-lint run ./...` (or `golangci-lint run` which is equivalent to the former command) instead? Would it include vendor too maybe?

kron4eg:
I suppose this is leftover since dark times when we had vendor :D

xmudrii:
Oh, right, we don't have `vendor` anymore. :man_shrugging: I'll update the PR to include all packages for lint.

xmudrii:
@kron4eg Updated the PR.

xmudrii:
@kron4eg Reverted the change of linter settings.

PR #631, opened 2019-08-21
Implement StaticAuditLog feature

### LINKED ISSUE #629: Support "static" Audit logging
KubeOne currently only support dynamic logging. "Static" logging is also useful since it is:
- stable
- does not require an insecure logstash sink

Please support this with a feature flag in the config. Thank you!

xmudrii:

This PR implements a feature called `StaticAuditLog`. This feature configures the audit backend to use the log backend, as described [here](https://kubernetes.io/docs/tasks/debug-application-cluster/audit/#log-backend).

Fixes #629

```release-note
Implement StaticAuditLog feature
```


xmudrii:
/assign @kron4eg @eqrx 
Ready for review

kron4eg:
terraform complaining ` Error: Error creating network Load Balancer: TooManyLoadBalancers: The maximum number of load balancers has been reached `

pkg/apis/kubeone/types.go
```diff
@@ -208,6 +209,33 @@ type PodSecurityPolicy struct {
...
+	Config StaticAuditLogConfig `json:"config"`
+}
+
+// StaticAuditLogConfig config
+type StaticAuditLogConfig struct {
+	// PolicyFilePath is a path on local file system to the audit policy manifest
+	// which defines what events should be recorded and what data they should include.
+	// PolicyFilePath is a required field.
+	// More info: https://kubernetes.io/docs/tasks/debug-application-cluster/audit/#audit-policy
+	PolicyFilePath string `json:"policyFilePath"`
```
kron4eg: aren't you hardcoded it later?

pkg/apis/kubeone/types.go
```diff
@@ -208,6 +209,33 @@ type PodSecurityPolicy struct {
...
+	// LogPath is path on control plane instances where audit log files are stored.
+	// Default value is /var/log/kubernetes/audit.log
+	LogPath string `json:"logPath"`
```
kron4eg: if we hardcode policy path, why not LogPath?

pkg/apis/kubeone/types.go
xmudrii: No, I didn't. As the comment says, this is a file on the local file system (on the system where KubeOne is running). It is hardcoded on instances, but it can be anywhere on the local system.

There were two options: to specify the whole file in the manifest (e.g. like we do for `cloud-config`) or to specify just path and read the file on runtime. After discussion with @eqrx, we decided to go with the latter option because the audit policy file can get very long.

pkg/apis/kubeone/types.go
xmudrii: The reason why not hardcoding it is that I'm not sure will anyone need it in a special place, so it's possible to integrate with the desired tooling.

xmudrii:
/retest
Flake. We should eventually increase timeouts.

pkg/apis/kubeone/types.go
kron4eg: OOOOH, it's a reference to the *local* file, on machine where kubeone is running?

PR #643, opened 2019-09-03
Allow using of hostname instead of fqdn on kubeadm init/join for node…

Nightapes:

During installation of the seed cluster for kubermatic we encountered a issue with kubeone.

Our setup:
* vsphere on premise 
* vsphere cloud provider
* custom ubuntu image (ubuntu + proxy settings)
* DHCP
  * `sh: hostname` -> vmname123
  * `sh: hostname -f` -> vmname123.domain.de

During installation of kubernetes with kubeone, kubeone hangs on `kubeadm init`.
After checking the log of kubectl, we see that the node uses the hostname `sh: hostname` instead of `sh: hostname -f` and can't finish the installation.

This PR adds a flag to use `hostname` instead of  `hostname -f` for the node names.

I added  `--node-name` , so that the --hostname-override is set, but this was not working with `hostname - f` as the node name. I think because of the following issue: https://github.com/kubernetes/kubernetes/issues/54482.

It worked only for us with `--node-name` and `hostname`

The Flags i added, are optional and everything should be backwards compatibly 

Fixes #

```release-note
```

### DIFF
--- a/pkg/cmd/install.go
+++ b/pkg/cmd/install.go
@@ -33,8 +33,10 @@ import (
 
 type installOptions struct {
 	globalOptions
-	Manifest   string
-	BackupFile string
+	Manifest       string
+	BackupFile     string
+	NoFQDN         bool
+	EnableNodeName bool
 }
 
 // installCmd setups install command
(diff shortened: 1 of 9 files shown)


kron4eg:
I think we already have what you want: https://github.com/kubermatic/kubeone/blob/af577ac52356b911aae2f46f36546d5b985b901c/examples/terraform/aws/output.tf#L34

You can specify hostnames directly in api (and thus in terraform outputs).

Nightapes:
@kron4eg Thanks for the information, it's working also with the hostnames in the `output.tf``.

PR #645, opened 2019-09-04
Change control-plane hostname to master on GCP

infinitydon:
What this PR does / why we need it:

Sometimes kubeone install fails on GCP if the GCP project name is long. When the long project name is combined with the cluster-name ("${var.cluster_name}-control-plane-${count.index + 1}") to form the full GCP hostname for the control-plane nodes it can exceed the maximum number of characters that is supported by kubeadm (63 characters). 

Using the name "master" will give more room to use desired cluster_name variable.

Sample error logs from kubeone:

```
WARN[15:24:42 CEST] Task failed, retrying…
Error: failed to init kubernetes on leader: + [[ -f /etc/kubernetes/admin.conf ]]
+ sudo kubeadm init --config=./kubeone/cfg/master_0.yaml
error execution phase upload-config/kubelet: Error writing Crisocket information for the control-plane node: timed out waiting for the condition: failed to exec command: export "PATH=$PATH:/sbin:/usr/local/bin:/opt/bin"

set -xeu pipefail


if [[ -f /etc/kubernetes/admin.conf ]]; then exit 0; fi
sudo kubeadm init --config=./kubeone/cfg/master_0.yaml
```

Sample kubeadm error logs:

```
Jul 18 13:36:42 kubermatic-demo-control-plane-1 kubelet[11953]: E0718 13:36:42.014744   11953 kubelet_node_status.go:94] Unable to register node "kubermatic-demo-control-plane-1.europe-west1-b.c.test-01-kubermatic.internal" with API server: Node "kubermatic-demo-control-plane-1.europe-west1-b.c.test-01-kubermatic.internal" is invalid: metadata.labels: Invalid value: "kubermatic-demo-control-plane-1.europe-west1-b.c.test-01-kubermatic.internal": must be no more than 63 characters
```

Does this PR introduce a user-facing change?:

```
"kubectl get nodes" output will show the name "master" instead of "control-plane"
````

### CODE REVIEW

File: examples/terraform/gce/main.tf [hunk 1]
```diff
@@ -132,7 +132,7 @@ resource "google_compute_forwarding_rule" "control_plane" {
 resource "google_compute_instance" "control_plane" {
   count = 3
 
-  name         = "${var.cluster_name}-control-plane-${count.index + 1}"
+  name         = "${var.cluster_name}-master-${count.index + 1}"
```
xmudrii: When we have such behavior, i.e. the FQDN is way too long, we usually override the hostname. You can see #607 as an example, which should fix the issue you mention for GCP, but it was never merged because I never got E2E tests working (which was the purpose of the PR).

If you want to go with that approach, you can update your PR to set the hostname and I'll rebase later.

infinitydon: Ok.. Which approach do you prefer? Any of the two workarounds is okay for me.

xmudrii: I would go with the setting the hostname approach as we use it for some other providers too.

infinitydon: Ok.. will wait for your change to be merged then. I also support the idea that the behavior should be consistent across the providers.

xmudrii: #607 has been merged to fix this.

PR #655, opened 2019-09-16
Do not add repositories when offline.list exists.

### LINKED ISSUE #654: Allow to preconfigure ubuntu repo sources
In certain environments the "regular" repository sources are not available and alternate sources (mirrors) need to be used. To support supplying a working repository sources configuration in infrastructure creation phase (e.g. terraform) it is required that KubeOne does not add the regular repositories.

It is not sufficient to just add the correct repositories, because the regular ones might break `apt-get update` if they are resolvable but return an error (happens with certain proxy configurations).

One approach would be to only add the regular repository sources if a certain source configuration does not already exist. Creating that exact source configuration during terraforming would then make things work because KubeOne would not break the repository setup by creating the regular repositories.

thz:
* This is specific to debian flavoured machines
* docker+kubernetes repositories are not added when
  /etc/apt/sources.list.d/offline.list exists

Fixes #654 

```release-note
Allow pre-configuration of debian package repositories.
```

### DIFF
--- a/pkg/installer/installation/prerequisites.go
+++ b/pkg/installer/installation/prerequisites.go
@@ -111,13 +111,15 @@ curl -fsSL https://packages.cloud.google.com/apt/doc/apt-key.gpg | \
 curl -fsSL https://download.docker.com/linux/${ID}/gpg | \
 	sudo apt-key add -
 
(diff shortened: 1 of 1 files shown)


kron4eg:
This PR will be superseded by #670

PR #663, opened 2019-09-16
Idempotent install

kron4eg:

make `kubeone install` be able to finish if interrupted.

```release-note
kubeone install is idempotent now
```

### DIFF
--- a/pkg/state/context.go
+++ b/pkg/state/context.go
@@ -17,8 +17,6 @@ limitations under the License.
 package state
 
 import (
-	"sync"
-
 	"github.com/sirupsen/logrus"
 
 	kubeoneapi "github.com/kubermatic/kubeone/pkg/apis/kubeone"
@@ -59,8 +57,6 @@ type State struct {
 	UpgradeMachineDeployments bool
 	PatchCNI                  bool
 	CredentialsFilePath       string
-
-	lock *sync.Mutex
 }
 
 func (s *State) KubeadmVerboseFlag() string {
--- a/pkg/state/context.go
+++ b/pkg/state/context.go
@@ -51,6 +51,13 @@ type State struct {
 	CredentialsFilePath       string
 }
 
+func (s *State) KubeAdmVerboseFlag() string {
+	if s.Verbose {
+		return "--v=6"
+	}
+	return ""
+}
+
 // Clone returns a shallow copy of the State.
 func (s *State) Clone() *State {
 	newState := *s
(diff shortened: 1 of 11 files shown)


pkg/state/context.go
```diff
@@ -49,10 +58,31 @@ type State struct {
 	UpgradeMachineDeployments bool
 	PatchCNI                  bool
 	CredentialsFilePath       string
+
+	lock *sync.Mutex
+}
+
+func (s *State) KubeAdmVerboseFlag() string {
```
xmudrii: This should be `KubeadmVerboseFlag()`.

pkg/state/context.go
```diff
@@ -49,10 +58,31 @@ type State struct {
...
+	if s.Verbose {
+		return "--v=6"
```
xmudrii: Can we make KubeOne take value for `--v`, so we can configure that easily?

pkg/state/context.go
```diff
@@ -49,10 +58,31 @@ type State struct {
...
 
 // Clone returns a shallow copy of the State.
 func (s *State) Clone() *State {
 	newState := *s
+	newState.lock = &sync.Mutex{}
 	return &newState
 }
+
+func GenerateBootstrapToken(s *State) error {
+	s.lock.Lock()
```
xmudrii: What is the reason for the lock here? Is this concurrect somewhere?

pkg/upgrader/upgrade/upgrade.go
```diff
@@ -47,6 +47,7 @@ const (
 func Upgrade(s *state.State) error {
 	// commonSteps are same for all worker nodes and they are safe to be run in parallel
 	commonSteps := []task.Task{
+		{Fn: state.GenerateBootstrapToken, ErrMsg: "failed to generate bootstrap token"},
```
xmudrii: Why do we need this? I don't remember we had any problem requiring us a new token.

pkg/state/context.go
kron4eg: Not currently :) But I now have better idea, so I'd remove this.

pkg/upgrader/upgrade/upgrade.go
kron4eg: Because inside pkg/templates/kubeadm we call for `kubeadmv1beta1.NewBootstrapTokenString(s.JoinToken)` (or `kubeadmv1beta2.NewBootstrapTokenString(s.JoinToken)`) which fails on case when token is empty, even that it's not required for upgrade procedures.

PR #670, opened 2019-09-19
New top-level API SystemPackages

### LINKED ISSUE #654: Allow to preconfigure ubuntu repo sources

kron4eg:
Provides more control over what's happen on machine in strict environments

Fixes #654

```release-note
new systemPackages configuration block
```


kron4eg:
@xmudrii in follow-up PR I'm planning also to add `systemPackages.configureProxy`.

pkg/apis/kubeone/types.go
```diff
@@ -48,6 +48,8 @@ type KubeOneCluster struct {
 	MachineController *MachineControllerConfig `json:"machineController,omitempty"`
 	// Features enables and configures additional cluster features
 	Features Features `json:"features,omitempty"`
+	// SystemPackages configure kubeone behaviour regarding OS packages
+	SystemPackages SystemPackages `json:"systemPackages"`
```
xmudrii: I'd make this a pointer rather than having a bool pointer. From the usability side, I think it's much better approach.

xmudrii: Also for consistency, because we use that approach for machine-controller config too.

pkg/installer/installation/prerequisites.go
```diff
@@ -338,18 +342,20 @@ func installKubeadm(s *state.State, node kubeoneapi.HostConfig) error {
 
 func installKubeadmDebian(s *state.State) error {
 	_, _, err := s.Runner.Run(kubeadmDebianScript, runner.TemplateVariables{
-		"KUBERNETES_VERSION": s.Cluster.Versions.Kubernetes,
-		"DOCKER_VERSION":     dockerVersion,
-		"CNI_VERSION":        s.Cluster.Versions.KubernetesCNIVersion(),
+		"KUBERNETES_VERSION":    s.Cluster.Versions.Kubernetes,
+		"DOCKER_VERSION":        dockerVersion,
+		"CNI_VERSION":           s.Cluster.Versions.KubernetesCNIVersion(),
+		"CONFIGUREREPOSITORIES": s.Cluster.SystemPackages.ConfigureRepositories,
```
xmudrii: I'd say `CONFIGURE_REPOSITORIES` instead.

pkg/apis/kubeone/types.go
kron4eg: Then I wouldn't be able to just
```go
"CONFIGUREREPOSITORIES": s.Cluster.SystemPackages.ConfigureRepositories,
```

I'd need to check if `s.Cluster.SystemPackages == nil`

xmudrii: But what if `s.Cluster.SystemPackages.ConfigureRepositories` is `nil`, wouldn't it NPE? Also, I think UX is more important than an additional check. :wink:

xmudrii: Also, we have defaulting and validation which would ensure that it isn't `nil`.

pkg/cmd/config.go
```diff
@@ -39,7 +39,7 @@ import (
 
 const (
 	// defaultKubernetesVersion is default Kubernetes version for the example configuration file
-	defaultKubernetesVersion = "1.14.1"
+	defaultKubernetesVersion = "1.16.0"
```
thz: This should not go in this PR.

xmudrii: Why? This is just for the example config, i.e. `kubeone config print -f`.

thz: Because it is not related to the purpose of this PR.

thz: And even more so because that change should go together in a change which updates the README about supported versions, perhaps the occurrences of `1.14.1` in the testdata manifests, and more.

pkg/apis/kubeone/types.go
```diff
@@ -205,6 +207,13 @@ type Features struct {
 	OpenIDConnect     *OpenIDConnect     `json:"openidConnect"`
 }
 
+// SystemPackages controls configurations of APT/YUM
+type SystemPackages struct {
+	// ConfigureRepositories (true by default) is a flag to control automatic
+	// configuration of kubeadm / docker repositories.
+	ConfigureRepositories bool `json:"configureRepositories,omitempty"`
```
xmudrii: Do we need `omitempty` here as this is not a pointer?

pkg/apis/kubeone/types.go
```diff
@@ -205,6 +207,13 @@ type Features struct {
 	OpenIDConnect     *OpenIDConnect     `json:"openidConnect"`
 }
 
+// SystemPackages controls configurations of APT/YUM
+type SystemPackages struct {
```
xmudrii: What if we call this `SystemConfiguration` or maybe even `HostConfiguration` (so it's matching `hosts`), so we can easily re-use this field for other purposes without having to bump the API or add new fields?

pkg/templates/kubeadm/kubeadm.go
```diff
@@ -28,6 +28,7 @@ var (
 	v13x = mustParseConstraint("1.13.x")
 	v14x = mustParseConstraint("1.14.x")
 	v15x = mustParseConstraint("1.15.x")
+	v16x = mustParseConstraint("1.16.x")
```
xmudrii: Can we simplify this with something like `>= 1.15.x`?

pkg/cmd/config.go
```diff
@@ -519,6 +519,11 @@ features:
       # be used.
       caFile: ""
 
+systemPackages:
+  # if set (true by default), will configure OS package manager with
```
xmudrii: I'd maybe say `will add Docker and Kubernetes repositories to OS package manager` instead.

pkg/apis/kubeone/types.go
kron4eg: This API is only about packages, not system or host in general.

pkg/templates/kubeadm/kubeadm.go
kron4eg: It's done this way to be future-proof. `>= 1.15.x` is moving target (will be valid for anything after 1.15), while  `1.16.x` is constrained.

PR #676, opened 2019-09-23
Add build options to the Makefile

eqrx:
The path is trimmed from binaries to make builds more reproducible.
Disable lazy binding and make global offset table read only to
harden binary a bit.

Enhances security of builds and makes them more reproducible.

Fixes #

None

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,6 +67,7 @@ lint:
 	golangci-lint run ./pkg/... ./test/...
 
 .PHONY: verify-licence
+verify-licence: GOFLAGS = -mod=readonly
 verify-licence: vendor
 	wwhrd check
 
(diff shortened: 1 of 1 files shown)


kron4eg:
AFAIK trimpath can be used without actual path. And it’s Go 1.13 only. If we’re ok with this then lgtm.

xmudrii:
> AFAIK trimpath can be used without actual path. And it’s Go 1.13 only. If we’re ok with this then lgtm.
I'm okay with this. If Go 1.13 is a problem, you can always use `go install` and `go build` manually.

xmudrii:
/hold
I want to confirm something first

eqrx:
Have changed the Makefile to use -trimpath without arguments. Thanks for noticing

eqrx:
@kron4eg yeah well, the build system cannot cope with that since its <1.13 ... should we upgrade here?

xmudrii:
This is blocked until we don't fully switch to Go 1.13.

PR #687, opened 2019-09-25
Add proxy env to /etc/environment

thz:
Add proxy env to /etc/environment

`#kubeone` markers are used to allow reset and idempotency

```release-note
NONE
```

### DIFF
--- a/pkg/installer/installation/prerequisites.go
+++ b/pkg/installer/installation/prerequisites.go
@@ -76,8 +76,12 @@ export NO_PROXY no_proxy
 
 {{ end }}
 EOF
-	`
 
+envtmp=/tmp/k1-etc-environment
+grep -v '#kubeone$' /etc/environment > $envtmp
+grep = /etc/kubeone/proxy-env | sed 's/$/#kubeone/' >> $envtmp
+sudo tee /etc/environment < $envtmp
+`
 	kubeadmDebianScript = `
 sudo swapoff -a
 sudo sed -i '/.*swap.*/d' /etc/fstab


kron4eg:
So what’s the rationale behind this?

thz:
It makes sure that all shell-sessions on the system will have the specified proxy environment. Kubeadm especially needs to pass that through to supply kube-controller-manager and others with the environment variables. I would argue that it is the right thing to do to supply proxy environment to the whole machine.

kron4eg:
I mean, /etc/environment is just a convention and is voluntary to source, so what’s the big difference compared to existing kubeone specific file?

Remember that control plane machines are not foe general unix usage, they have single responsibility to host CP components.

P.S.
I wonder what machine-controller doing on this regard.

thz:
`/etc/environment` is more than just a convention. it is not even meant for being sourced. that is why there is no export statement for the variables, which would be needed to make it useful for sourcing.

`/etc/environment` is the default key-value file of the pam module pam_env. that mechanism makes sure that pam injects the variables to all "sessions".

thz:
> P.S.
> I wonder what machine-controller doing on this regard.

machine-controller is writing /etc/environment by cloud-init directives.

kron4eg:
Maybe then let’s replace kubeone specific env with /etc/environment?

thz:
> Maybe then let’s replace kubeone specific env with /etc/environment?

I thought about that too. I think it is good to have /etc/kubeone/proxy-env as a "transport". Also /etc/environment needs you to re-open your shell session to get the environment. For now I would say keeping both is the better choice.

kron4eg:
/hold cancel 

No, it good.

PR #692, opened 2019-10-02
Force non-interactive mode for dpkg via apt-get.

thz:
This fixes ubuntu based `apt-get install` calls hanging for
some dpkg interactivity.
Upstream issue reference:
https://bugs.launchpad.net/ubuntu/+source/ansible/+bug/1833013

```release-note
Fix hanging ubuntu setup (make dpkg non-interactive).
```


kron4eg:
This will influence interactive sessions...

PR #714, opened 2019-10-24
Divert kube-api access via SSH tunnel

### LINKED ISSUE #659: Divert kubeapi access to private loadbalancer via SSH port forwarding

kron4eg:
KubeOne operator may create kube-API loabdalancer as internal, only accessible from inside the VPC or internal-perimeter, so KubeOne will divert all kube-API access over SSH tunnel from the leader node which guaranteed to have direct kube-api access (since kubelet need to communicate with it).

Fixes #659 

```release-note
kube-api communications are going via the SSH tunnel
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -19,6 +19,7 @@ linters:
   - goconst
   - gocritic
   - gocyclo
+  - golint
   - gofmt
   - goimports
   - gosec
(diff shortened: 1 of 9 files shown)


xmudrii:
Those look like some weird flakes.
/retest

PR #722, opened 2019-10-29
Update DigitalOcean Cloud Controller Manager to v0.1.21

xmudrii:

Update the DigitalOcean Cloud Controller Manager to v0.1.21.

```release-note
Update the DigitalOcean Cloud Controller Manager to v0.1.21.
```

/assign @kron4eg

### DIFF
--- a/pkg/templates/externalccm/digitalocean.go
+++ b/pkg/templates/externalccm/digitalocean.go
@@ -34,7 +34,7 @@ import (
 )
 
 const (
-	digitaloceanCCMVersion     = "v0.1.16"
+	digitaloceanCCMVersion     = "v0.1.21"
 	digitaloceanSAName         = "cloud-controller-manager"
 	digitaloceanDeploymentName = "digitalocean-cloud-controller-manager"
 )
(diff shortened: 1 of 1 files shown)


pkg/templates/externalccm/digitalocean.go
```diff
@@ -174,6 +174,8 @@ func doDeployment() *appsv1.Deployment {
 					},
 				},
 				Spec: corev1.PodSpec{
+					DNSPolicy:          corev1.DNSDefault,
+					HostNetwork:        true,
```
kron4eg: does it really need hostnetwork?

xmudrii: I'm not sure at all. I added it since it's present in the manifest https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/16c082483c473e72bee2ec188894bbc8d8d3f753/releases/v0.1.20.yml#L17-L18.

kron4eg: we overcome problem of circular dependency (CCM vs CNI) with:

```
      - effect: NoSchedule
        key: node.cloudprovider.kubernetes.io/uninitialized
        value: "true"
```

xmudrii: Fixed.

PR #727, opened 2019-10-31
Fix kubeone reset

kron4eg:
```release-note
Fix kubeone reset
```

### DIFF
--- a/pkg/templates/externalccm/ccm.go
+++ b/pkg/templates/externalccm/ccm.go
@@ -69,7 +69,7 @@ func waitForInitializedNodes(s *state.State) error {
 
 	s.Logger.Info("Waiting for nodes to initialize by CCM…")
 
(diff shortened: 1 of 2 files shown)


pkg/templates/externalccm/ccm.go
```diff
@@ -69,7 +69,7 @@ func waitForInitializedNodes(s *state.State) error {
 
 	s.Logger.Info("Waiting for nodes to initialize by CCM…")
 
-	return wait.Poll(5*time.Second, 60*time.Second, func() (bool, error) {
+	return wait.Poll(5*time.Second, 10*time.Minute, func() (bool, error) {
```
xmudrii: Why did you increase the timeout this much?

kron4eg: Sometimes cloud API is flaky (like in DO case, but could happen to anyone).

xmudrii:
/hold
I want to discuss something.

PR #734, opened 2019-11-14
Implement components health checks and the status command

### LINKED ISSUE #589: Add a status subcommand
`kubeone status [...]` shall print out information about the cluster. This should include the Kubernetes version that was installed and if an upgrade is possible.

This could be used to automate certain deployments (like with ansible) easier to handle

xmudrii:

The focus of this PR is:
* Implementing the health checks for API server and etcd. Those checks are currently used for the status command and will be used for repairs in the future,
* Implementing the status command that wraps those checks

The status command is used similar to other KubeOne commands: `kubeone status config.yaml -t .`. The output of the command looks like:
```
INFO[13:23:24 CET] Building Kubernetes clientset…               
INFO[13:23:27 CET] Verifying that Docker, Kubelet and Kubeadm are installed… 
INFO[13:23:28 CET] Verifying that nodes in the cluster match nodes defined in the manifest… 
INFO[13:23:28 CET] Verifying that all nodes in the cluster are ready… 
INFO[13:23:28 CET] Verifying that there is no upgrade in the progress… 
NODE                                          VERSION   APISERVER   ETCD      
ip-172-31-8-28.eu-west-3.compute.internal     v1.16.3   healthy     healthy   
ip-172-31-23-254.eu-west-3.compute.internal   v1.16.3   healthy     healthy   
ip-172-31-39-85.eu-west-3.compute.internal    v1.16.3   healthy     healthy   
```

* API server being healthy means that `/healthz` endpoint is returning `ok` with status code 200.
* Etcd being healthy means that `/health` returns true and the node is a member of the etcd cluster.

Fixes #589

```release-note
Implement health checks for API server and etcd
Add the 'kubeone status' command
```

/assign @kron4eg

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,6 @@ go 1.13
 require (
 	github.com/Masterminds/semver v1.4.2
 	github.com/aws/aws-sdk-go v1.20.15
-	github.com/coreos/etcd v3.3.13+incompatible
 	github.com/ghodss/yaml v1.0.0
 	github.com/go-logr/zapr v0.1.1 // indirect
 	github.com/gogo/protobuf v1.2.1 // indirect
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.13
 require (
 	github.com/Masterminds/semver v1.4.2
 	github.com/aws/aws-sdk-go v1.20.15
+	github.com/coreos/etcd v3.3.13+incompatible
 	github.com/ghodss/yaml v1.0.0
 	github.com/go-logr/zapr v0.1.1 // indirect
 	github.com/gogo/protobuf v1.2.1 // indirect
@@ -12,6 +13,7 @@ require (
 	github.com/json-iterator/go v1.1.7 // indirect
 	github.com/koron-go/prefixw v0.0.0-20181013140428-271b207a7572
 	github.com/kr/fs v0.1.0 // indirect
+	github.com/liggitt/tabwriter v0.0.0-20181228230101-89fcab3d43de
 	github.com/onsi/ginkgo v1.8.0 // indirect
 	github.com/onsi/gomega v1.5.0 // indirect
 	github.com/pkg/errors v0.8.1
--- a/pkg/clusterstatus/apiserver/apiserver.go
+++ b/pkg/clusterstatus/apiserver/apiserver.go
@@ -37,11 +37,11 @@ type Status struct {
 
 // CheckAPIServer uses the /healthz endpoint to check are all API server instances healthy
 func GetStatus(s *state.State, node kubeoneapi.HostConfig) (*Status, error) {
-	client, err := httptunnel.NewHTTPTunnel(s, &tls.Config{InsecureSkipVerify: true})
+	tunneler, err := httptunnel.NewHTTPTunnel(s, &tls.Config{InsecureSkipVerify: true})
 	if err != nil {
 		return nil, err
 	}
-	health, err := apiserverHealth(client, node.PrivateAddress)
+	health, err := apiserverHealth(tunneler, node.PrivateAddress)
 	if err != nil {
 		return nil, err
 	}
@@ -52,14 +52,14 @@ func GetStatus(s *state.State, node kubeoneapi.HostConfig) (*Status, error) {
 }
 
 // apiserverHealth checks is API server healthy
-func apiserverHealth(t *httptunnel.HTTPTunnel, nodeAddress string) (bool, error) {
+func apiserverHealth(t httptunnel.HTTPDoer, nodeAddress string) (bool, error) {
 	endpoint := fmt.Sprintf(healthzEndpoint, nodeAddress)
 	request, err := http.NewRequest("GET", endpoint, nil)
 	if err != nil {
 		return false, err
 	}
 
-	resp, err := t.Client.Do(request)
+	resp, err := t.Do(request)
 	if err != nil {
 		return false, nil
 	}
--- a/pkg/clusterstatus/apiserver/apiserver.go
+++ b/pkg/clusterstatus/apiserver/apiserver.go
@@ -1,0 +1,91 @@
+/*
+Copyright 2019 The KubeOne Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package apiserver
+
+import (
+	"crypto/tls"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"time"
+
+	kubeoneapi "github.com/kubermatic/kubeone/pkg/apis/kubeone"
+	"github.com/kubermatic/kubeone/pkg/ssh"
+)
+
+const (
+	defaultHTTPTimeout = 10 * time.Second
+	healthzEndpoint    = "https://%s:6443/healthz"
+)
+
+type Status struct {
+	Health bool `json:"health,omitempty"`
+}
+
+type HTTPDoer interface {
+	Do(*http.Request) (*http.Response, error)
+}
+
+// CheckAPIServer uses the /healthz endpoint to check are all API server instances healthy
+func GetStatus(node kubeoneapi.HostConfig, tunneler ssh.Tunneler) (*Status, error) {
+	client := httpClient(tunneler)
+	health, err := apiserverHealth(client, node.PrivateAddress)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Status{
+		Health: health,
+	}, nil
+}
+
+// apiserverHealth checks is API server healthy
+func apiserverHealth(c HTTPDoer, nodeAddress string) (bool, error) {
+	endpoint := fmt.Sprintf(healthzEndpoint, nodeAddress)
+	request, err := http.NewRequest("GET", endpoint, nil)
+	if err != nil {
+		return false, err
+	}
+
+	resp, err := c.Do(request)
+	if err != nil {
+		return false, nil
+	}
+	defer resp.Body.Close()
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return false, err
+	}
+
+	return string(body) == "ok", nil
+}
+
+// httpClient builds an HTTP client used to access the API server
+func httpClient(tunneler ssh.Tunneler) HTTPDoer {
+	transport := &http.Transport{
+		DialContext: tunneler.TunnelTo,
+		TLSClientConfig: &tls.Config{
+			InsecureSkipVerify: true,
+		},
+	}
+
+	return &http.Client{
+		Timeout:   defaultHTTPTimeout,
+		Transport: transport,
+	}
+}
--- a/pkg/clusterstatus/apiserver/apiserver.go
+++ b/pkg/clusterstatus/apiserver/apiserver.go
@@ -21,28 +21,26 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
-	"time"
 
 	kubeoneapi "github.com/kubermatic/kubeone/pkg/apis/kubeone"
-	"github.com/kubermatic/kubeone/pkg/ssh"
+	"github.com/kubermatic/kubeone/pkg/httptunnel"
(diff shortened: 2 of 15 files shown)


pkg/clusterstatus/apiserver/apiserver.go
```diff
@@ -0,0 +1,91 @@
...
+
+	return string(body) == "ok", nil
+}
+
+// httpClient builds an HTTP client used to access the API server
+func httpClient(tunneler ssh.Tunneler) HTTPDoer {
+	transport := &http.Transport{
+		DialContext: tunneler.TunnelTo,
+		TLSClientConfig: &tls.Config{
+			InsecureSkipVerify: true,
```
xmudrii: For some reason, this endpoint only works with `InsecureSkipVerify` set to `true`. If I provide certs I get an error saying `Unauthorized` that looks like:
```
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {
    
  },
  "status": "Failure",
  "message": "Unauthorized",
  "reason": "Unauthorized",
  "code": 401
}
```

go.mod
```diff
@@ -5,13 +5,15 @@ go 1.13
 require (
 	github.com/Masterminds/semver v1.4.2
 	github.com/aws/aws-sdk-go v1.20.15
+	github.com/coreos/etcd v3.3.13+incompatible
```
xmudrii: This one is not used anywhere and I don't have any idea how it ended up there...

pkg/clusterstatus/apiserver/apiserver.go
kron4eg: @xmudrii probably because you've tried to use etcd client keys to access apiserver endpoint :)

go.mod
kron4eg: try running `go mod tidy`

xmudrii: I tried (and pushed a commit), but there is still an entry in `go.sum`.

pkg/clusterstatus/apiserver/apiserver.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+const (
+	defaultHTTPTimeout = 10 * time.Second
+	healthzEndpoint    = "https://%s:6443/healthz"
+)
+
+type Status struct {
+	Health bool `json:"health,omitempty"`
+}
```
xmudrii: I'm thinking about getting rid of this struct and just returning bool from a function.

pkg/clusterstatus/apiserver/apiserver.go
```diff
@@ -0,0 +1,91 @@
...
+	endpoint := fmt.Sprintf(healthzEndpoint, nodeAddress)
+	request, err := http.NewRequest("GET", endpoint, nil)
+	if err != nil {
+		return false, err
+	}
+
+	resp, err := c.Do(request)
+	if err != nil {
+		return false, nil
+	}
```
xmudrii: Eventually, I should do some advanced error checking, such as checking the status code for 200.

pkg/clusterstatus/apiserver/apiserver.go
xmudrii: No, I have used the API server keys. Here is the `curl` command that also fails:
```
sudo curl --cacert /etc/kubernetes/pki/ca.crt --cert /etc/kubernetes/pki/apiserver.crt --key /etc/kubernetes/pki/apiserver.key https://<endpoint>:6443/healthz
```

pkg/clusterstatus/clusterstatus.go
```diff
@@ -0,0 +1,152 @@
...
+	Etcd      bool   `json:"etcd,omitempty"`
+}
+
+func PrintClusterStatus(s *state.State) error {
+	status, err := GetClusterStatus(s)
+	if err != nil {
+		return errors.Wrap(err, "unable to get cluster status")
+	}
+
+	printer := tabwriter.GetNewTabWriter(os.Stdout)
```
xmudrii: We should consider a follow-up to add additional formats, such as JSON.

pkg/clusterstatus/clusterstatus.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	err := s.DynamicClient.List(context.Background(), &nodes, &nodeListOpts)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to list nodes")
+	}
+
+	// Run preflight checks
+	if preflightErr := preflight.RunPreflightChecks(s, nodes); err != nil {
+		return nil, preflightErr
+	}
```
xmudrii: Preflight checks should be part of the Status API and we should show it in the table, but this requires more changes. I'm thinking about leaving it for a follow-up.

pkg/clusterstatus/apiserver/apiserver.go
kron4eg: this is server cert, but you need client

pkg/clusterstatus/etcd/etcd.go
```diff
@@ -0,0 +1,228 @@
...
+	request, err := http.NewRequest("GET", endpoint, nil)
+	if err != nil {
+		return nil, err
+	}
+	request.Header.Set("Content-type", "application/json")
+
+	resp, err := c.Do(request)
+	if err != nil {
+		return &healthRaw{Health: "false"}, nil
+	}
```
xmudrii: The problem here is if `etcd` is not working, this is going to fail and cause everything to fail instead of not showing `unhealthy`. I should improve error handling, but I'm wondering should I leave it like this for now and fix it in a follow-up to make PR smaller or just go ahead?

pkg/clusterstatus/apiserver/apiserver.go
kron4eg: I think it's OK to keep this as a struct, probably we going to add some fields there (some metrics/stats?).

pkg/clusterstatus/apiserver/apiserver.go
```diff
@@ -0,0 +1,74 @@
...
+	healthzEndpoint = "https://%s:6443/healthz"
+)
+
+type Status struct {
+	Health bool `json:"health,omitempty"`
+}
+
+// CheckAPIServer uses the /healthz endpoint to check are all API server instances healthy
+func GetStatus(s *state.State, node kubeoneapi.HostConfig) (*Status, error) {
+	client, err := httptunnel.NewHTTPTunnel(s, &tls.Config{InsecureSkipVerify: true})
```
kron4eg: it's not a client, it's a tunneler

pkg/clusterstatus/apiserver/apiserver.go
```diff
@@ -0,0 +1,74 @@
...
+	if err != nil {
+		return nil, err
+	}
+	health, err := apiserverHealth(client, node.PrivateAddress)
```
kron4eg: here instead of client (tunneler) you can pass down actual client, because apiserverHealth() doesn't do anything with tunneler except calling .Client.Do()

pkg/clusterstatus/apiserver/apiserver.go
```diff
@@ -0,0 +1,74 @@
...
+		return nil, err
+	}
+
+	return &Status{
+		Health: health,
+	}, nil
+}
+
+// apiserverHealth checks is API server healthy
+func apiserverHealth(t *httptunnel.HTTPTunnel, nodeAddress string) (bool, error) {
```
kron4eg: please replace `httptunnel.HTTPTunnel` with ether http.Client or even better `Doer` interface

pkg/clusterstatus/clusterstatus.go
kron4eg: 👍

pkg/clusterstatus/etcd/etcd.go
```diff
@@ -0,0 +1,212 @@
...
+type membersListRaw struct {
+	Members []struct {
+		ID         string   `json:"id,omitempty"`
+		Name       string   `json:"name,omitempty"`
+		PeerURLs   []string `json:"peerURLs,omitempty"`
+		ClientURLs []string `json:"clientURLs,omitempty"`
+	} `json:"members"`
+}
+
+type HTTPDoer interface {
```
kron4eg: we need to find better home for this interface

pkg/clusterstatus/etcd/etcd.go
```diff
@@ -0,0 +1,212 @@
...
+	Do(*http.Request) (*http.Response, error)
+}
+
+// EtcdStatus analyzes health of an etcd cluster
+func GetStatus(s *state.State, node kubeoneapi.HostConfig) (*Status, error) {
+	tlsCfg, err := loadTLSConfig(s)
+	if err != nil {
+		return nil, err
+	}
+	client, err := httptunnel.NewHTTPTunnel(s, tlsCfg)
```
kron4eg: same comments applies here as in pkg/clusterstatus/apiserver/apiserver.go

pkg/clusterstatus/etcd/etcd.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	return &Status{
+		Health: health,
+		Member: member,
+	}, nil
+}
+
+// memberHealth returns health for a requested etcd member
+func memberHealth(t *httptunnel.HTTPTunnel, nodeAddress string) (*healthRaw, error) {
```
kron4eg: same as in pkg/clusterstatus/apiserver/apiserver.go, tunneler -> http.Client or even better — an interface.

pkg/httptunnel/httptunnel.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/kubermatic/kubeone/pkg/state"
+)
+
+const (
+	defaultHTTPTimeout = 10 * time.Second
+)
+
+type HTTPTunnel struct {
+	Client HTTPDoer
```
kron4eg: I suppose if we embed http.Client, then HTTPTunnel will automatically satisfy HTTPDoer interface.

pkg/httptunnel/httptunnel.go
```diff
@@ -0,0 +1,58 @@
...
+func NewHTTPTunnel(s *state.State, tlsConfig *tls.Config) (*HTTPTunnel, error) {
+	tunn, err := s.Connector.Tunnel(s.Cluster.RandomHost())
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get SSH tunnel")
+	}
+
+	transport := &http.Transport{
+		DialContext:     tunn.TunnelTo,
+		TLSClientConfig: tlsConfig,
+	}
```
kron4eg:
```go
&http.Transport{
    DialContext:           tunn.TunnelTo,
    TLSClientConfig:       tlsConfig,
    MaxIdleConns:          100,
    IdleConnTimeout:       90 * time.Second,
    TLSHandshakeTimeout:   10 * time.Second,
    ExpectContinueTimeout: 1 * time.Second,
}
```

pkg/clusterstatus/etcd/etcd.go
kron4eg: turns out, it's already in httptunneler package

pkg/clusterstatus/apiserver/apiserver.go
xmudrii: :+1:

pkg/clusterstatus/apiserver/apiserver.go
xmudrii: I'm now using the embedding approach you mentioned below.

pkg/clusterstatus/apiserver/apiserver.go
xmudrii: I went with the `Doer` interface.

pkg/httptunnel/httptunnel.go
xmudrii: Yes, that's right. :+1: 

I've updated the PR to reflect this.

pkg/httptunnel/httptunnel.go
```diff
@@ -0,0 +1,58 @@
...
+		TLSClientConfig:       tlsConfig,
+		MaxIdleConns:          100,
+		IdleConnTimeout:       90 * time.Second,
+		TLSHandshakeTimeout:   10 * time.Second,
+		ExpectContinueTimeout: 1 * time.Second,
+	}
+
+	return &HTTPTunnel{
+		Client: &http.Client{
+			Timeout:   10 * time.Second,
```
xmudrii: @kron4eg Do we still need this timeout in combination with the above timeouts?

pkg/httptunnel/httptunnel.go
xmudrii: :+1:

pkg/httptunnel/httptunnel.go
kron4eg: I suppose no

xmudrii: Removed.

PR #760, opened 2019-12-12
Fix overwriting proxy settings

kron4eg:
in #698 we introduced a bug that unconditionally overwrite proxy settings sourced from the terraform, EVEN IF THEY ARE EMPTY.
That means, if there are proxy settings configured in config, they will be effectively nullified, which is a bug.

Proxy settings defined in kubeone manifest have precedence over those defined in terraform output.

```release-note
Fix terraform proxy output handling
Proxy settings defined in kubeone manifest have precedence over those defined in terraform output.
```

### DIFF
--- a/pkg/terraform/config.go
+++ b/pkg/terraform/config.go
@@ -19,6 +19,7 @@ package terraform
 import (
 	"encoding/json"
 
+	"github.com/imdario/mergo"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


kron4eg:
/override pull-kubeone-e2e-aws-upgrade-1.16-1.17

upgrade currently is broken

PR #764, opened 2019-12-16
Ensure correct order of operations for cluster upgrades

xmudrii:

This PR ensures we use the correct order of operation when upgrading a cluster. The correct order is stated in [the upstream kubeadm documentation](https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/).

This is going to fix 1.16 to 1.17 upgrades, but 1.13 to 1.14 upgrades will not work anymore.

```release-note
It is possible to upgrade 1.16 to 1.17
It is NOT possible to upgrade 1.13 to 1.14 anymore
```

### DIFF
--- a/pkg/scripts/node.go
+++ b/pkg/scripts/node.go
@@ -1,0 +1,39 @@
+/*
+Copyright 2019 The KubeOne Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package scripts
+
+const (
+	drainNodeScriptTemplate = `
+kubectl drain {{ .NODE_NAME }} --ignore-daemonsets
+`
+
+	cordonNodeScriptTemplate = `
+kubectl cordon {{ .NODE_NAME }}
+`
+)
+
+func DrainNode(nodeName string) (string, error) {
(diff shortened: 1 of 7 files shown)


pkg/scripts/node.go
```diff
@@ -0,0 +1,39 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package scripts
+
+const (
+	drainNodeScriptTemplate = `
+kubectl drain {{ .NODE_NAME }} --ignore-daemonsets
```
xmudrii: Maybe `--ignore-daemonsets` should be optional, i.e. guarded with a flag, but I don't think that's needed right now.

pkg/scripts/os.go
```diff
@@ -246,22 +242,65 @@ RELEASE="v{{ .KUBERNETES_VERSION }}"
...
+sudo mv /var/tmp/kube-binaries/kubeadm .
+sudo chmod +x kubeadm
 
 sudo mkdir -p /etc/systemd/system/kubelet.service.d
 curl -sSL "https://raw.githubusercontent.com/kubernetes/kubernetes/${RELEASE}/build/debs/10-kubeadm.conf" |
 	sed "s:/usr/bin:/opt/bin:g" |
 	sudo tee /etc/systemd/system/kubelet.service.d/10-kubeadm.conf
+
+sudo systemctl daemon-reload
+sudo systemctl start kubelet
```
xmudrii: Even that this is related to `kubeadm`, I guess this should go to another script (kubelet-related one).

kron4eg: definitely

pkg/scripts/node.go
kron4eg: ignore-daemonsets is needed, otherwise it will fail and block the whole thing. so it's OK.

pkg/scripts/os.go
xmudrii: Fixed and rebased the PR.

PR #774, opened 2020-01-14
MachineController v1.9.0

kron4eg:
Upgrade machine-controller to the latest v1.9.0

```release-note
Upgrade machine-controller to the latest v1.9.0
```


xmudrii:
@kron4eg There is a new field added in the AWS spec in https://github.com/kubermatic/machine-controller/pull/663. Should we add it to KubeOne too?

kron4eg:
@xmudrii we can do this in follow-up PR.

PR #782, opened 2020-02-06
Implement Addons

xmudrii:

This PR implements the KubeOne Addons, as described in the [Addons proposal](https://github.com/kubermatic/kubeone/blob/4e8fe167b1bbc605e8ad99f6334f97d26f6caa4e/docs/proposals/20200205-addons.md).

The documentation will be done in a follow-up PR.

```release-note
Implement Addons
```

/assign @kron4eg 
/hold


pkg/addons/addons.go
```diff
@@ -0,0 +1,70 @@
...
+
+func Ensure(s *state.State) error {
+	if s.Cluster.Addons == nil || !s.Cluster.Addons.Enable {
+		s.Logger.Infoln("Skipping applying addons because addons are not enabled…")
+		return nil
+	}
+	s.Logger.Infoln("Applying addons…")
+
+	if err := getManifests(s); err != nil {
+		return err
```
kron4eg: please use `errors.WithStack(err)` at least

pkg/addons/manifest.go
```diff
@@ -0,0 +1,173 @@
...
+	files, err := ioutil.ReadDir(addonsPath)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to read the addons directory %s", addonsPath)
+	}
+
+	for _, file := range files {
+		filePath := filepath.Join(addonsPath, file.Name())
+		if file.IsDir() {
+			fmt.Printf("Found directory '%s' in the addons path. Ignoring.\n", file.Name())
+			continue
```
kron4eg: I'm not really sure if we need to limit addons to flat structure of files

xmudrii: I'm worried that it would just introduce unnecessary complexity. A big problem is in which order to apply manifests:
* should we first apply manifests from directories,
* should we first apply manifests from the root directory and then from other directories
* what about subdirectories (i.e. directory in the directory)

IMO this is way too complex and not worth it at all. With the flat structure of files, you can easily control the order just by setting the appropriate filename. This will not be that easy if we include directories too. Anyways, usually there is only one manifest per component since YAML allows multiple resources to be defined in a single manifest.

pkg/addons/manifest.go
```diff
@@ -0,0 +1,173 @@
...
+	metav1unstructured "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	kyaml "k8s.io/apimachinery/pkg/util/yaml"
+)
+
+func getManifests(s *state.State) error {
+	vars := map[string]interface{}{
+		"KubeOne": s.Cluster,
+	}
+	manifests, err := loadAddonsManifests(s.Cluster.Addons.Path, s.Verbose, vars)
```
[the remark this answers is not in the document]
xmudrii: I'm wondering should we have a CLI flag instead of the API structure. For example, something such as `kubeone install config.yaml --addons ./addons`. Maybe it would improve the user experience, as we would avoid situations such as that one. WDYT about it?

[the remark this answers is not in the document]
kron4eg: Hm, and the fact of setting `--addons` by the called can be treated as a signal to enable addons handling, and wise-versa, lack of this flag means — don't do anything about addons.

pkg/addons/addons.go
```diff
@@ -0,0 +1,85 @@
...
+	return nil
+}
+
+func applyAddons(s *state.State) error {
+	err := s.RunTaskOnLeader(runKubectl)
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
kron4eg: Replacing body of this function with `return s.RunTaskOnLeader(runKubectl)` would do exactly the same. Did you intended to use `errors.Wrap(err)` but forgot? Then it will be 
`return errors.Wrap(s.RunTaskOnLeader(runKubectl), "failed to apply addons")`

xmudrii: Fixed. PTAL.

pkg/addons/manifest.go
kron4eg: OK, let's have it like this for now.

pkg/cmd/config.go
```diff
@@ -522,6 +522,10 @@ systemPackages:
   # will add Docker and Kubernetes repositories to OS package manager
   configureRepositories: true # it's true by default
 
+addons:
+  enable: false
+  path: "./addons"
```
kron4eg: Please mention here the fact that this path will be relative (in case when it's relative) to the config file itself.

PR #783, opened 2020-02-07
ARM64 support

### LINKED ISSUE #773: ARM Support
Enable KubeOne to support ARM

kron4eg:
This is limited ARM64 support. Limited because there is no machine-controller for arm64.

Fixes #773

* flannel image was replaced by our mirror at quay.io/kubermatic/coreos_flannel for multi-arch support
* metrics-server image was replaces by our mirror at quay.io/kubermatic/metrics-server-mirror for multi-arch support

Mirrors are needed because upstream repositories doesn't have support for multi-arch manifests.

In follow-up PR I'll be adding new API field(s) to regulate master taints.

```release-note
ARM support
```

### DIFF
--- a/pkg/templates/canal/canal.go
+++ b/pkg/templates/canal/canal.go
@@ -33,7 +33,7 @@ import (
 const (
 	installCNIImage = "calico/cni:v3.10.0"
 	calicoImage     = "calico/node:v3.10.0"
-	flannelImage    = "quay.io/coreos/flannel:v0.11.0"
+	flannelImage    = "quay.io/kubermatic/coreos_flannel:v0.11.0@sha256:3de983d62621898fe58ffd9537a4845c7112961a775efb205cab56e089e163b6"
 
 	// cniNetworkConfig configures installation on the each node. The special values in this config will be
 	// automatically populated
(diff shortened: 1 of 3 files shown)


pkg/templates/canal/canal.go
```diff
@@ -33,7 +33,7 @@ import (
 const (
 	installCNIImage = "calico/cni:v3.10.0"
 	calicoImage     = "calico/node:v3.10.0"
-	flannelImage    = "quay.io/coreos/flannel:v0.11.0"
+	flannelImage    = "quay.io/kubermatic/coreos_flannel:v0.11.0@sha256:3de983d62621898fe58ffd9537a4845c7112961a775efb205cab56e089e163b6"
```
xmudrii: I'd prefer that we use the original image as default if that's possible and mirror only if that's not possible. Using the custom image is going to make some users suspicious (which is totally okay and understandable).

I'm aware that there is SHA256, but I'd still prefer if we use the original image if possible. :slightly_smiling_face:

kron4eg: It's not possible to use originals, because there is no multi-arch manifest pushed to upstream.

xmudrii: You are right. We don't have arch available in the API, so there is no way to check.

PR #791, opened 2020-02-17
Fix apt proxy configuration on ubunt/debian

erdii:
**What this PR does / why we need it**: this pr fixes apt proxy configuration
before - the templated script would generate:

```
cat <<EOF | sudo tee /etc/apt/apt.conf.d/proxy.confAcquire::https::Proxy = "{{ .HTTPS_PROXY }}";
```

which broke the setup when one enabled proxy support in k1

Fixes #

```release-note
Fix broken apt config file generation when http/https proxies are enabled. 
```

### DIFF
--- a/pkg/scripts/os.go
+++ b/pkg/scripts/os.go
@@ -55,12 +55,12 @@ EOF
 
 sudo mkdir -p /etc/apt/apt.conf.d
 cat <<EOF | sudo tee /etc/apt/apt.conf.d/proxy.conf
-{{- if .HTTPS_PROXY }}
+{{ if .HTTPS_PROXY -}}
 Acquire::https::Proxy "{{ .HTTPS_PROXY }}";
 {{ end -}}
 {{- if .HTTP_PROXY -}}
 Acquire::http::Proxy "{{ .HTTP_PROXY }}";
-{{ end -}}
+{{ end }}
 EOF
 
 sudo apt-get update
--- a/pkg/scripts/os.go
+++ b/pkg/scripts/os.go
@@ -55,12 +55,12 @@ EOF
 
 sudo mkdir -p /etc/apt/apt.conf.d
 cat <<EOF | sudo tee /etc/apt/apt.conf.d/proxy.conf
-{{ if .HTTPS_PROXY -}}
+{{- if .HTTPS_PROXY }}
 Acquire::https::Proxy "{{ .HTTPS_PROXY }}";
-{{ end -}}
-{{- if .HTTP_PROXY -}}
+{{- end }}
+{{- if .HTTP_PROXY }}
 Acquire::http::Proxy "{{ .HTTP_PROXY }}";
-{{ end }}
+{{- end }}
 EOF
 
 sudo apt-get update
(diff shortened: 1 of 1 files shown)


erdii:
should we/I maybe think about adding an E2E test with proxies?

pkg/scripts/os.go
```diff
@@ -55,13 +55,12 @@ EOF
...
-{{ end }}
-
+{{- if .HTTPS_PROXY }}
+Acquire::https::Proxy "{{ .HTTPS_PROXY }}";
+{{ end -}}
 {{- if .HTTP_PROXY -}}
-Acquire::http::Proxy = "{{ .HTTP_PROXY }}";
-{{ end }}
+Acquire::http::Proxy "{{ .HTTP_PROXY }}";
+{{ end -}}
```
kron4eg: this last `{{ end -}}` broke the rendering result (from bash PoV).
The rendering result is `cat <<EOF | sudo tee /etc/apt/apt.conf.d/proxy.confEOF` which caused kubeadm to be never installed and fail everything else.

erdii: thanks for the feedback - i will fix it and push again after lunch - can i retrigger the tests myself?

erdii:
/retest

i changed the template to not strip whitespace at the end (to fix the bug I introduced) and at the beginning

pkg/scripts/os.go
```diff
@@ -55,12 +55,11 @@ EOF
...
-Acquire::https::Proxy = "{{ .HTTPS_PROXY }}";
-{{ end }}
-
+{{ if .HTTPS_PROXY -}}
+Acquire::https::Proxy "{{ .HTTPS_PROXY }}";
+{{ end -}}
 {{- if .HTTP_PROXY -}}
-Acquire::http::Proxy = "{{ .HTTP_PROXY }}";
+Acquire::http::Proxy "{{ .HTTP_PROXY }}";
 {{ end }}
```
kron4eg:
```
{{- if .HTTPS_PROXY }}
Acquire::https::Proxy "{{ .HTTPS_PROXY }}";
{{- end }}
{{- if .HTTP_PROXY }}
Acquire::http::Proxy "{{ .HTTP_PROXY }}";
{{- end }}
```

here you can play with it: https://play.golang.org/p/v_uScHqYcKH

erdii:
@kron4eg tests are passing again - how do I continue from here? also:

> should we/I maybe think about adding an E2E test with proxies?

pkg/scripts/os.go
erdii: thx - your solution looks and behaves cleaner - i will take that

kron4eg:
>adding an E2E test with proxies

That will complicate out current e2e matrix, which is already out of control. It's a task for unknown future.

erdii:
ok sounds fair to me! thx for clarification

kron4eg:
@erdii please take a look at https://github.com/kubermatic/kubeone/blob/master/CONTRIBUTING.md#release-notes-block section (this PR currently doesn't feature a release-note block, that's why it has automatic `do-not-merge/release-note-label-needed` label)

PR #795, opened 2020-02-18
(WIP) fixes for coreos control plane

erdii:
**What this PR does / why we need it**: currently centos control plane setup fails on two different things:
1. `pkg/scripts/os.go/kubeadmCentOSTemplate` is referencing `HOST_ARCH` but it is never set - fixed through copy paste from `kubeadmDebianTemplate` (btw same thing is missing from `upgradeKubeletAndKubectlCoreOSScriptTemplate`)
2. on coreos the path `/usr/libexec/kuberentes/plugins/volume` is readonly and this breaks flexvolumes for kube-controller-manager - solution: force different by setting (file: `pkg/templates/kubeadm/v1beta{1,2}`)
  * KubeletExtraArgs["volume-plugin-dir"] = "/etc/kubernetes/kubelet-plugins"
  * ControllerManager.ExtraArgs["flex-volume-plugin-dir"] = "/etc/kubernetes/kubelet-plugins/volume/exec"
  
But I don't know if this specific new path is a good choice and I don't know a clean way to bring knowledge about the nodes target OS to `pkg/templates/kubeadm/v1beta{1,2}`
 
Also:
```
CoreOS Container Linux will no longer receive updates after May 26, 2020.
Begin migrating your workloads to another operating system.
For more info, see https://coreos.com/os/eol/
```

Fixes #

```release-note
```

### DIFF
--- a/pkg/templates/kubeadm/v1beta1/kubeadm.go
+++ b/pkg/templates/kubeadm/v1beta1/kubeadm.go
@@ -69,6 +69,8 @@ func NewConfig(s *state.State, host kubeoneapi.HostConfig) ([]runtime.Object, er
 			"read-only-port":      "0",
 			"rotate-certificates": "true",
 			"cluster-dns":         nodelocaldns.VirtualIP,
+			// default path is not writable on coreos
+			"volume-plugin-dir": "/etc/kubernetes/kubelet-plugins",
 		},
 	}
 
@@ -147,7 +149,9 @@ func NewConfig(s *state.State, host kubeoneapi.HostConfig) ([]runtime.Object, er
 			CertSANs: []string{strings.ToLower(cluster.APIEndpoint.Host)},
 		},
 		ControllerManager: kubeadmv1beta1.ControlPlaneComponent{
-			ExtraArgs:    map[string]string{},
+			ExtraArgs: map[string]string{
+				"flex-volume-plugin-dir": "/etc/kubernetes/kubelet-plugins/volume/exec",
+			},
 			ExtraVolumes: []kubeadmv1beta1.HostPathMount{},
 		},
 		ClusterName: cluster.Name,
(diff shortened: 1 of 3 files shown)


pkg/templates/kubeadm/v1beta1/kubeadm.go
```diff
@@ -69,6 +69,8 @@ func NewConfig(s *state.State, host kubeoneapi.HostConfig) ([]runtime.Object, er
 			"read-only-port":      "0",
 			"rotate-certificates": "true",
 			"cluster-dns":         nodelocaldns.VirtualIP,
+			// default path is not writable on coreos
+			"volume-plugin-dir": "/etc/kubernetes/kubelet-plugins",
```
kron4eg: Let's use /var/lib

```suggestion
			"volume-plugin-dir": "/var/lib/kubelet/volumeplugins",
```

kron4eg:
I see that there is no support for `volume-plugin-dir` in machine-controller...

pkg/templates/kubeadm/v1beta1/kubeadm.go
erdii: Can we use this folder on all supported distros? Otherwise we have to bring knowledge about the used control-plane distros to this template?

erdii:
> I see that there is no support for `volume-plugin-dir` in machine-controller...

 Oh no! :see_no_evil:
I'll do some research and create an issue over there.

erdii:
> I see that there is no support for volume-plugin-dir in machine-controller...

Wait... @kron4eg would you need that setting for control-plane or worker nodes?

Because worker nodes work (heh :grinning:) with centos, coreos and ubuntu.
I tested that with an ubuntu control-plane and 3 different MachineDeployments for each supported worker OS.

kron4eg:
@erdii I *think* this dir needs to be presented everywhere, where mounts can happen.

kron4eg:
Scripts with HOST_ARCH has been fixed in #882
machine-controller PR is pending here: https://github.com/kubermatic/machine-controller/pull/746. Once that PR is merged, I'll add create a PR with new MC version, with kubeadm config changes.

All of above makes this PR obsolete so I'll close it. @erdii thanks for pointing to the issues!

kron4eg:
here's PR to fix flexvolumes #885

PR #802, opened 2020-02-24
Enable Strict mode to decode KubeOne API

### LINKED ISSUE #759: Use strict config unmarshaling
Kubeone config should be unmarshaled from the file using strict mode: unexpected fields should cause an error and be reported to the user.

Sometimes people mistype, this will help to detect problem very early and provide immediate feedback.

kron4eg:
To warn people at very early stages that they have some mistakes in their YAMLs

Fixes #759

```release-note
Strict KubeOne config decoding
```

### DIFF
--- a/pkg/apis/kubeone/scheme/scheme.go
+++ b/pkg/apis/kubeone/scheme/scheme.go
@@ -31,7 +31,7 @@ import (
 var Scheme = runtime.NewScheme()
 
 // Codecs is a CodecFactory object used to provide encoding and decoding for the scheme
-var Codecs = serializer.NewCodecFactory(Scheme)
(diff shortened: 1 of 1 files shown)


xmudrii:
Would it make sense to add a unit test to take the example configuration from [`pkg/cmd/config.go`](https://github.com/kubermatic/kubeone/blob/e3106a066a25ae7dba368d02aaa76c02dce7860a/pkg/cmd/config.go#L402-L631) and try to parse it? In that case, we would know that example config and API are correct and that unmarshaling works as expected.

We can leave this for a follow-up, but I'm just wondering would that make any sense.

kron4eg:
To is it would be "testing k8s.io/apimachinery/pkg/runtime/serializer" and why would we want it? So I avoided those tests.

PR #823, opened 2020-03-13
Add new Untaint API Field to avoid default master taint

### LINKED ISSUE #821: API flag to avoid control plane nodes tainted with NoSchedule

kron4eg:
In some usecases (i.e. at the edge) there is a need to init 1 or ever 3 nodes cluster, but only control-plane nodes without workers in a way that normal workloads would be allowed by default on the those nodes.

Fixes #821 

```release-note
New Untaint API field in config to remove default taints from control plane Nodes
```

### DIFF
--- a/pkg/cmd/config.go
+++ b/pkg/cmd/config.go
@@ -546,6 +546,9 @@ addons:
 #   # prefixed with "env:" to refer to an environment variable.
 #   sshPrivateKeyFile: '/home/me/.ssh/id_rsa'
 #   sshAgentSocket: 'env:SSH_AUTH_SOCK'
+#   # setting this to true will skip node-role.kubernetes.io/master taint from
+#   # Node object on this host
+#   untaint: false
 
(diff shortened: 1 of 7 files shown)


kron4eg:
@xmudrii done, along with rebase

pkg/cmd/config.go
```diff
@@ -546,6 +546,9 @@ addons:
 #   # prefixed with "env:" to refer to an environment variable.
 #   sshPrivateKeyFile: '/home/me/.ssh/id_rsa'
 #   sshAgentSocket: 'env:SSH_AUTH_SOCK'
+#	# setting this to true will skip node-role.kubernetes.io/master taint from
+#	# Node object on this host
+#	untaint: false
```
xmudrii: Do you mind fixing indentation?

kron4eg: screwed up tabs and spaced 🤷‍♂

Fixed

PR #831, opened 2020-03-18
Allow node resources in examples/terraform/vsphere to be changed by variables

eqrx:
Makes is easier to install KubeOne within vsphere without running OOM

Fixes #

```release-note
Increased vsphere template resource definitions
```


kron4eg:
@eqrx sorry, those settings a default for our tight E2E cluster so we have to keep them low, so I can not accept this. Of course you are free to modify it locally as long as you want.

P.S.
Can I propose to replace this change with "terraform vars", with default values are set to what we have now?

kron4eg:
/hold

to avoid any accidents

eqrx:
@kron4eg vars sounds good. will work on that

eqrx:
Changed to variables :)

kron4eg:
thanks!

Issue #816, opened 2020-03-05
KubeOne join for worker nodes


scheeles:
**What feature would you like to be added?**

If a KubeOne user don’t use machine-controller (e.g. bare-metal, edge deployment, Ansible) instead of executing several commands on a worker a command like `kubeone join`  would make the installation easier. 
`kubeone join`will installs all dependencies and and execute `kubeadm join` command. Only one cmd is required to join a worker node.

moelsayed:
As we discussed offline. We will not implement a dedicated `join` command, instead we will use a `StaticWorkers` API field.

PR #843, opened 2020-03-30
Draft proposal for join command

moelsayed:

Fixes #

```release-note
NONE
```

### DIFF
--- a/docs/proposals/20200329-join.md
+++ b/docs/proposals/20200329-join.md
@@ -1,0 +1,48 @@
+# KubeOne Join
+
+**Author:** Mohamed Elsayed ([@moelsayed](https://github.com/moelsayed))  
+**Status:** **Draft** | Review | Final
+**Created:** 2020-03-29
+**Last updated:** 2020-03-29
+
+## Abstract
+KubeOne initially supported deploying control plane nodes only, with the option to deploy machine-controller and dynamically provision worker nodes on a configured cloud provider.
+
+PR [#834](https://github.com/kubermatic/kubeone/pull/834) adds support for configuring static worker nodes to be deployed during initial cluster provisioning.
+
+The `join` command should provide allow users to easily add additional worker nodes to an existing KubeOne cluster.
+
+Currently, this is only doable by applying several manual steps on each worker node to deploy the required packages/files and execute `kubeadm join`.
+
+This new subcommand should allow users to add new static worker nodes to the clusters with a single command, and include these nodes in upgrade process as well.
+
+## Goals
+* Implement the `join` subcommand.
+  * Add the ability to join a static worker node to an existing KubeOne cluster.
+  * Support upgrade process for the added static nodes. These nodes will _not_ be in the cluster configuration file and this should be handled correctly.
+
+## Non-goals
+* Support adding control plane nodes.
+* Support node removal/cleanup. This can be done manually by the user.
+
+
+## Implementation
+* Require standard subcommand arguments
+* Require new node configurations
+  * minimal required CLI options to join the nodes
+  * `HostConfig` YAML manifest for the node
+* Parse options and node configuration, create state
+  * Apply defaults on the node
+
+### Join
+* Continuing after the initial steps
+* Run the following tasks on the new node:
+  * `installPrerequisitesOnNode`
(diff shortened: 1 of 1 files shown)


docs/proposals/20200329-join.md
```diff
@@ -0,0 +1,48 @@
...
+* Parse options and node configuration, create state
+  * Apply defaults on the node
+
+### Join
+* Continuing after the initial steps
+* Run the following tasks on the new node:
+  * `installPrerequisitesOnNode`
+  * `generateKubeadmOnNode`
+  * `joinStaticWorkerInternal`
+* Store node configuration in a ConfigMap/machine object upon successful join.
```
moelsayed: Not very happy about this. State shouldn't be stored on the cluster. So, open to ideas!

docs/proposals/20200329-join.md
```diff
@@ -0,0 +1,48 @@
+# KubeOne Join
```
xmudrii: I'd rename this proposal to something like `KubeOne Static Workers` or `Provisioning Worker Nodes Using KubeOne` to make it clear what this proposal is about.

docs/proposals/20200329-join.md
```diff
@@ -0,0 +1,48 @@
...
+PR [#834](https://github.com/kubermatic/kubeone/pull/834) adds support for configuring static worker nodes to be deployed during initial cluster provisioning.
+
+The `join` command should provide allow users to easily add additional worker nodes to an existing KubeOne cluster.
+
+Currently, this is only doable by applying several manual steps on each worker node to deploy the required packages/files and execute `kubeadm join`.
+
+This new subcommand should allow users to add new static worker nodes to the clusters with a single command, and include these nodes in upgrade process as well.
+
+## Goals
+* Implement the `join` subcommand.
```
xmudrii: To my understanding, the plan was that we don't want to have the `join` command, but instead to extend the `install` command (like in #834). Do we actually want the `join` command? Has the plan changed or I misunderstood something?

docs/proposals/20200329-join.md
```diff
@@ -0,0 +1,48 @@
...
+  * Add the ability to join a static worker node to an existing KubeOne cluster.
+  * Support upgrade process for the added static nodes. These nodes will _not_ be in the cluster configuration file and this should be handled correctly.
```
xmudrii: What is the reason for not having those nodes in the cluster configuration file? If we decide to take on this approach, this should be documented somewhere.

docs/proposals/20200329-join.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Non-goals
+* Support adding control plane nodes.
```
xmudrii: I don't understand this non-goal. We already support adding control plane nodes, but we don't support replacing and rotating. Is this what you meant with this point?

docs/proposals/20200329-join.md
```diff
@@ -0,0 +1,48 @@
...
+* Support node removal/cleanup. This can be done manually by the user.
```
xmudrii: IMO, we should extend the `kubeone reset` command to allow this. Once you unprovision the control plane, you can't use worker nodes until you don't clean them up.

docs/proposals/20200329-join.md
```diff
@@ -0,0 +1,48 @@
...
+
+
+## Implementation
+* Require standard subcommand arguments
+* Require new node configurations
+  * minimal required CLI options to join the nodes
+  * `HostConfig` YAML manifest for the node
```
xmudrii: I think we should extend the core KubeOneCluster API like in #843. I think the best approach is to have a single source of truth (`config.yaml`) and store all cluster configuration there.

docs/proposals/20200329-join.md
xmudrii: I'm also not happy about this at all. This feature should work like we handle control planes. Have a YAML manifest where you would store the config needed to provision those nodes. Then, it's the operator's responsibility to store and keep that file.

moelsayed:
https://github.com/kubermatic/kubeone/issues/816

moelsayed:
As discussed offline, we will go with Static Workers #834 instead of this.

PR #846, opened 2020-03-31
Fix E2E tests image building

xmudrii:

This PR fixes the E2E tests image building. `kubetest` now requires `gcloud` to extract the needed binaries. To avoid setting up `gcloud` in the container, I decided to propose and try this approach. We are already successfully using it for other repositories.

I haven't tested does it work in our case because I can't push the image manually or run the E2E tests locally.

```release-note
NONE
```

/assign @kron4eg

### DIFF
--- a/hack/images/kubeone-e2e/install-kube-tests-binaries.sh
+++ b/hack/images/kubeone-e2e/install-kube-tests-binaries.sh
@@ -22,17 +22,29 @@ full_versions["1.17"]="v1.17.4"
 full_versions["1.18"]="v1.18.0"
 
 root_dir=${KUBETESTS_ROOT:-"/opt/kube-test"}
+tmp_root=${TMP_ROOT:-"/tmp/get-kube"}
 
 for version in "${!full_versions[@]}"; do
     full_version="${full_versions[${version}]}"
     directory="${root_dir}/kubernetes-${version}"
+    tmp_dir="${tmp_root}/kubernetes-${version}"
     if [[ ! -d "${directory}" ]]; then
+        mkdir -p "${tmp_dir}"
         mkdir -p "${directory}"
-        cd "${directory}"
-        kubetest --extract="${full_version}"
+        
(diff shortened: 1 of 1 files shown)


hack/images/kubeone-e2e/install-kube-tests-binaries.sh
```diff
@@ -22,17 +22,29 @@ full_versions["1.17"]="v1.17.4"
...
 
 for version in "${!full_versions[@]}"; do
     full_version="${full_versions[${version}]}"
     directory="${root_dir}/kubernetes-${version}"
+    tmp_dir="${tmp_root}/kubernetes-${version}"
     if [[ ! -d "${directory}" ]]; then
+        mkdir -p "${tmp_dir}"
         mkdir -p "${directory}"
-        cd "${directory}"
-        kubetest --extract="${full_version}"
```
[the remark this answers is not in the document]
xmudrii: As said in the PR description, some of kubetest scripts now require gcloud, which we don’t have in our image. To avoid installing and configuring it, I decided to try this approach.

hack/images/kubeone-e2e/install-kube-tests-binaries.sh
```diff
@@ -22,17 +22,29 @@ full_versions["1.17"]="v1.17.4"
...
+        
+        curl -L http://gcsweb.k8s.io/gcs/kubernetes-release/release/"${full_version}"/kubernetes.tar.gz -o "${tmp_dir}"/kubernetes.tar.gz
```
kron4eg, line 35:
```suggestion
        curl -L https://gcsweb.k8s.io/gcs/kubernetes-release/release/"${full_version}"/kubernetes.tar.gz -o "${tmp_dir}"/kubernetes.tar.gz
```

xmudrii: Done.

PR #854, opened 2020-04-02
Change default make target to build

eqrx:
Usually the default target of `make` is something that builds the project it manages. Currently kubeone has `install` as default target which installs kubeone to the gopath, but does not touch the `dist/` folder. This leads to confusion (on my side). I would like to change the default target to `build` and make `install` depending on it, even though both targets build the binary for themselves.

Fixes #

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,14 +24,17 @@ GITTAG=$(shell git describe --tags --always)
 GOLDFLAGS?=-extldflags=-zrelro -extldflags=-znow -s -w -X github.com/kubermatic/kubeone/pkg/cmd.version=$(GITTAG) -X github.com/kubermatic/kubeone/pkg/cmd.commit=$(GITCOMMIT) -X github.com/kubermatic/kubeone/pkg/cmd.date=$(BUILD_DATE)
 
 .PHONY: all
-all: install
-
-.PHONY: install
-install: buildenv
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -24,15 +24,18 @@ GITTAG=$(shell git describe --tags --always)
...
+all: build
 
 .PHONY: build
 build: dist/kubeone
 
+dist/kubeone: buildenv
+	go build -ldflags='$(GOLDFLAGS)' -v -o $@ .
+
+.PHONY: install
+install: build
```
kron4eg, line 36:
```suggestion
install:
```

there is no need to depend on `build` here, as `go install` will also `go build`

eqrx: the reason for that is is described [here](https://www.gnu.org/software/make/manual/html_node/Standard-Targets.html#Standard-Targets). 
```
Compile the program and copy the executables, libraries, and so on to the file names where they should reside for actual use. If there is a simple test to verify that a program is properly installed, this target should run that test.
```

I mean yes, we do not need it.... but if somebody is starting with kubeone it sure will be fascinating that we do not need to compile the program to install it

kron4eg:
/close

for now we will keep Makefile the way we used to

PR #862, opened 2020-04-12
CNIProviderExternal is an external CNI plugin 

scheeles:
The CNI plugin can be installed as Addon or manually

```release-note
External CNI Plugin - The CNI plugin can be installed as Addon or manually
```


pkg/apis/kubeone/types.go
```diff
@@ -148,6 +148,11 @@ const (
 	// * In case when encryption is enabled, strong secret will be autogenerated
 	// More info: https://www.weave.works/docs/net/latest/kubernetes/kube-addon/
 	CNIProviderWeaveNet CNIProvider = "weave-net"
+
+	// CNIProviderExternal is an external CNI plugin .
```
xmudrii, line 152:
```suggestion
	// CNIProviderExternal is an external CNI plugin.
```
Tiny nit: space is not needed here

pkg/cmd/config.go
```diff
@@ -425,7 +425,8 @@ clusterNetwork:
   cni:
     # possible values:
     # * canal
-    # * weave-net
+	# * weave-net
+	# * external - The CNI plugin can be installed as Addon or manually
```
xmudrii, line 429:
```suggestion
    # * weave-net
    # * external - The CNI plugin can be installed as Addon or manually
```
Tiny nit: wrong indentation, please use spaces instead of tabs

pkg/apis/kubeone/types.go
scheeles: fixed

pkg/cmd/config.go
scheeles: fixed

PR #863, opened 2020-04-12
Improve the installation experience

scheeles:
```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,31 +50,29 @@ upgrade to the v0.6.0 or newer as soon as possible.
 
 ### Downloading a binary from GitHub Releases
 
-The recommended way to obtain KubeOne is to grab the
-binary from the [GitHub Releases][9] page. On the
-releases page, find the binary for your operating system
-and architecture and download it or grab the URL and use
-`wget` or `curl` to download the binary.
-
-**Version:** version of KubeOne  
-**Operating system:** `linux` or `darwin` for macOS
-
+The fastest way to get KubeOne:
 ```bash
-curl -LO https://github.com/kubermatic/kubeone/releases/download/v<version>/kubeone_<version>_<operating_system>_amd64.zip
+curl https://raw.githubusercontent.com/kubermatic/kubeone/master/intsallsh | sh
 ```
+
+If you want to have more control over how KubeOne is installed, download the
+binary from the [GitHub Releases][9] page. 
+
+On the releases page, you find the binary for your operating system
+and architecture 
+
+Download it or grab the URL and use `wget` or `curl` to download the binary.
 
 Extract the binary to the KubeOne directory. On Linux and macOS, you can use `unzip`.
 
-```bash
-unzip kubeone_<version>_<operating_system>_amd64.zip -d kubeone_<version>_<operating_system>_amd64
-```
-
-Move the `kubeone` binary to your path, so you can easily
-invoke it from your terminal.
+Move the `kubeone` binary to your path, so you can easily invoke it from your terminal.
 
 ```bash
-sudo mv kubeone_<version>_<operating_system>_amd64/kubeone /usr/local/bin
-
+OS=$(uname)
+VERSION=$(curl -w '%{url_effective}' -I -L -s -S https://github.com/kubermatic/kubeone/releases/latest -o /dev/null | sed -e 's|.*/v||')
+curl -LO https://github.com/kubermatic/kubeone/releases/download/v<version>/kubeone_<version>_<operating_system>_amd64.zip
+unzip kubeone_${VERSION}_${OS}_amd64.zip -d kubeone_${VERSION}_${OS}amd64
+sudo mv kubeone_${VERSION}_${OS}amd64/kubeone /usr/local/bin
 ```
 
 ### Building KubeOne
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -50,31 +50,29 @@ upgrade to the v0.6.0 or newer as soon as possible.
...
 ```
 
-Extract the binary to the KubeOne directory. On Linux and macOS, you can use `unzip`.
+If you want to have more control over how KubeOne is installed, download the
+binary from the [GitHub Releases][9] page. 
 
-```bash
-unzip kubeone_<version>_<operating_system>_amd64.zip -d kubeone_<version>_<operating_system>_amd64
-```
+On the releases page, you find the binary for your operating system
```
xmudrii, line 61:
```suggestion
On the releases page, you can find the binary for your operating system
```

README.md
```diff
@@ -50,31 +50,29 @@ upgrade to the v0.6.0 or newer as soon as possible.
...
+Download it or grab the URL and use `wget` or `curl` to download the binary.
 
-```bash
-sudo mv kubeone_<version>_<operating_system>_amd64/kubeone /usr/local/bin
+Extract the binary to the KubeOne directory. On Linux and macOS, you can use `unzip`.
+
+Move the `kubeone` binary to your path, so you can easily invoke it from your terminal.
 
+```bash
+OS=$(uname)
```
xmudrii, line 71: This approach doesn't work on Linux, because `uname` returns `Linux` (with uppercase `L`), which breaks `unzip` and `mv`. Is there any (easy) way to make it lowercase or can we use some other approach here?

install.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/sh
+#
```
xmudrii, line 2: The license header is missing, which makes verify presubmit to fail.

install.sh
```diff
@@ -0,0 +1,40 @@
...
+# This is a simple installer script for KubeOne 
+#
+#
+# What OS is used
+OS=$(uname)
```
xmudrii, line 7: As mentioned for the README, this doesn't work on Linux. I like the script approach, but we should make it work on Linux as well.

README.md
```diff
@@ -50,31 +50,29 @@ upgrade to the v0.6.0 or newer as soon as possible.
...
+VERSION=$(curl -w '%{url_effective}' -I -L -s -S https://github.com/kubermatic/kubeone/releases/latest -o /dev/null | sed -e 's|.*/v||')
+curl -LO https://github.com/kubermatic/kubeone/releases/download/v<version>/kubeone_<version>_<operating_system>_amd64.zip
+unzip kubeone_${VERSION}_${OS}_amd64.zip -d kubeone_${VERSION}_${OS}amd64
```
xmudrii, line 74:
```suggestion
unzip kubeone_${VERSION}_${OS}_amd64.zip -d kubeone_${VERSION}_${OS}_amd64
```
Very tiny nit: I think it would look better with `_` between `OS` and `arch`. It also matches the archive name that way.

install.sh
```diff
@@ -0,0 +1,40 @@
...
+DEST=/usr/local/bin
+
+# Download the latest version for the OS and save it as zip
+curl -LO $URL
+
+if [ $? -eq 0 ]
+then 
+    echo "Copying kubeone binary into $DEST"
+    # unpack:
+    unzip kubeone_${VERSION}_${OS}_amd64.zip -d kubeone_${VERSION}_${OS}amd64
```
xmudrii, line 26:
```suggestion
    unzip kubeone_${VERSION}_${OS}_amd64.zip -d kubeone_${VERSION}_${OS}_amd64
```
Very tiny nit: I think it would look better with `_` between `OS` and `arch`. It also matches the archive name that way.

install.sh
xmudrii: It should be possible to use `tr` or `awk` to lowercase the string:
```
OS=$(uname | tr '[:upper:]' '[:lower:]')
```
or with `awk`
```
OS=$(uname | awk '{print tolower($0)}')
```
(reference: https://stackoverflow.com/questions/2264428/how-to-convert-a-string-to-lower-case-in-bash)

README.md
scheeles: This is working on Ubuntu. You can download the file from https://github.com/kubermatic/kubeone/releases/download/v0.11.1/kubeone_0.11.1_Linux_amd64.zip

scheeles:
> This looks great! I like the script approach, it makes getting KubeOne much easier. We should just try to make it work on both macOS and Linux.
> 
> We should also update [the landing page](https://github.com/loodse/docs/blob/master/content/kubeone/master/_index.en.md) in `loodse/docs`. Do you want to take care of that or should we do that after the PR is merged?

I would change it afterwards after this is merged.

README.md
```diff
@@ -50,31 +50,29 @@ upgrade to the v0.6.0 or newer as soon as possible.
...
-```bash
-sudo mv kubeone_<version>_<operating_system>_amd64/kubeone /usr/local/bin
+Extract the binary to the KubeOne directory. On Linux and macOS, you can use `unzip`.
+
+Move the `kubeone` binary to your path, so you can easily invoke it from your terminal.
 
+```bash
+OS=$(uname)
+VERSION=$(curl -w '%{url_effective}' -I -L -s -S https://github.com/kubermatic/kubeone/releases/latest -o /dev/null | sed -e 's|.*/v||')
+curl -LO https://github.com/kubermatic/kubeone/releases/download/v<version>/kubeone_<version>_<operating_system>_amd64.zip
```
xmudrii, line 73: Should we use the environment variables here?
```suggestion
curl -LO https://github.com/kubermatic/kubeone/releases/download/v${VERSION}/kubeone_${VERSION}_${OS}_amd64.zip
```

PR #873, opened 2020-04-21
addons: apply only yaml/yml/json manifests

xmudrii:

Apply only addons that have `.yaml`, `.yml` and `.json` extensions.

This prevents errors such as:
```
INFO[12:46:28 CEST] Applying addons…                             
INFO[12:46:28 CEST] Parsing addons manifest 'README.md'          
WARN[12:46:28 CEST] Task failed…                                 
WARN[12:46:28 CEST] error was: failed to decode manifest README.md: error converting YAML to JSON: yaml: line 4: mapping values are not allowed in this context 
```

```release-note
Apply only addons with .yaml, .yml and .json extensions
```

/assign @kron4eg

### DIFF
--- a/pkg/addons/manifest.go
+++ b/pkg/addons/manifest.go
@@ -78,6 +78,14 @@ func loadAddonsManifests(addonsPath string, logger logrus.FieldLogger, verbose b
 			logger.Infof("Found directory '%s' in the addons path. Ignoring.\n", file.Name())
 			continue
 		}
+		ext := strings.ToLower(filepath.Ext(filePath))
+		// Only YAML/YML and JSON manifests are supported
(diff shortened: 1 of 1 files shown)


pkg/addons/manifest.go
```diff
@@ -78,6 +78,13 @@ func loadAddonsManifests(addonsPath string, logger logrus.FieldLogger, verbose b
 			logger.Infof("Found directory '%s' in the addons path. Ignoring.\n", file.Name())
 			continue
 		}
+		ext := strings.ToLower(filepath.Ext(filePath))
+		if ext != ".yaml" && ext != ".yml" {
```
kron4eg, line 82: I suppose .json should be also viable

xmudrii: Makes sense. Fixed.

PR #880, opened 2020-04-24
CLI BREAKING Replace positional argument for config.yaml with global flag

### LINKED ISSUE #866: make config default to ./kubeone.yaml
## What feature would you like to be added?

In case when the config is not provided explicitly, default config to `./kubeone.yaml` path (in current directory).

## What are use cases of the feature?

As we have some additional subcommands in pipeline coming, it will be super handy to not to type config file every single time.

### LINKED ISSUE #867: Replace positional argument for config.yaml with global flag
## What feature would you like to be added?
Basically subject. Path to the config file should be taken as global flag, instead of positional argument.

## What are use cases of the feature?
As there will be other (sub)command that will require their own positional arguments it will be hard for us to distinguish between config and subcommand args.

kron4eg:
Fixes #866
Fixes #867

docs: https://github.com/loodse/docs/pull/303

```release-note
* CLI BREAKING replace positional argument for config.yaml with global (`--manifest`) flag
* Default new --manifest flag to ./kubeone.yaml
```

### DIFF
--- a/pkg/cmd/install.go
+++ b/pkg/cmd/install.go
@@ -73,7 +73,7 @@ func installCmd(rootFlags *pflag.FlagSet) *cobra.Command {
 	opts := &installOpts{}
 
 	cmd := &cobra.Command{
-		Use:   "install <manifest>",
+		Use:   "install",
 		Short: "Install Kubernetes",
 		Long: `
 Install Kubernetes on pre-existing machines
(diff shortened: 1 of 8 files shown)


kron4eg:
/hold

this PR requires docs changes

kron4eg:
/hold cancel

docs: https://github.com/loodse/docs/pull/303

Issue #870, opened 2020-04-14
CoreOS to Flatcar migration


kron4eg:
## What feature would you like to be added?

Because of approaching CoreOS EOL we need to prepare migration to Flatcar: decide if we need to automate [migration from CoreOS to Flatcat][flatcar-migration]

UPDATE (26/04/2020):
Support for Flatcar Linux has landed in the source tree, now it depends on the machine-controller to start supporting all the providers (at the moment it's AWS only).

UPDATE (28/04/2020):
[CoreOS -> Flatcar Linux][flatcar-migration-script] migration doc and [shell script][flatcar-migration-script]

[flatcar-migration]: https://docs.flatcar-linux.org/os/migrate-from-container-linux/
[flatcar-migration-doc]: https://docs.flatcar-linux.org/os/update-from-container-linux/
[flatcar-migration-script]: https://docs.flatcar-linux.org/update-to-flatcar.sh

kron4eg:
OK, since there is an upstream-provided migration [shell script][flatcar-migration-script], we don't need to do anything here.

[flatcar-migration-script]: https://docs.flatcar-linux.org/update-to-flatcar.sh

PR #882, opened 2020-04-26
CoreOS host architecture detection fix

kron4eg:
Ported fixes for coreOS from #795
Plus golden tests

```release-note
Fixed CoreOS host architecture detection
```

### DIFF
--- a/pkg/scripts/os_test.go
+++ b/pkg/scripts/os_test.go
@@ -1,0 +1,248 @@
+/*
+Copyright 2019 The KubeOne Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package scripts
+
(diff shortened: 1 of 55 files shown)


pkg/scripts/os_test.go
```diff
@@ -0,0 +1,248 @@
...
+	"testing"
+
+	"github.com/kubermatic/kubeone/pkg/apis/kubeone"
+	"github.com/kubermatic/kubeone/pkg/testhelper"
+)
+
+func genCluster() kubeone.KubeOneCluster {
+	return kubeone.KubeOneCluster{
+		Versions: kubeone.VersionConfig{
+			Kubernetes: "v1.17.4",
```
xmudrii, line 29:
```suggestion
			Kubernetes: "1.17.4",
```

The leading `v` in the version generates invalid scripts. Do we have some sort of validation or do we drop the leading `v` if the user provides it?

kron4eg: kubeone.KubeOneCluster is validated and defaulted of course, long before it get to this point.

xmudrii: I've tested this and if the user provides a config file such as the following one, provisioning would fail.

```yaml
kind: KubeOneCluster
versions:
  kubernetes: "v1.18.2"
cloudProvider:
  name: aws
```

We should add a validation rule or defaulting to default to the version without the leading `v`. If you prefer, we can do it in a follow-up.

kron4eg: heh, yeah, `v, err := semver.NewVersion(version.Kubernetes)` will parse with `v` or without successfully.

kron4eg:
xref #870

PR #889, opened 2020-05-04
Add KubeOneCluster v1beta1 API

### LINKED ISSUE #828: [brainstorm] v1beta1 API ideas
This issue will serve as a collection of ideas to gather for new API version. Later good ideas will be included into the proposal and eventually implemented.

EDIT
Things TODO:
- [x] Docs (https://github.com/kubermatic/docs/pull/328)
- [x] Move e2e to new API (#917)

xmudrii:

This PR adds the KubeOneCluster v1beta1 API. Compared to v1alpha1, the following changes are implemented:
* CloudProviders are defined as structs instead of string/enums
* `MachineControllerConfig.Provider` field is removed
* CNI plugins are defined as structs instead of string/enums
* `HostConfig.Untaint` field is replaced with `HostConfig.Taints`
  * Type is `[]corev1.Taint` instead of string
  * If `HostConfig.Taints` is nil, default taint will be applied for control plane nodes (`node-role.kubernetes.io/master` with NoSchedule effect). For worker nodes, no taint will be applied.
  * If `len(HostConfig.Taints)` is 0, no taints will be applied
* Added operating systems enum
* `Credentials` top-level field is removed
* Node-related fields are renamed
  * `Hosts` -> `ControlPlane.Hosts`
  * `StaticWorkers` -> `StaticWorkers.Hosts`
  * `Workers` -> `DynamicWorkers` (type `WorkerConfig` is renamed to `DynamicWorkerConfig`)
* Move `Network.NetworkID` to `HetznerSpec.NetworkID`
* Rename `ProviderSpec.Network` to `ProviderSpec.ProviderStaticNetworkConfig`

xref #828 

```release-note
Add KubeOneCluster v1beta1 API
```

/assign @kron4eg


.golangci.yml
```diff
@@ -59,3 +59,4 @@ issues:
   - "func SetDefaults_SystemPackages should be SetDefaultsSystemPackages"
   - "type name will be used as kubeone.KubeOneCluster by other packages"
   - "G402: TLS InsecureSkipVerify set true"
+  - "methods on the same type should have the same receiver name"
```
kron4eg, line 62: this will kill it globally, please don't

pkg/apis/kubeone/v1beta1/types.go
```diff
@@ -0,0 +1,324 @@
...
+
+// OperatingSystemName defines the operating system used on instances
+type OperatingSystemName string
+
+var (
+	OperatingSystemNameUbuntu  OperatingSystemName = "ubuntu"
+	OperatingSystemNameCentOS  OperatingSystemName = "centos"
+	OperatingSystemNameCoreOS  OperatingSystemName = "coreos"
+	OperatingSystemNameFlatcar OperatingSystemName = "flatcar"
+	OperatingSystemNameDefault OperatingSystemName = ""
```
kron4eg, line 98:
```suggestion
	OperatingSystemNameUbuntu  OperatingSystemName = "ubuntu"
	OperatingSystemNameCentOS  OperatingSystemName = "centos"
	OperatingSystemNameCoreOS  OperatingSystemName = "coreos"
	OperatingSystemNameFlatcar OperatingSystemName = "flatcar"
	OperatingSystemNameUnknown OperatingSystemName = ""
```

pkg/apis/kubeone/v1beta1/types.go
```diff
@@ -0,0 +1,324 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// KubeOneCluster is KubeOne Cluster API Schema
+type KubeOneCluster struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// Name is the name of the cluster
+	Name string `json:"name"`
+	// Hosts describes the control plane nodes and how to access them
+	Hosts []HostConfig `json:"hosts,omitempty"`
```
kron4eg, line 37:
```suggestion
	ControlPlane []HostConfig `json:"controlPlane,omitempty"`
```

pkg/apis/kubeone/v1beta1/types.go
```diff
@@ -0,0 +1,324 @@
...
+	// CloudProvider configures the cloud provider specific features
+	CloudProvider CloudProviderSpec `json:"cloudProvider,omitempty"`
+	// Versions defines which Kubernetes version will be installed
+	Versions VersionConfig `json:"versions,omitempty"`
+	// ClusterNetwork configures the in-cluster networking
+	ClusterNetwork ClusterNetworkConfig `json:"clusterNetwork,omitempty"`
+	// Proxy configures proxy used while installing Kubernetes and by the Docker daemon
+	Proxy ProxyConfig `json:"proxy,omitempty"`
+	// Workers is used to create worker nodes using the Kubermatic machine-controller
+	Workers []WorkerConfig `json:"workers,omitempty"`
```
kron4eg, line 51:
```suggestion
	DynamicWorkers []WorkerConfig `json:"dynamicWorkers,omitempty"`
```

Albeit, I'm not 100% sure about this name.

pkg/apis/kubeone/v1beta1/types.go
```diff
@@ -0,0 +1,324 @@
...
+type AWSSpec struct{}
+
+// AzureSpec defines the Azure cloud provider
+type AzureSpec struct{}
+
+// DigitalOceanSpec defines the DigitalOcean cloud provider
+type DigitalOceanSpec struct{}
+
+// GceSpec defines the GCE cloud provider
+type GceSpec struct{}
```
kron4eg, line 136:
```suggestion
type GCESpec struct{}
```

pkg/apis/kubeone/v1beta1/types.go
```diff
@@ -0,0 +1,324 @@
...
+
+// CloudProviderSpec describes the cloud provider that is running the machines.
+// Only one cloud provider must be defined at the single time.
+type CloudProviderSpec struct {
+	External     bool              `json:"external"`
+	CloudConfig  string            `json:"cloudConfig"`
+	AWS          *AWSSpec          `json:"aws"`
+	Azure        *AzureSpec        `json:"azure"`
+	DigitalOcean *DigitalOceanSpec `json:"digitalocean"`
+	Gce          *GceSpec          `json:"gce"`
```
kron4eg, line 118:
```suggestion
	GCE          *GCESpec          `json:"gce"`
```

pkg/apis/kubeone/v1beta1/types.go
```diff
@@ -0,0 +1,324 @@
...
+// NetworkConfig contains a machine's static network configuration
+type NetworkConfig struct {
+	CIDR    string    `json:"cidr"`
+	Gateway string    `json:"gateway"`
+	DNS     DNSConfig `json:"dns"`
+}
+
+// MachineControllerConfig configures kubermatic machine-controller deployment
+type MachineControllerConfig struct {
+	Deploy bool `json:"deploy"`
```
kron4eg, line 233: I think this field is not needed anymore, and so whole structure can go away.

pkg/apis/kubeone/v1beta1/types.go
```diff
@@ -0,0 +1,324 @@
...
+
+// VsphereSpec defines the vSphere provider
+type VsphereSpec struct{}
+
+// NoneSpec defines a none provider
+type NoneSpec struct{}
+
+// VersionConfig describes the versions of components that are installed on the machines
+type VersionConfig struct {
+	Kubernetes semver.Semver `json:"kubernetes"`
```
kron4eg, line 155: I'm totally no fan of including other packages as types in API. Let's handle semver at validator level.

pkg/apis/kubeone/v1beta1/types.go
```diff
@@ -0,0 +1,324 @@
...
+	// MachineController configures the Kubermatic machine-controller component
+	MachineController *MachineControllerConfig `json:"machineController,omitempty"`
+	// Features enables and configures additional cluster features
+	Features Features `json:"features,omitempty"`
+	// Addons are used to deploy additional manifests
+	Addons *Addons `json:"addons,omitempty"`
+	// SystemPackages configure kubeone behaviour regarding OS packages
+	SystemPackages *SystemPackages `json:"systemPackages,omitempty"`
+	// Credentials used for machine-controller and external CCM
+	Credentials map[string]string `json:"credentials,omitempty"`
```
kron4eg, line 61: This field form v1alpha1 is not used in our codebase, do we really need it?

.golangci.yml
xmudrii: This is now gone since I removed the semver API.

pkg/apis/kubeone/v1beta1/types.go
xmudrii: Sounds good. Removed the semver API and switched back to string.

pkg/apis/kubeone/v1beta1/types.go
xmudrii: We don't even use it in v1alpha1. I removed it from v1beta1.

pkg/apis/kubeone/v1beta1/types.go
xmudrii: I disagree with this. The operator should be able to choose do they want to use machine-controller or not. We don't depend on machine-controller (you can choose to not deploy workers or use static workers), so it should remain optional.

xmudrii:
@kron4eg Renamed fields as well. PTAL.

pkg/apis/kubeone/v1beta1/types.go
xmudrii: Sounds good enough to me.

pkg/apis/kubeone/v1beta1/types.go
```diff
@@ -0,0 +1,321 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// KubeOneCluster is KubeOne Cluster API Schema
+type KubeOneCluster struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// Name is the name of the cluster
+	Name string `json:"name"`
+	// ControlPlane describes the control plane nodes and how to access them
+	ControlPlane []HostConfig `json:"controlPlane,omitempty"`
```
kron4eg, line 35: @xmudrii how do you think, does it make sense to turn this into a struct with inside .Hosts field?

xmudrii: Do you mean something along:
```go
type Hosts struct {
    ControlPlane  []HostConfig `json:"controlPlane"`
    StaticWorkers []HostConfig `json:"staticWorkers"`
}
```

In this case `StaticWorkers` can even be called just `Workers`, but I'd maybe prefer `StaticWorkers`.

Did you think of something like this?

kron4eg: not quite, more like 
```go
...
ControlPlane ControlPlane `json:"controlPlane"`
...
type ControlPlane struct {
    Hosts  []HostConfig `json:"hosts,omitempty"`
}
```

kron4eg: Adding something to `type ControlPlane struct` at later times is easy, unlike to `[]Hosts`.

xmudrii: @kron4eg Updated, PTAL.

pkg/apis/kubeone/v1beta1/types.go
```diff
@@ -0,0 +1,331 @@
...
+	PrivateAddress    string `json:"privateAddress"`
+	SSHPort           int    `json:"sshPort"`
+	SSHUsername       string `json:"sshUsername"`
+	SSHPrivateKeyFile string `json:"sshPrivateKeyFile"`
+	SSHAgentSocket    string `json:"sshAgentSocket"`
+	Bastion           string `json:"bastion"`
+	BastionPort       int    `json:"bastionPort"`
+	BastionUser       string `json:"bastionUser"`
+	Hostname          string `json:"hostname"`
+	IsLeader          bool   `json:"isLeader"`
```
xmudrii, line 85: @kron4eg WDYT about dropping this field and switching to automatically determining the leader? I think that's important for the `apply` use case.

kron4eg: if we can do it, that'd be awesome

pkg/apis/kubeone/v1beta1/types.go
```diff
@@ -0,0 +1,331 @@
...
+
+	// If not provided (i.e. nil) defaults to TaintEffectNoSchedule, with key
+	// node-role.kubernetes.io/master for control plane nodes.
+	//
+	// Explicitly empty (i.e. []corev1.Taint{}) means no taints will be applied
+	// (this is default for worker nodes).
+	Taints []corev1.Taint `json:"taints,omitempty"`
```
kron4eg, line 92: @xmudrii WDYT about moving this down to ControlPlaneConfig?

pkg/apis/kubeone/v1beta1/types.go
```diff
@@ -0,0 +1,331 @@
...
+	Host string `json:"host"`
+
+	// Port is the port used to reach to the API
+	Port int `json:"port"`
+}
+
+// CloudProviderSpec describes the cloud provider that is running the machines.
+// Only one cloud provider must be defined at the single time.
+type CloudProviderSpec struct {
+	External     bool              `json:"external"`
```
kron4eg, line 120: Hm. This probably also should be a struct.

pkg/apis/kubeone/v1beta1/types.go
```diff
@@ -0,0 +1,331 @@
...
+// ClusterNetworkConfig describes the cluster network
+type ClusterNetworkConfig struct {
+	PodSubnet         string `json:"podSubnet"`
+	ServiceSubnet     string `json:"serviceSubnet"`
+	ServiceDomainName string `json:"serviceDomainName"`
+	NodePortRange     string `json:"nodePortRange"`
+	CNI               *CNI   `json:"cni,omitempty"`
+
+	// +optional
+	NetworkID string `json:"networkID,omitempty"`
```
kron4eg, line 174: This needs to be dropped (or moved to HetznerSpec), as it's used in one single place, to assign hetzner CCM deployment an argument.

pkg/apis/kubeone/v1beta1/types.go
```diff
@@ -0,0 +1,331 @@
...
+	OverwriteCloudConfig *string `json:"overwriteCloudConfig,omitempty"`
+}
+
+// DNSConfig contains a machine's DNS configuration
+type DNSConfig struct {
+	Servers []string `json:"servers"`
+}
+
+// NetworkConfig contains a machine's static network configuration
+type NetworkConfig struct {
```
kron4eg, line 232:
```suggestion
	Network *ProviderStaticNetworkConfig `json:"network,omitempty"`

	// +optional
	OverwriteCloudConfig *string `json:"overwriteCloudConfig,omitempty"`
}

// DNSConfig contains a machine's DNS configuration
type DNSConfig struct {
	Servers []string `json:"servers"`
}

// ProviderStaticNetworkConfig contains a machine's static network configuration
type ProviderStaticNetworkConfig struct {
```

pkg/apis/kubeone/v1beta1/types.go
xmudrii: For the sake of flexibility, I think it would be best if it stays in the `HostConfig`.

Moving it to `ControlPlaneConfig` would mean that the same taints would be applied to all control plane nodes. That's not always what you want, for example, in HA setups, you can leave one node just for CP components and use other nodes for other workloads as well. This is especially important if machines don't have the same hardware. Same applies for the worker nodes.

pkg/apis/kubeone/v1beta1/types.go
xmudrii: What we would gain from that? And how it would look like actually? From the API usability side, I think having it as a bool would be the best approach. I'd just say that we should make it a named-bool instead.

I've been thinking about putting it into the provider-specific structs that we added. But I have realized that you can choose to use external for any provider, even for AWS, Azure... A top-level field in CloudProviderSpec makes more sense because of that.

Maybe it could be useful to have a struct such as:
```go
type ExternalCloudProviderConfig struct {
    DeployCCM bool `json:"deployCCM"`
}
```
That would allow some additional control for providers with the external CCM templates built-in.

pkg/apis/kubeone/v1beta1/types.go
xmudrii: I vote that we moved it to the `HetznerSpec` if it's Hetzner-specific.

pkg/apis/kubeone/v1beta1/types.go
xmudrii: This diff got a bit messy, but as far as I understand, the proposal is to change `NetworkConfig` to `ProviderStaticNetworkConfig`. Is there anything else that I'm missing?

kron4eg: nothing more :)

pkg/apis/kubeone/v1beta1/types.go
xmudrii: I've thought about this a bit more. I don't think we can get rid of it because we still need the leader concept in order to easily provision and upgrade the cluster.

We can change the defaulting instead. Instead of choosing the first host as the leader, we would do it at the runtime. We would run a set of commands to determine is there Kubernetes running on the instance and once we find the first instance with Kubernetes running, pick it as the leader. If there is no such instance, pick the first one as the leader.

That would result in the same result as now, but there is one important case handled. If the first instance is not responsible, we would pick the second one instead, and we would be able to do the stuff as usual.

I would consider making the field inaccessible to the users because if some user sets the field to the wrong host, apply might give wrong results. On the other side, if we incorrectly determine the leader, bad things would happen.

What do you think about this? What should we do?

pkg/apis/kubeone/v1beta1/types.go
kron4eg: A combination of provider=none and external=false gives us what ExternalCloudProviderConfig.DeployCCM can give (and they seems like mutually exclusive), so I'd be against `ExternalCloudProviderConfig`.

kron4eg: OK, let it stay a bool

pkg/apis/kubeone/v1beta1/types.go
kron4eg: :lgtm:

pkg/apis/kubeone/v1beta1/types.go
kron4eg: Choosing a leader should be done really carefully.
There is a situation when first instance in responsive, but empty. It happens when 1 instances has been rotated/wiped and needs "join back to the cluster" actions.

xmudrii: Let's leave this discussion for when we start working on the apply.

xmudrii:
@kron4eg Addressed comments, PTAL.

PR #896, opened 2020-05-19
Bind docker-ce-cli to the same version as docker-ce

xmudrii:

This PR binds the `docker-ce-cli` package to the same version as the `docker-ce` package. This fixes the incompatibilities between CLI and daemon, causing provisioning to fail.

Before this fix, we were getting the following error when trying to provision the cluster:

```
Server:
ERROR: Error response from daemon: client version 1.40 is too new. Maximum supported API version is 1.39
```

This error was present at the `Running kubeadm` phase. It was happening because the latest Docker CLI requires the daemon supporting API version 1.40 or newer.

```release-note
Bind docker-ce-cli to the same version as docker-ce. This fixes the cluster provisioning failures.
```

/assign @kron4eg


xmudrii:
/hold
to prevent automatic retest. Quay is currently down, so it's not possible to pull the CNI image, causing all tests to fail

PR #920, opened 2020-06-12
Add support for the PodNodeSelector admission controller

### LINKED ISSUE #852: Add support for PodNodeSelector admission controller

User wants to enable the PodNodeSelector admission controller.

To solve this, either add support to enable this admission controller in the KubeOne config or give the possibility to enable arbitrary admission controllers.

xmudrii:

Add support for the PodNodeSelector admission controller.

The admission controller is enabled by enabling the appropriate feature, such as:

```yaml
...
features:
  podNodeSelector:
    enable: true
    config:
      configFilePath: "./podnodeselector.yaml"
```

The `podnodeselector.yaml` file is located on the **local file system** and contains the plugin configuration. More about the configuration format can be found in the [Kubernetes docs](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#configuration-file-format-1).

Fixes #852 

* To make the `kube-system` components able to get scheduled on all nodes, an empty `scheduler.alpha.kubernetes.io/node-selector` annotation is added to the `kube-system` namespace. The `kube-system` namespace mustn't be restricted in the plugin configuration, as pods such as CNI and kube-proxy must be scheduled on all nodes.
* Before the `kube-system` namespace gets annotated, `kubeadm` creates deployments/daemonsets for the needed components. The underlying pods have the incorrect node selectors and therefore are stuck in the Pending state. To resolve this problem, all pending pods in the `kube-system` namespace are removed, so they get recreated with the correct node selector.
* The AdmissionConfiguration API has been backported, so we don't have to import the `k8s.io/apiserver` package. Importing the package would require us to bump all Kubernetes dependencies to v1.18, as well as, controller-runtime to v0.5.0 or newer.

```release-note
Add support for the PodNodeSelector admission controller
BREAKING: Make paths for PodNodeSelector and StaticAudit configs relative to the manifest
```

/assign @kron4eg


pkg/apis/kubeone/v1alpha1/types.go
```diff
@@ -215,6 +215,7 @@ type MachineControllerConfig struct {
 
 // Features controls what features will be enabled on the cluster
 type Features struct {
+	PodNodeSelector   *PodNodeSelector   `json:"podNodeSelector"`
```
kron4eg, line 218: While there is no technical obstacle to add this new feature to v1alpha1, I think we should avoid doing this, as v1alpha1 should be "conserved" and not changed anymore.

xmudrii: I've thought about this as well, but I think we should keep both APIs in sync until we don't officially announce the deprecation with the 1.0 release. After that, only v1beta1 should be extended.

xmudrii:
@kron4eg PTAL

pkg/apis/kubeone/types.go
```diff
@@ -257,6 +258,21 @@ type SystemPackages struct {
...
+	Config PodNodeSelectorConfig `json:"config"`
+}
+
+// PodNodeSelectorConfig config
+type PodNodeSelectorConfig struct {
+	// ConfigFilePath is a path on the local file system to the PodNodeSelector
+	// configuration file.
+	// ConfigFilePath is a required field.
+	// More info: https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#podnodeselector
+	ConfigFilePath string `json:"configFilePath"`
```
kron4eg, line 273:
```suggestion
	Config string `json:"config"`
```

Let's have it inline instead of file

xmudrii: I've thought about this as well and I'm not sure this is a good approach. I think we should have the standardized API, even if it's a little bit redundant sometimes. A couple of things I have on mind:

- `Config` is a struct for all other features. Having it as a string here could introduce confusion
- The problem mentioned above could be solved by calling it `ConfigFilePath` like it's now, but then we come to another problem...
- If we ever have to add another field for any reason, we would drift from the standard API way too much and we would end up in a mess.

In the long run, I think it's better as it is right now.

kron4eg:
Everything looks good, except one part that really bothers me, file references (namely `ConfigFilePath` fields), those are points to external resources that we have to dereference somehow. I know we already doing this and I'm concern about this too :)

Yet, the API design discussion probably doesn't belongs to this PR.

/lgtm
/approve

PR #923, opened 2020-06-17
Bump Go and k8s versions in the kubeone-e2e image

xmudrii:

Update the kubeone-e2e image to:
- Bump Go to 1.14.4. We have been using Go 1.14 to cut releases since 1.14 went out, so we should use 1.14 in CI as well
- Bump Kubernetes versions to the latest patch releases
- Bump the image version to v0.1.8

```release-note
NONE
```

/assign @kron4eg

### DIFF
--- a/hack/images/kubeone-e2e/Dockerfile
+++ b/hack/images/kubeone-e2e/Dockerfile
@@ -14,7 +14,7 @@
 
 # building image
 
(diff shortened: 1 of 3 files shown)


xmudrii:
/hold
for the latest patch releases to become available

PR #929, opened 2020-06-19
Fix CNI URL for CoreOS/Flatcar

xmudrii:

This PR fixes the CNI URL for CoreOS and Flatcar. Currently, it's impossible to provision/upgrade clusters on those platforms, as CNI URLs got changed after v0.7.5.

The approach used in this PR is that we now use CNI v0.8.6 for all Kubernetes versions on CoreOS/Flatcar. The same approach is used by machine-controller, although, it can be slightly risky because older versions (e.g. 1.15 and older releases of 1.16/1.17) are not verified by upstream with the latest CNI. Since we use this approach in MC anyways, I think it should be okay to proceed with this change.

```release-note
Fix CoreOS/Flatcar cluster provisioning/upgrading: use the correct URL for the CNI plugins 
```

/assign @kron4eg


xmudrii:
@kron4eg PTAL again

PR #936, opened 2020-06-25
Stop copying admin.conf to $HOME

### LINKED ISSUE #935: Upgrade process is broken on Flatcar
**What happened**:

Trying to upgrade the Flatcar cluster from 1.18.3 to 1.18.4 returns the following error:

```
INFO[15:30:43 CEST] Draining leader control plane…                node=172.31.5.226
[172.31.5.226] + export PATH=/usr/bin:/bin:/usr/sbin:/sbin:/sbin:/usr/local/bin:/opt/bin
[172.31.5.226] + PATH=/usr/bin:/bin:/usr/sbin:/sbin:/sbin:/usr/local/bin:/opt/bin
[172.31.5.226] + export KUBECONFIG=/etc/kubernetes/admin.conf
[172.31.5.226] + KUBECONFIG=/etc/kubernetes/admin.conf
[172.31.5.226] + sudo kubectl drain ip-172-31-5-226.eu-west-3.compute.internal --ignore-daemonsets --delete-local-data
[172.31.5.226] The connection to the server localhost:8080 was refused - did you specify the right host or port?
WARN[15:30:44 CEST] Task failed, error was: failed to drain leader control plane node: failed to exec command: set -xeu pipefail
export "PATH=$PATH:/sbin:/usr/local/bin:/opt/bin"

export KUBECONFIG=/etc/kubernetes/admin.conf
sudo kubectl drain ip-172-31-5-226.eu-west-3.compute.internal --ignore-daemonsets --delete-local-data
Error: failed to upgrade cluster: failed to upgrade leader control plane: failed to drain leader control plane node: failed to exec command: set -xeu pipefail
export "PATH=$PATH:/sbin:/usr/local/bin:/opt/bin"
```

The upgrade process works as expected.

Upgrade the cluster running on Flatcar from 1.18.3 to 1.18.4.

KubeOne version (`kubeone version`): `master`
Operating system: Ubuntu
Provider you're deploying cluster on: AWS
Operating system you're deploying on: Flatcar

kron4eg:
Fixes bug on instances where $HOME is located on tmpfs and is not persisted across restarts.

Fixes #935

```release-note
Stop copying admin.conf to $HOME
```

### DIFF
--- a/pkg/scripts/node.go
+++ b/pkg/scripts/node.go
@@ -18,13 +18,13 @@ package scripts
 
 const (
 	drainNodeScriptTemplate = `
-export KUBECONFIG=/etc/kubernetes/admin.conf
-sudo kubectl drain {{ .NODE_NAME }} --ignore-daemonsets --delete-local-data
+sudo KUBECONFIG=/etc/kubernetes/admin.conf \
+    kubectl drain {{ .NODE_NAME }} --ignore-daemonsets --delete-local-data
 `
 
 	uncordonNodeScriptTemplate = `
-export KUBECONFIG=/etc/kubernetes/admin.conf
-sudo kubectl uncordon {{ .NODE_NAME }}
+sudo KUBECONFIG=/etc/kubernetes/admin.conf \
+    kubectl uncordon {{ .NODE_NAME }}
 `
 )
 
--- a/pkg/addons/addons.go
+++ b/pkg/addons/addons.go
@@ -29,9 +29,8 @@ import (
 )
 
 const (
-	addonLabel = "kubeone.io/addon"
-
-	kubectlApplyScript = `kubectl apply -f {{.FILE_NAME}} --prune -l "%s"`
+	addonLabel         = "kubeone.io/addon"
+	kubectlApplyScript = `sudo kubectl --kubeconfig=/etc/kubernetes/admin.conf apply -f {{.FILE_NAME}} --prune -l "%s"`
 )
 
 // TemplateData is data available in the addons render template
--- a/pkg/addons/addons.go
+++ b/pkg/addons/addons.go
@@ -30,7 +30,10 @@ import (
 
 const (
 	addonLabel         = "kubeone.io/addon"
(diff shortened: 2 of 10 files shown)


xmudrii:
With this change, we would have to run all `kubectl` commands with `sudo`, which is not the greatest practice and there were some objections against this in the past.

Why we don't just copy the kubeconfig when starting upgrade as well? That would be a simpler change that would also work with tmpfs.

kron4eg:
> Why we don't just copy the kubeconfig when starting upgrade as well? That would be a simpler change that would also work with tmpfs.

IDK  ¯\\\_(ツ)\_/¯, why _to copy_ if we can _not to copy_?

It's easy to loose track where is what and when, meaning: copying $KUBECONFIG around will have be an "always make sure it was done" operation or make assumption that it has been done. This PR fixes one such assumption which turned to be not true. 

On the other hand admin.conf location is well known and it's always there.

pkg/scripts/node.go
```diff
@@ -18,13 +18,13 @@ package scripts
 
 const (
 	drainNodeScriptTemplate = `
-export KUBECONFIG=/etc/kubernetes/admin.conf
-sudo kubectl drain {{ .NODE_NAME }} --ignore-daemonsets --delete-local-data
+sudo KUBECONFIG=/etc/kubernetes/admin.conf \
+    kubectl drain {{ .NODE_NAME }} --ignore-daemonsets --delete-local-data
```
xmudrii, line 22: Can we make it consistent between addons and node scripts and use either flag or the environment variable?

xmudrii:
> IDK ¯\_(ツ)_/¯, why to copy if we can not to copy?
>
> It's easy to loose track where is what and when, meaning: copying $KUBECONFIG around will have be an "always make sure it was done" operation or make assumption that it has been done. This PR fixes one such assumption which turned to be not true.

That's true, but you also have to make sure that you don't forget a flag/environment variable when running kubectl commands and that you do it in a proper way. Copying would be done once at the beginning, like we do for the clientset anyways.

However, I'm okay with whatever solution. I've left a comment on the PR if we go ahead with the proposed approach.

pkg/scripts/node.go
kron4eg: 👍 done

Issue #531, opened 2019-07-01
Implement the kubeone apply command


xmudrii:
**What feature would you like to be added?**

Implement the `kubeone apply` command to handle the cluster reconciliation process, including:
* installing the cluster (`kubeone install`)
* upgrading the cluster (`kubeone upgrade`)
* reconciling manifests and machines
* repairing the cluster (the planned `kubeone repair`)

**Action items:**

* [x] Write a proposal
* [ ] Implement the `kubeone apply` command
* [ ] Deprecate `install` and `upgrade` commands

PR #942, opened 2020-07-02
Probes

kron4eg:
This PR implements foundation for probes from [apply](https://github.com/kubermatic/kubeone/blob/master/docs/proposals/20200224-apply.md) proposal

Xref #531 

```release-note
NONE
```

### DIFF
--- a/pkg/cmd/apply.go
+++ b/pkg/cmd/apply.go
@@ -1,10 +1,28 @@
+/*
+Copyright 2019 The KubeOne Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package cmd
 
 import (
-	"github.com/kubermatic/kubeone/pkg/credentials"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+
+	"github.com/kubermatic/kubeone/pkg/credentials"
+	"github.com/kubermatic/kubeone/pkg/tasks"
 )
 
 type applyOpts struct {
@@ -72,6 +90,16 @@ func runApply(opts *applyOpts) error {
 	if err != nil {
 		return errors.Wrap(err, "failed to validate credentials")
 	}
+
+	probbing := tasks.WithHostnameOS(nil)
+	probbing = tasks.WithProbes(probbing)
+
+	if err = probbing.Run(s); err != nil {
+		return err
+	}
+
+	// later in this point we going to make decision and run different tasks, should we run install or upgrade based on
+	// the state we accumulated in s.LiveCluster
 
 	return nil
 }
--- a/pkg/cmd/apply.go
+++ b/pkg/cmd/apply.go
@@ -1,0 +1,77 @@
+package cmd
+
+import (
+	"github.com/kubermatic/kubeone/pkg/credentials"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+type applyOpts struct {
+	globalOptions
+	NoInit      bool `longflag:"no-init"`
+	Force       bool `longflag:"force"`
+	AutoApprove bool `longflag:"auto-approve"`
+}
+
+func applyCmd(rootFlags *pflag.FlagSet) *cobra.Command {
+	opts := &applyOpts{}
+
+	cmd := &cobra.Command{
+		Hidden: true, // for now
+		Use:    "apply",
+		Short:  "apply reconcile the cluster",
+		Long: `
+Reconcile (Install/Upgrade/Repair/Restore) Kubernetes cluster on pre-existing machines
+
+This command takes KubeOne manifest which contains information about hosts and how the cluster should be provisioned.
+It's possible to source information about hosts from Terraform output, using the '--tfjson' flag.
+`,
+		Example: `kubeone apply -m mycluster.yaml -t terraformoutput.json`,
+		RunE: func(_ *cobra.Command, args []string) error {
+			gopts, err := persistentGlobalOptions(rootFlags)
+			if err != nil {
+				return errors.Wrap(err, "unable to get global flags")
+			}
+
+			opts.globalOptions = *gopts
+
+			return runApply(opts)
+		},
+	}
+
+	cmd.Flags().BoolVar(
+		&opts.NoInit,
+		longFlagName(opts, "NoInit"),
+		false,
+		"don't initialize the cluster (only install binaries)")
+
+	cmd.Flags().BoolVar(
+		&opts.AutoApprove,
+		longFlagName(opts, "AutoApprove"),
+		false,
+		"auto approve plan")
+
+	cmd.Flags().BoolVar(
+		&opts.Force,
+		longFlagName(opts, "Force"),
+		false,
+		"use force to install new binary versions (!dangerous!)")
+
+	return cmd
+}
+
+func runApply(opts *applyOpts) error {
+	s, err := opts.BuildState()
+	if err != nil {
+		return errors.Wrap(err, "failed to initialize State")
+	}
+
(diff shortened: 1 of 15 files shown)


pkg/cmd/apply.go
```diff
@@ -0,0 +1,105 @@
+/*
+Copyright 2019 The KubeOne Authors.
```
xmudrii, line 2: Tiny nit: this should usually be 2020 for the new files.

pkg/cmd/apply.go
```diff
@@ -0,0 +1,105 @@
...
+	NoInit      bool `longflag:"no-init"`
+	Force       bool `longflag:"force"`
+	AutoApprove bool `longflag:"auto-approve"`
+}
+
+func applyCmd(rootFlags *pflag.FlagSet) *cobra.Command {
+	opts := &applyOpts{}
+
+	cmd := &cobra.Command{
+		Hidden: true, // for now
```
xmudrii, line 39: We should consider having this command as `kubeone alpha apply` instead of hiding it. The `alpha` approach is very common in the Kubernetes community and safer than just hiding the command.

pkg/cmd/apply.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/spf13/pflag"
+
+	"github.com/kubermatic/kubeone/pkg/credentials"
+	"github.com/kubermatic/kubeone/pkg/tasks"
+)
+
+type applyOpts struct {
+	globalOptions
+	NoInit      bool `longflag:"no-init"`
+	Force       bool `longflag:"force"`
```
xmudrii, line 31: I think `Force` is a bit confusing, so I'd propose that we rename it to something that would bring more context.

pkg/cmd/apply.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	// Validate credentials
+	_, err = credentials.ProviderCredentials(s.Cluster.CloudProvider, opts.CredentialsFile)
+	if err != nil {
+		return errors.Wrap(err, "failed to validate credentials")
+	}
+
+	probbing := tasks.WithHostnameOS(nil)
+	probbing = tasks.WithProbes(probbing)
```
xmudrii, line 95: Why we don't make this function take a pointer and just change the passed variable?

pkg/cmd/root.go
```diff
@@ -109,6 +109,7 @@ func newRoot() *cobra.Command {
 
 	rootCmd.AddCommand(
```
xmudrii, line 110: Tiny nit: can we make this sorted in alphabetical order?

pkg/ssh/connection.go
```diff
@@ -273,31 +271,39 @@ func (c *connection) Close() error {
 	return c.sshclient.Close()
 }
 
-func (c *connection) Stream(cmd string, stdout io.Writer, stderr io.Writer) (int, error) {
+func (c *connection) POpen(cmd string, stdin io.Reader, stdout io.Writer, stderr io.Writer) (int, error) {
```
xmudrii, line 274: The function name might not be the clearest. What does `P` refer to? Does the name refer to the `popen` syscall? You should add a comment describing this.

pkg/ssh/connection.go
```diff
@@ -273,31 +271,39 @@ func (c *connection) Close() error {
...
+	sess.Stdin = stdin
 	sess.Stdout = stdout
 	sess.Stderr = stderr
 
 	exitCode := 0
-	err = sess.Run(strings.TrimSpace(cmd))
-	if err != nil {
-		exitCode = 1
+	if err = sess.Run(cmd); err != nil {
+		exitCode = -1
```
xmudrii, line 287: Why we use exit code `-1`? Negative exit codes are not really a thing, so this might be confusing.

pkg/ssh/connection.go
```diff
@@ -273,31 +271,39 @@ func (c *connection) Close() error {
...
 
 func (c *connection) Exec(cmd string) (string, string, int, error) {
-	var stdoutBuf, stderrBuf bytes.Buffer
+	var stdoutBuf, stderrBuf strings.Builder
 
-	exitCode, err := c.Stream(cmd, &stdoutBuf, &stderrBuf)
+	exitCode, err := c.POpen(cmd, nil, &stdoutBuf, &stderrBuf)
 
-	return strings.TrimSpace(stdoutBuf.String()), strings.TrimSpace(stderrBuf.String()), exitCode, err
+	return strings.TrimSpace(stdoutBuf.String()), stderrBuf.String(), exitCode, err
```
xmudrii, line 306: Why we dropped `strings.TrimSpace` for `stderrBuf`, but not for `stdoutBuf`?

pkg/state/cluster.go
```diff
@@ -0,0 +1,90 @@
...
+	"sync"
+
+	"github.com/Masterminds/semver"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+
+	"github.com/kubermatic/kubeone/pkg/apis/kubeone"
+)
+
+type Cluster struct {
```
xmudrii, line 29: Maybe it would make more sense to call this `ActualCluster` instead of just `Cluster`.

pkg/state/cluster.go
```diff
@@ -0,0 +1,90 @@
...
+type ComponentStatus struct {
+	Version *semver.Version
+	Status  int64
+}
+
+const (
+	SystemDStatusUnknown    = 1 << iota // systemd unit unknown
+	SystemdDStatusDead                  // systemd unit dead
+	SystemDStatusRestarting             // systemd unit restarting
+	ComponentInstalled                  // installed (package, or direct download)
```
xmudrii, line 55: Why is this one in the middle of other `systemd` related consts?

pkg/state/cluster.go
```diff
@@ -0,0 +1,90 @@
...
+	PublicAddress    string
+}
+
+type ComponentStatus struct {
+	Version *semver.Version
+	Status  int64
+}
+
+const (
+	SystemDStatusUnknown    = 1 << iota // systemd unit unknown
```
xmudrii, line 52: We might miss some `systemd` statuses here.

pkg/state/cluster.go
```diff
@@ -0,0 +1,90 @@
...
+	Kubeconfig       []byte
+	Kubernetes       ComponentStatus
+	OS               kubeone.OperatingSystemName
+	PrivateAddress   string
+	PublicAddress    string
+}
+
+type ComponentStatus struct {
+	Version *semver.Version
+	Status  int64
```
xmudrii, line 48: Why is Status type int64? What we would store there?

pkg/state/context.go
```diff
@@ -46,6 +46,7 @@ func New(ctx context.Context) (*State, error) {
 // utilities like logger
 type State struct {
 	Cluster                   *kubeoneapi.KubeOneCluster
+	LiveCluster               *Cluster
```
xmudrii, line 49: I'd probably prefer if we call it ActualCluster.

pkg/tasks/probes.go
```diff
@@ -0,0 +1,225 @@
...
+)
+
+const (
+	systemdShowStatusCMD = `systemctl show %s -p LoadState,ActiveState,SubState`
+
+	dockerVersionDPKG = `dpkg-query --show --showformat='${Version}' docker-ce | cut -d: -f2 | cut -d~ -f1`
+	dockerVersionRPM  = `rpm -qa --queryformat '%{RPMTAG_VERSION}' docker-ce`
+
+	kubeletVersionDPKG = `dpkg-query --show --showformat='${Version}' kubelet | cut -d- -f1`
+	kubeletVersionRPM  = `rpm -qa --queryformat '%{RPMTAG_VERSION}' kubelet`
```
xmudrii, line 39: We should also probe for other Kubernetes packages, like `kubeadm` and `kubernetes-cni`. We should also maybe check is the version of packages the expected one.

pkg/cmd/apply.go
kron4eg: `kubeone alpha apply` is the next stage, when it will actually start working :D 
For now, while it's not doing what it's suppose to, and while we are moving it towards "working state" — it's hidden.

pkg/state/cluster.go
kron4eg: it's just another status, it's free form name.

pkg/state/cluster.go
kron4eg: we don't need them all, only those that we are querying and later reading.

pkg/tasks/probes.go
kron4eg: we will, in future PRs

PR #947, opened 2020-07-02
Update machine-controller to v1.15.1 and add the imagePlan field for Azure

xmudrii:

Update machine-controller to v1.15.1 and add the ImagePlan field for Azure.

The type of the `ImagePlan` field is different than for other fields. This time, the field type is a struct, and so far we only used primitive types. I don't have access to Azure, so I can't this is this going to work. I assume it should work this way, but I'm not sure.

```release-note
Update machine-controller to v1.15.1
Add the ImagePlan field for Azure Terraform integration
```

### DIFF
--- a/pkg/templates/machinecontroller/cloudprovider_specs.go
+++ b/pkg/templates/machinecontroller/cloudprovider_specs.go
@@ -121,6 +121,7 @@ type AzureSpec struct {
 	RouteTableName    string            `json:"routeTableName"`
 	SecurityGroupName string            `json:"securityGroupName"`
 	Zones             []string          `json:"zones"`
+	ImagePlan         *AzureImagePlan   `json:"imagePlan"`
 	SubnetName        string            `json:"subnetName"`
 	Tags              map[string]string `json:"tags"`
 	VMSize            string            `json:"vmSize"`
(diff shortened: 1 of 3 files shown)


pkg/templates/machinecontroller/cloudprovider_specs.go
```diff
@@ -121,6 +121,7 @@ type AzureSpec struct {
 	RouteTableName    string            `json:"routeTableName"`
 	SecurityGroupName string            `json:"securityGroupName"`
 	Zones             []string          `json:"zones"`
+	ImagePlan         *AzureImagePlan   `json:"ImagePlan"`
```
xmudrii, line 124: * This should eventually be `omitempty`, since it's a pointer, but there's no `omitempty` in the MC type, so I didn't add it here.
* The capital `I` in `ImagePlan` is expected because it's also like that in the MC type.

pkg/terraform/v1beta1/config.go
```diff
@@ -495,6 +496,11 @@ func setWorkersetFlag(w *kubeonev1beta1.DynamicWorkerConfig, name string, value
 		if s == nil {
 			return nil
 		}
+	case machinecontroller.AzureImagePlan:
```
xmudrii, line 499: Probably not needed because we use the pointer type, but I added it anyway because I can't test is it going to work without it.

pkg/templates/machinecontroller/cloudprovider_specs.go
kron4eg: ouch, that's a BUG in the MC

xmudrii: MC 1.15.1 fixes this issue. I've updated the PR to use 1.15.1 instead of 1.15.0. PTAL.

PR #951, opened 2020-07-03
Improvements to probes, Kubelet initialized

xmudrii:
```release-note
NONE
```

/assign @kron4eg

### DIFF
--- a/pkg/tasks/probes.go
+++ b/pkg/tasks/probes.go
@@ -32,8 +32,12 @@ import (
 const (
 	systemdShowStatusCMD = `systemctl show %s -p LoadState,ActiveState,SubState`
 
-	dockerVersionCMD  = `sudo docker version -f "{{ .Server.Version }}"`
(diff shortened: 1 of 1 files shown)


pkg/tasks/probes.go
```diff
@@ -32,11 +32,10 @@ import (
 const (
 	systemdShowStatusCMD = `systemctl show %s -p LoadState,ActiveState,SubState`
 
-	dockerVersionDPKG = `dpkg-query --show --showformat='${Version}' docker-ce | cut -d: -f2 | cut -d~ -f1`
-	dockerVersionRPM  = `rpm -qa --queryformat '%{RPMTAG_VERSION}' docker-ce`
+	dockerVersionCMD  = `sudo docker version -f "{{ .Server.Version }}"`
+	kubeletVersionCMD = `kubelet --version | cut -d' ' -f2`
```
xmudrii, line 36: We should probably check for both the version and the package. On the other side, if some of this commands fail, maybe we can assume that there's no package.

PR #968, opened 2020-07-10
Read Openstack values from clouds.yaml

### LINKED ISSUE #965: Openstack: Allow to use clouds.yaml

I would like to add support for the [clouds.yaml](https://docs.openstack.org/python-openstackclient/pike/configuration/index.html), terraform already supports this (see: https://www.inovex.de/blog/managing-secrets-openstack-terraform). 

This feature allows that I can use my existing `clouds.yaml` without redefining all the `OS_` environment variables. IMHO the usage of the `clouds.yaml` should be preferred over the environment variable.

My idea would be to check if the `OS_CLOUD` variable is set, if so read the content of `clouds.yaml` otherwise use the environment variables.

johscheuer:
Signed-off-by: Johannes M. Scheuermann <joh.scheuer@gmail.com>

This PR allows to use the Openstack setting from a clouds.yaml like described in https://github.com/kubermatic/kubeone/issues/965 without the need to additionally specify the same variables ass env vars.

Fixes #965 

```release-note
Openstack values can be read from the Opentstack clouds.yaml
```

### DIFF
--- a/pkg/credentials/credentials.go
+++ b/pkg/credentials/credentials.go
@@ -24,11 +24,12 @@ import (
 	"strings"
 
 	"github.com/aws/aws-sdk-go/aws/credentials"
-	"github.com/pkg/errors"
-	"gopkg.in/yaml.v2"
-
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack"
 	"github.com/gophercloud/utils/openstack/clientconfig"
 	"github.com/kubermatic/kubeone/pkg/apis/kubeone"
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v2"
 )
 
 // The environment variable names with credential in them
@@ -114,15 +115,7 @@ func ProviderCredentials(p kubeone.CloudProviderSpec, credentialsFilePath string
 			{Name: HetznerTokenKey, MachineControllerName: HetznerTokenKeyMC},
 		}, defaultValidationFunc)
 	case p.Openstack != nil:
-		return f.parseCredentialVariables([]ProviderEnvironmentVariable{
-			{Name: OpenStackAuthURL},
-			{Name: OpenStackUserName, MachineControllerName: OpenStackUserNameMC},
-			{Name: OpenStackPassword},
-			{Name: OpenStackDomainName},
-			{Name: OpenStackRegionName},
-			{Name: OpenStackTenantID},
-			{Name: OpenStackTenantName},
-		}, openstackValidationFunc)
+		return f.parseOpenStackCredentials()
 	case p.Packet != nil:
 		return f.parseCredentialVariables([]ProviderEnvironmentVariable{
 			{Name: PacketAPIKey, MachineControllerName: PacketAPIKeyMC},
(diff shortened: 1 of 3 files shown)


pkg/credentials/credentials.go
```diff
@@ -27,6 +27,7 @@ import (
 	"github.com/pkg/errors"
 	"gopkg.in/yaml.v2"
 
+	"github.com/gophercloud/utils/openstack/clientconfig"
```
xmudrii, line 30: Nit: this import should go in a group above.

pkg/credentials/credentials.go
```diff
@@ -235,6 +236,11 @@ func (f *fetcher) parseAWSCredentials() (map[string]string, error) {
 }
 
 func (f fetcher) parseCredentialVariables(envVars []ProviderEnvironmentVariable, validationFunc func(map[string]string) error) (map[string]string, error) {
+	// Load variables from clouds.yaml
+	if err := f.parseOpenstackCloudCredentials(); err != nil {
+		return nil, errors.Wrap(err, "error reading clouds.yaml")
+	}
+
```
xmudrii, line 243: We should run this function only if the provider is OpenStack. We don't need to read OpenStack credentials if the cluster is not running on OpenStack and that wouldn't be expected by users.

Also, we believe that the highest priority should still be on the environment variables, with a `clouds.yaml` being a fall-back mechanism. This is very important to preserve the backward compatibility, as well as, that we have a standardized mechanism across all providers. For example, we support `~/.aws/credentials` file for AWS, but the default source is still the environment variables.

You should consider the following changes:

* Create a dedicated function for parsing OpenStack credentials (e.g. `parseOpenStackCredentials`) like the one that's used for [AWS (`parseAWSCredentials`)](https://github.com/kubermatic/kubeone/blob/68c604054b6e2bcc6f451c4470326de1644e4af8/pkg/credentials/credentials.go#L189-L235)
  * The mechanism should be the same as for AWS. We first respect the [credentials file](https://github.com/kubermatic/kubeone/blob/68c604054b6e2bcc6f451c4470326de1644e4af8/pkg/credentials/credentials.go#L190-L195), then the [environment variables](https://github.com/kubermatic/kubeone/blob/68c604054b6e2bcc6f451c4470326de1644e4af8/pkg/credentials/credentials.go#L197-L212), and then at the end [AWS profile file](https://github.com/kubermatic/kubeone/blob/68c604054b6e2bcc6f451c4470326de1644e4af8/pkg/credentials/credentials.go#L214-L232).
  * Similar to AWS, I'd propose that `clouds.yaml` is only respected if the appropriate environment variable is set (I see that you already did that in a function that you added)
  * I'd recommend that we take credentials only from a single source. If we can't fetch all credentials from the credentials file/environment, we forget about what we picked up and try to read everything from `clouds.yaml`. If `clouds.yaml` is also missing some required credentials, we fail. The reason for this is that the environment variables and `clouds.yaml` might point to different OS instances or have different credentials, which can make issues that are hard to debug.
* Update the `ProviderCredentials` function to use the newly added function. The [following snippet](https://github.com/kubermatic/kubeone/blob/68c604054b6e2bcc6f451c4470326de1644e4af8/pkg/credentials/credentials.go#L116-L124) should be replaced with a function call like for [AWS](https://github.com/kubermatic/kubeone/blob/68c604054b6e2bcc6f451c4470326de1644e4af8/pkg/credentials/credentials.go#L88).

pkg/credentials/credentials.go
```diff
@@ -266,11 +272,50 @@ func defaultValidationFunc(creds map[string]string) error {
 	return nil
 }
 
+func (f *fetcher) parseOpenstackCloudCredentials() error {
+	cloudName, ok := os.LookupEnv("OS_CLOUD")
+
+	if !ok {
+		return nil
+	}
```
xmudrii, line 280: `os.LookupEnv` returns true even if it's set but is empty. I'd consider `clouds.yaml` only if there's some content in the environment variable, e.g. name of the configuration to be used.

johscheuer:
Thanks for the feedback, I will rework the PR.

pkg/credentials/credentials.go
johscheuer: Thanks for the feedback I implemented the changes.

johscheuer:
I would update the docs once the PR looks good.

pkg/credentials/credentials.go
```diff
@@ -234,6 +228,73 @@ func (f *fetcher) parseAWSCredentials() (map[string]string, error) {
...
+			OpenStackTenantName: f.F(OpenStackTenantName),
+			OpenStackUserNameMC: f.F(OpenStackUserName),
+		}, nil
+	}
+
+	// 2.) Check environment variables only use them if all required parameters are set
+	opts, err := openstack.AuthOptionsFromEnv()
+	if err == nil {
+		return openstackAuthOptionsToCreds(&opts, os.Getenv("OS_REGION_NAME"))
+	}
```
xmudrii, line 269: There are a couple of things I don't like about this approach. I've investigated this function and found out that:

* It fetches all `OS_*` variables. This seems a little bit wrong if we use only a few of them
* It validates `OS_*` variables that we don't use, for example https://github.com/gophercloud/gophercloud/blob/c8c96f74de7f0c91297732d1a1468bc225a68137/openstack/auth_env.go#L99-L110
  * This also seems a little bit wrong because there's no point in validating and blocking on variables that we don't use.
* Their validation is not strict as ours

There are (at least) two potential approaches to this:

* Acknowledge this and go ahead with his approach. The problems with this approach are outlined above
* Continue using `os.Getenv` for the environment variables and use library only for `clouds.yml`

In the second case, we might run into inconsistencies between how environment variables and `clouds.yml` are handled.

Therefore, I'd probably prefer using the approach you proposed in this PR, but we are currently having some internal discussions about this, as well as, we want to test it a little bit.

pkg/credentials/credentials.go
```diff
@@ -265,22 +326,3 @@ func defaultValidationFunc(creds map[string]string) error {
 	}
 	return nil
 }
-
-func openstackValidationFunc(creds map[string]string) error {
```
xmudrii, line 269: I would keep this validation function because the validation in the library is not as strict as our validation. A couple of examples:

* the validation would pass if [`OS_USERID` is set instead of `OS_USERNAME`](https://github.com/gophercloud/gophercloud/blob/c8c96f74de7f0c91297732d1a1468bc225a68137/openstack/auth_env.go#L67-L75), but we don't support `OS_USERID`
* the validation would pass if [`OS_PASSCODE` is set instead of `OS_PASSWORD`](https://github.com/gophercloud/gophercloud/blob/c8c96f74de7f0c91297732d1a1468bc225a68137/openstack/auth_env.go#L77-L83)

pkg/credentials/credentials.go
```diff
@@ -234,6 +228,73 @@ func (f *fetcher) parseAWSCredentials() (map[string]string, error) {
...
+	creds[OpenStackDomainName] = opts.DomainName
+	creds[OpenStackRegionName] = region
+
+	if opts.TenantID != "" {
+		creds[OpenStackTenantID] = opts.TenantID
+	}
+
+	if v, ok := creds[OpenStackTenantID]; !ok || len(v) == 0 {
+		creds[OpenStackTenantName] = opts.TenantName
+	}
```
xmudrii, line 246: There is a behavior change here that we need to mention in the release note and the docs.

Previously, we just respected the `OS_TENANT_ID` and `OS_TENANT_NAME` environment variables. The library works a bit different as it would set TenantID and TenantName to `OS_PROJECT_ID` and `OS_PROJECT_NAME` values respectively (if those environment variables are present).

> OS_TENANT_ID and OS_TENANT_NAME are deprecated forms of OS_PROJECT_ID and
OS_PROJECT_NAME and the latter are expected against a v3 auth api.

The comment makes it clear and I'm okay with this change, but it should be documented once this PR gets accepted.

pkg/credentials/credentials.go
```diff
@@ -234,6 +228,73 @@ func (f *fetcher) parseAWSCredentials() (map[string]string, error) {
...
+			OpenStackDomainName: f.F(OpenStackDomainName),
+			OpenStackRegionName: f.F(OpenStackRegionName),
+			OpenStackTenantID:   f.F(OpenStackTenantID),
+			OpenStackTenantName: f.F(OpenStackTenantName),
+			OpenStackUserNameMC: f.F(OpenStackUserName),
+		}, nil
+	}
+
+	// 2.) Check environment variables only use them if all required parameters are set
+	opts, err := openstack.AuthOptionsFromEnv()
```
xmudrii, line 266: @johscheuer We just discussed this and we came with an idea to avoid using this function, but instead, write a function that would initialize `gophercloud.AuthOptions{}` manually. It's supposed to fetch and validate just the needed environment variables.

Does this idea sound good to you and can you modify the PR to implement it?

PR #994, opened 2020-07-21
Verify that crd.projectcalico.org CRDs established

### LINKED ISSUE #934: Canal CNI defective in rare cases
**What happened**:

When installing a KubeOne cluster, in rare cases the Canal CNI pods are shown to be running but not usable. The entire KubeOne installation fails when deploying the machine-controller in this case. Machine-controller and CoreDNS pods stay in pending state due to the following condition:

`Failed to create pod sandbox: rpc error: code = Unknown desc = [failed to set up sandbox container "6675d4875be32a0a78d89e208f3fb61d09b72e597e6565ee14121e69336006f8" network for pod "coredns-6955765f44-pggjl": networkPlugin cni failed to set up pod "coredns-6955765f44-pggjl_kube-system" network: error getting ClusterInformation: resource does not exist: ClusterInformation(default) with error: the server could not find the requested resource (get ClusterInformations.crd.projectcalico.org default), failed to clean up sandbox container "6675d4875be32a0a78d89e208f3fb61d09b72e597e6565ee14121e69336006f8" network for pod "coredns-6955765f44-pggjl": networkPlugin cni failed to teardown pod "coredns-6955765f44-pggjl_kube-system" network: error getting ClusterInformation: resource does not exist: ClusterInformation(default) with error: the server could not find the requested resource (get ClusterInformations.crd.projectcalico.org default)]`

Canal's calico container logs the following error:

`2020-06-24 11:20:44.067 [ERROR][8] client.go 251: Error creating cluster information config ClusterInformation=&v3.ClusterInformation{TypeMeta:v1.TypeMeta{Kind:"ClusterInformation", APIVersion:"projectcalico.org/v3"}, ObjectMeta:v1.ObjectMeta{Name:"default", GenerateName:"", Namespace:"", SelfLink:"", UID:"4e7c4b94-e00d-437b-959a-e277c807caf7", ResourceVersion:"", Generation:0, CreationTimestamp:v1.Time{Time:time.Time{wall:0xbfb4eba303d8c464, ext:6639844301, loc:(*time.Location)(0x2d61840)}}, DeletionTimestamp:(*v1.Time)(nil), DeletionGracePeriodSeconds:(*int64)(nil), Labels:map[string]string(nil), Annotations:map[string]string(nil), OwnerReferences:[]v1.OwnerReference(nil), Initializers:(*v1.Initializers)(nil), Finalizers:[]string(nil), ClusterName:"", ManagedFields:[]v1.ManagedFieldsEntry(nil)}, Spec:v3.ClusterInformationSpec{ClusterGUID:"f1988d2388444411b0bd7f42db3471f8", ClusterType:"k8s,canal,kdd", CalicoVersion:"v3.10.0", DatastoreReady:(*bool)(0xc0002b2973)}} error=resource does not exist: ClusterInformation(default) with error: the server could not find the requested resource (post ClusterInformations.crd.projectcalico.org)`

The CRDs in question are available after this. So this looks like a race condition, caused by CRDs not usable yet when the calico container is started. Deleting the canal pods solves the problem.

Maybe separating the CRD deployment and waiting until it is actually available makes the cluster creation more robust.

Appears occasionally during a `kubeone install`

Might be reinforced by an unresponsive etcd ring due to https://github.com/kubermatic/kubeone/issues/915

KubeOne version (`kubeone version`): v1.0.0-alpha.6
Operating system: CentOS (master)
Operating system you're deploying on: OSX

kron4eg:
Verify that crd.projectcalico.org CRDs established when using Canal CNI

Fixes #934

```release-note
Verify that crd.projectcalico.org CRDs established  when using Canal CNI
```

### DIFF
--- a/pkg/clientutil/crds.go
+++ b/pkg/clientutil/crds.go
@@ -25,7 +25,7 @@ import (
 )
 
 var (
-	CRDNotEstablishedErr = errors.New("crd is not established")
+	ErrCRDNotEstablished = errors.New("crd is not established")
 )
 
 func VerifyCRD(ctx context.Context, client dynclient.Client, crdName string) (bool, error) {
@@ -41,5 +41,5 @@ func VerifyCRD(ctx context.Context, client dynclient.Client, crdName string) (bo
 		}
 	}
 
-	return false, CRDNotEstablishedErr
+	return false, ErrCRDNotEstablished
 }
--- a/pkg/clientutil/crds.go
+++ b/pkg/clientutil/crds.go
@@ -1,0 +1,45 @@
+/*
+Copyright 2020 The KubeOne Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package clientutil
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
+	dynclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	CRDNotEstablishedErr = errors.New("crd is not established")
+)
+
+func VerifyCRD(ctx context.Context, client dynclient.Client, crdName string) (bool, error) {
+	crd := &apiextensions.CustomResourceDefinition{}
+	key := dynclient.ObjectKey{Name: crdName}
+	if err := client.Get(ctx, key, crd); err != nil {
+		return false, err
+	}
+
+	for _, cond := range crd.Status.Conditions {
+		if cond.Type == apiextensions.Established && cond.Status == apiextensions.ConditionTrue {
+			return true, nil
+		}
+	}
+
+	return false, CRDNotEstablishedErr
+}
--- a/pkg/clientutil/crds.go
+++ b/pkg/clientutil/crds.go
@@ -19,7 +19,7 @@ package clientutil
 import (
 	"context"
 
-	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
+	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	dynclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
--- a/pkg/clientutil/crds.go
+++ b/pkg/clientutil/crds.go
@@ -19,27 +19,30 @@ package clientutil
 import (
 	"context"
 
-	"github.com/pkg/errors"
 	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
 	dynclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-var (
-	ErrCRDNotEstablished = errors.New("crd is not established")
-)
+// CRDsReadyCondition generate a k8s.io/apimachinery/pkg/util/wait.ConditionFunc function to be used in
+// k8s.io/apimachinery/pkg/util/wait.Poll* family of functions. It will check all provided GKs (GroupKinds) to exists
+// and have Established status
+func CRDsReadyCondition(ctx context.Context, client dynclient.Client, groupKinds []string) func() (bool, error) {
+	return func() (bool, error) {
+		for _, gk := range groupKinds {
+			crd := apiextensions.CustomResourceDefinition{}
+			key := dynclient.ObjectKey{Name: gk}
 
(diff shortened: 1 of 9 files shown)


pkg/clientutil/wait_pods.go
```diff
@@ -0,0 +1,51 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package clientutil
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
```
xmudrii, line 22: Nit: this should go into a dedicated group.

pkg/clientutil/crds.go
```diff
@@ -0,0 +1,45 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package clientutil
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
```
xmudrii, line 22: Nit: this should go into a dedicated group.

pkg/clientutil/crds.go
```diff
@@ -0,0 +1,45 @@
...
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
+	dynclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	ErrCRDNotEstablished = errors.New("crd is not established")
+)
+
+func VerifyCRD(ctx context.Context, client dynclient.Client, crdName string) (bool, error) {
```
xmudrii, line 31: Nit: should we call function `CRDEstablished` or something like that to keep the consistency with other functions (e.g. `PodsReady`)?

pkg/templates/machinecontroller/helper.go
```diff
@@ -20,14 +20,13 @@ import (
 	"context"
 	"time"
 
+	clusterv1alpha1 "github.com/kubermatic/machine-controller/pkg/apis/cluster/v1alpha1"
```
xmudrii, line 23: Nit: this should go into the kubermatic group.

pkg/templates/machinecontroller/helper.go
```diff
@@ -62,77 +61,51 @@ func WaitReady(s *state.State) error {
...
 	_ = wait.Poll(5*time.Second, 3*time.Minute, func() (bool, error) {
 		lastErr = VerifyCRDs(s)
 		return lastErr == nil, nil
 	})
+
 	return errors.Wrap(lastErr, "failed waiting for CRDs to become ready and established")
 }
 
 // VerifyCRDs verifies are Cluster-API CRDs deployed
 func VerifyCRDs(s *state.State) error {
```
xmudrii, line 91: I don't like this function's name because it's too similar to `VerifyCRD`.

pkg/clientutil/crds.go
xmudrii, line 31: Why we don't make this function poll for multiple CRDs (i.e. take a slice of names)? I think this is our most common use case, so it would make sense to have such a function.

pkg/clientutil/crds.go
```diff
@@ -0,0 +1,48 @@
...
+			key := dynclient.ObjectKey{Name: gk}
+
+			if err := client.Get(ctx, key, &crd); err != nil {
+				return false, err
+			}
+
+			for _, cond := range crd.Status.Conditions {
+				if cond.Type == apiextensions.Established && cond.Status != apiextensions.ConditionTrue {
+					return false, nil
+				}
```
xmudrii, line 42: I was thinking about this approach and I'm not sure that it's safe at all.

This check assumes that the Established condition always exists on the CRD object, but I'm not sure is that true. If `client.Get` returns the result before CRD controller sets the initial conditions, it maybe could happen that `crd.Status.Conditions` don't contain any Established condition. In that case, this check would pass and the function would return true, while the CRD is not Established.

I believe that the following check would be safer

```suggestion
				if cond.Type == apiextensions.Established && cond.Status == apiextensions.ConditionTrue {
					return true, nil
				}
```

It needs some other changes, as well as, it should be tested.

pkg/clientutil/wait_pods.go
```diff
@@ -0,0 +1,61 @@
...
+
+	for _, pod := range podsList.Items {
+		if pod.Status.Phase != corev1.PodRunning {
+			return false
+		}
+
+		for _, podcond := range pod.Status.Conditions {
+			if podcond.Type == corev1.PodReady && podcond.Status != corev1.ConditionTrue {
+				return false
+			}
```
xmudrii, line 56: Similar to the Established condition, should we check is `ConditionTrue` present and in that case, return true?

pkg/clientutil/crds.go
kron4eg: Changing it to such form means introducing a bug. We want to check all of CRDs, but in this form the first found CRD will report positive on behalf of all CRDs under test.

xmudrii: In our case, I believe something like this would work and not introduce a bug:
```go
var found bool
for _, cond := range crd.Status.Conditions {
    if cond.Type == apiextensions.Established && cond.Status == apiextensions.ConditionTrue {
        found = true
        break
    }
}
if !found {
    return false, nil
}
```

kron4eg: yeah, changed to something similar

PR #1005, opened 2020-07-28
Improve networking

kron4eg:

* Upgrade Canal to v3.15.1
* Make ubuntu 20.04 work unconditionally
* Canal CNI got new MTU field
* Default Canal MTU, taking into account different providers
* Label everything we deploy with "kubeone.io/component: <COMPONENT-NAME>"

Fixes #

```release-note
* Upgrade Canal to v3.15.1
* Make ubuntu 20.04 work unconditionally
* Canal CNI got new MTU field
* Default Canal MTU, taking into account different providers
* Label everything we deploy with "kubeone.io/component: <COMPONENT-NAME>"
```

### DIFF
--- a/pkg/apis/kubeone/types.go
+++ b/pkg/apis/kubeone/types.go
@@ -183,7 +183,7 @@ type CNI struct {
 
 // CanalSpec defines the Canal CNI plugin
 type CanalSpec struct {
-	MTU string `json:"mtu"`
+	MTU int `json:"mtu"`
 }
 
 // WeaveNetSpec defines the WeaveNet CNI plugin
--- a/pkg/apis/kubeone/types.go
+++ b/pkg/apis/kubeone/types.go
@@ -182,7 +182,9 @@ type CNI struct {
 }
 
 // CanalSpec defines the Canal CNI plugin
-type CanalSpec struct{}
+type CanalSpec struct {
+	MTU string `json:"mtu"`
+}
 
 // WeaveNetSpec defines the WeaveNet CNI plugin
 type WeaveNetSpec struct {
--- a/pkg/apis/kubeone/v1beta1/defaults.go
+++ b/pkg/apis/kubeone/v1beta1/defaults.go
@@ -34,8 +34,8 @@ const (
 	DefaultNodePortRange = "30000-32767"
 	// DefaultStaticNoProxy defined static NoProxy
 	DefaultStaticNoProxy = "127.0.0.1/8,localhost"
-	// Default MTU
-	DefaultMTU = 1500
+	// DefaultVXLanMTU defines default VXLAN MTU for Canal CNI
+	DefaultCanalMTU = 1450
 )
 
 func addDefaultingFuncs(scheme *runtime.Scheme) error {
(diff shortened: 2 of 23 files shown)


pkg/apis/kubeone/v1beta1/defaults.go
```diff
@@ -34,6 +34,8 @@ const (
 	DefaultNodePortRange = "30000-32767"
 	// DefaultStaticNoProxy defined static NoProxy
 	DefaultStaticNoProxy = "127.0.0.1/8,localhost"
+	// Default MTU
+	DefaultMTU = 1500
```
xmudrii, line 38: I don't see that this variable is used anywhere.

pkg/apis/kubeone/v1beta1/defaults.go
```diff
@@ -127,8 +129,23 @@ func SetDefaults_ClusterNetwork(obj *KubeOneCluster) {
...
+		switch {
+		case obj.CloudProvider.AWS != nil:
+			canal.MTU = "8951" // 9001 AWS Jumbo Frame - 50 VXLAN bytes
+		case obj.CloudProvider.GCE != nil:
+			canal.MTU = "1410" // GCE specific 1460 bytes - 50 VXLAN bytes
+		case obj.CloudProvider.Hetzner != nil:
+			canal.MTU = "1400" // Hetzner specific 1450 bytes - 50 VXLAN bytes
+		case obj.CloudProvider.Openstack != nil:
+			canal.MTU = "1400" // Openstack specific 1450 bytes - 50 VXLAN bytes
+		}
```
xmudrii, line 145: MTU should be defaulted regardless of is `clusterNetwork.CNI == nil`. We should skip defaulting only if `MTU != ""`.

pkg/apis/kubeone/types.go
```diff
@@ -182,7 +182,9 @@ type CNI struct {
 }
 
 // CanalSpec defines the Canal CNI plugin
-type CanalSpec struct{}
+type CanalSpec struct {
+	MTU string `json:"mtu"`
```
xmudrii, line 186: Why is MTU a string and not an int?

pkg/clientutil/create_or_update.go
```diff
@@ -28,8 +28,20 @@ import (
...
+	return func(c client.Client, obj runtime.Object) {
+		LabelComponent(componentname, obj)
+	}
+}
+
 // CreateOrUpdate makes it easy to "apply" objects to kubernetes API server
-func CreateOrUpdate(ctx context.Context, c client.Client, obj runtime.Object) error {
+func CreateOrUpdate(ctx context.Context, c client.Client, obj runtime.Object, updaters ...Updater) error {
+	for _, update := range updaters {
+		update(c, obj)
```
xmudrii, line 42: Where is this update function defined? I don't see it.

pkg/apis/kubeone/types.go
kron4eg: Because it's being consumed as string

pkg/clientutil/create_or_update.go
kron4eg: right here, `WithComponentLabel()`

pkg/apis/kubeone/v1beta1/defaults.go
kron4eg: If I'll assign to obj.ClusterNetwork.CNI regardless of its nilness, that will override other possible options selected by the user (weave / external), isn't it?

kron4eg:
@xmudrii comments addressed

PR #1015, opened 2020-08-03
Fix new v1.130 golangci-lint errors

kron4eg:
Signed-off-by: Artiom Diomin <kron82@gmail.com>

golangci-lint v1.130 introduces new checks, which is fixed.

Fixes #

```release-note
Fix new v1.130 golangci-lint errors
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,7 @@ buildenv:
 .PHONY: lint
 lint:
 	@golangci-lint --version
-	golangci-lint run ./pkg/... ./test/...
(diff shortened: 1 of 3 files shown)


hack/tools.go
```diff
@@ -19,7 +19,6 @@ limitations under the License.
 package tools
 
 import (
-	_ "k8s.io/code-generator"
```
xmudrii, line 22: @kron4eg It seems like this change breaks the verify codegen job.

kron4eg: yeah, reverted it

PR #1020, opened 2020-08-05
Increase default number of task retries to 10

### LINKED ISSUE #915: Cluster creation on Azure unreliable
**What happened**:

After the kubeadm calls, the cluster is often in a state where frequent etcd leader changes happen and API calls fail because of this. This causes the `kubeone install` to fail frequently because installing the CNI or machine-controller was not possible. The problem seem to be temporary as a second `kubeone install` run fixes the cluster.

Create kubeone cluster on Azure.

Tried to switch to more powerful Azure resources than given in the Terraform example. Faster storage and VM types didn't bring a significant change. However, when increasing the number of retries that are performed by KubeOne, the installation is successful.

KubeOne output:
...
INFO[21:41:46 CEST] Downloading kubeconfig…
INFO[21:41:46 CEST] Building Kubernetes clientset…
INFO[21:41:47 CEST] Ensure node local DNS cache…
INFO[21:41:47 CEST] Activating additional features…
INFO[21:41:49 CEST] Applying canal CNI plugin…
WARN[21:41:55 CEST] Task failed…
WARN[21:41:55 CEST] error was: failed to get *v1beta1.CustomResourceDefinition object: etcdserver: leader changed
WARN[21:42:00 CEST] Retrying task…
INFO[21:42:00 CEST] Applying canal CNI plugin…
WARN[21:42:13 CEST] Task failed…
WARN[21:42:13 CEST] error was: failed to get *v1beta1.CustomResourceDefinition object: etcdserver: leader changed
WARN[21:42:23 CEST] Retrying task…
INFO[21:42:23 CEST] Applying canal CNI plugin…
WARN[21:42:32 CEST] Task failed…
WARN[21:42:32 CEST] error was: failed to get *v1.ConfigMap object: etcdserver: leader changed
Error: failed to install cni plugin: failed to get *v1.ConfigMap object: etcdserver: leader changed
failed to install cni plugin: failed to get *v1.ConfigMap object: etcdserver: leader changed
ERROR[0412] Error installing kubernetes:

KubeOne version (`kubeone version`): v1.0.0-alpha
Operating system: CentOS

kron4eg:
Any cloud activity could be temporary failure, we should retry harder

Fixes #915

```release-note
Increase default number of task retries to 10
```

### DIFF
--- a/pkg/tasks/task.go
+++ b/pkg/tasks/task.go
@@ -45,7 +45,7 @@ type Task struct {
 // Run runs a task
 func (t *Task) Run(s *state.State) error {
 	if t.Retries == 0 {
-		t.Retries = 3
+		t.Retries = 10
 	}
 
 	backoff := defaultRetryBackoff(t.Retries)


pkg/tasks/task.go
```diff
@@ -45,7 +45,7 @@ type Task struct {
 // Run runs a task
 func (t *Task) Run(s *state.State) error {
 	if t.Retries == 0 {
-		t.Retries = 3
+		t.Retries = 10
```
xmudrii, line 48: I don't think that it makes a lot of sense to increase the number of retries to 10 by default. That seems way too high for many tasks. I think we should do that only for specific tasks.

Also, the problem mentioned in the linked issue is supposed to be mitigated with #994. Do we have some idea did that fix helped?

kron4eg: I think 10 is too small actually, checkout the number of retries in kubermatic code (especially shell) for example. Yes, maybe some parts don't have too much sense to retry ever 2 times. I prefer to have many retries by default and maybe exclude some parts to opposite approach (no or small retries, with some parts explicitly have big retries).

kron4eg: >Do we have some idea did that fix helped?

It was initially known that elevated number of retries would fix the problem.

xmudrii: > I think 10 is too small actually, checkout the number of retries in kubermatic code (especially shell) for example.

The number of retries depend on how the functionality is implemented. Note that we use the exponential backoff, so with each retry, the user is going to wait a bit more each time.

> Yes, maybe some parts don't have too much sense to retry ever 2 times. I prefer to have many retries by default and maybe exclude some parts to opposite approach (no or small retries, with some parts explicitly have big retries).

That's why I think it's best to do this on the task level and keep some sane default number of retries. I can think of a few tasks that don't make sense to retry a lot, like installing the binaries or running kubeadm.

On top of that, I think it's important to understand how each task works and to make them behave consistently. For example, some tasks already have implemented retries in the task logic, so adding more retries would just increase the time spent on useless waiting.

kron4eg: >and keep some sane default number of retries

Totally agree. I just see that the sane number is bigger then 3, because 3 retries are "spent" superfast even with exponential backoff, simply because there is no room for exponent to be considerable long.

kron4eg: >installing the binaries
>running kubeadm

Both case fall because of intermediate networking issues, like for example temporary down DNS that will be up in 30 seconds. This is exact demonstration why we should keep this number relatively high by default. Because "we don't know" in what unpredictable way everything may go south "in this particular instance".

PR #1035, opened 2020-08-12
kubeone proxy command

### LINKED ISSUE #719: kubeone tunnel proxy
`kubeone proxy` subcommand, that will open SSH tunnel and will act as local HTTPS proxy to provide kube-api access in case when it's not available from the internet (e.g. internal/firewalled LB).

`kubeone proxy` should print instructions for the user on start up.

With merge of https://github.com/kubermatic/kubeone/pull/714 kubeone access to kube-api is going to originate from the instances using SSH tunnels. In order to provide access kube-api from localhost we need a HTTPS proxy for kubectl that will also tunnel requests.

kron4eg:
Launch local HTTPS capable proxy (CONNECT method) to tunnel HTTPS request via SSH. This is especially useful in cases when kubeapi endpoint is not accessible from the public internets.

Fixes #719 

```release-note
added: kubeone proxy command
```

### DIFF
--- a/pkg/cmd/proxy.go
+++ b/pkg/cmd/proxy.go
@@ -1,0 +1,143 @@
+/*
+Copyright 2020 The KubeOne Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


pkg/cmd/proxy.go
```diff
@@ -0,0 +1,143 @@
...
+
+func proxyCmd(rootFlags *pflag.FlagSet) *cobra.Command {
+	opts := &proxyOpts{}
+
+	cmd := &cobra.Command{
+		Use:   "proxy",
+		Short: "Proxy to the kube-apiserver using SSH tunnel",
+		Long: `
+HTTPS Proxy (CONNECT method) SSH tunnel.
+`,
```
xmudrii, line 44: Should we include some details when this command is used, i.e. mention that you need this if your load balancer is not publicly accessible?

kron4eg: done

PR #1058, opened 2020-08-18
Insert small delay before starting node-local-dns

### LINKED ISSUE #1037: race condition for pod starting order results in not working clusterIPs in case nf_tables must be used
**What happened**:

Current major versions of Linux distributions don't support iptables-legacy anymore. Instead nf_tables is used (e.g. RHEL8 or debian Buster).
The availability of nf_tables only leads to a race condition for starting kube-proxy and node-local-dns in the correct order after a node is started.

Currently node-local-dns supports only iptables-legacy. For the support of nf_tables an open/blocked [issue](https://github.com/kubernetes/dns/issues/338) exists.
kube-proxy supports both iptables modes (see [here](https://github.com/kubernetes/kubernetes/pull/82966)) and determines during it's starting phase what it should use (see [here](https://github.com/kubernetes/kubernetes/pull/82966/files#diff-c60536f6b948668f1b8aa52529a35f5aR28)).
This means, if the node-local-dns pod on the node is starting first and creates it's iptables-legacy rules, kube-proxy finds these legacy-rules and starts using the legacy mode too. Unfortunately, kube-proxy uses some chains which the kubelet creates when it starts. E.g. the chain "KUBE_MARK-DROP". Because the OS is offering nf_talbles only, the kubelet creates the particular chains with nf_tables and not iptables-legacy.
If now kube-proxy is starting in iptables-legacy mode, it tries to write to the kubelet chains and fails, becaue it cannot find the nf_tables chains:
```
kubectl -n kube-system logs $(kubectl -n kube-system get pods -o wide | grep proxy | grep cp-0 | awk '{print $1}') -f
W0813 15:17:48.190861       1 server_others.go:559] Unknown proxy mode "", assuming iptables proxy
I0813 15:17:48.203720       1 node.go:136] Successfully retrieved node IP: 172.16.10.5
I0813 15:17:48.203751       1 server_others.go:186] Using iptables Proxier.
I0813 15:17:48.205552       1 server.go:583] Version: v1.18.6
I0813 15:17:48.206007       1 conntrack.go:52] Setting nf_conntrack_max to 131072
I0813 15:17:48.206845       1 config.go:315] Starting service config controller
I0813 15:17:48.206863       1 shared_informer.go:223] Waiting for caches to sync for service config
I0813 15:17:48.206886       1 config.go:133] Starting endpoints config controller
I0813 15:17:48.206898       1 shared_informer.go:223] Waiting for caches to sync for endpoints config
I0813 15:17:48.307068       1 shared_informer.go:230] Caches are synced for endpoints config 
I0813 15:17:48.307209       1 shared_informer.go:230] Caches are synced for service config 
E0813 15:17:48.347064       1 proxier.go:1555] Failed to execute iptables-restore: exit status 2 (iptables-restore v1.8.3 (legacy): Couldn't load target `KUBE-MARK-DROP':No such file or directory

Error occurred at line: 84
Try `iptables-restore -h' or 'iptables-restore --help' for more information.
)
I0813 15:17:48.347141       1 proxier.go:825] Sync failed; retrying in 30s
```
For reference, see
- [kube-proxy code](https://github.com/kubernetes/kubernetes/blob/release-1.18/pkg/proxy/iptables/proxier.go#L74)
- [github discussion](https://github.com/kubernetes/kubernetes/issues/80462#issuecomment-517502159)
- [temporary fix in kube-proxy container](https://github.com/kubernetes/kubernetes/issues/80462)

Because of this kube-proxy is entering an endless loop, trying to write to the chain KUBE-MARK-DROP and never creates the iptables rules for clusterIPs.

On the other side, if kube-proxy is starting before node-local-dns is creating it's iptables-legacy rules, the kube-proxy creates all rules using nf_tables. Hence, in this case the chain KUBE-MARK-DROP exists and everything is working as expected.

Working clusterIPs after each node start.

Rebooting nodes for testing the race condition.

For a running node delete the running kube-proxy pod and flush the nf_tables rules on the host
```
#remove/flush all rules & delete chains
iptables -F
iptables -X
iptables -t nat -F
iptables -t nat -X
iptables -t mangle -F
iptables -t mangle -X
iptables -P INPUT ACCEPT
iptables -P OUTPUT ACCEPT
iptables -P FORWARD ACCEPT
```

The root cause is also hitting calico, see [here](https://github.com/projectcalico/calico/issues/3709).

I tested to use kube-proxy in ipvs mode. Unfortunately, I hit here another blocking issue for Azure (most likely for other public cloud providers as well), regarding services of type LoadBalancer. Hence, using kube-proxy in ipvs mode is not an option.

Because the root cause is the not patched node-local-dns pod I think a good option could be to disable it's deployment for now. Maybe introducing node-local-dns as a feature in kubeone which can be deactivated (e.g. like PodSecurityPolicies)?

KubeOne version (`1.0.0-beta.2`): 
Operating system: RHEL8
Operating system you're deploying on: CentOS8

kron4eg:
On some slow at the boot systems it may happen (race condition) that nodelocal dns will start before kube-proxy and that will mess up whole iptables setup. To make room for kube-proxy to be first to run we delay node-local-dns at the start s little bit.

This is TEMPORARY workaround, until the time when node-local-dns would start supporting iptables-nft

Fixes #1037 

```release-note
Fix race condition between kube-proxy and node-local-dns
```

### DIFF
--- a/pkg/templates/nodelocaldns/nodelocaldns.go
+++ b/pkg/templates/nodelocaldns/nodelocaldns.go
@@ -39,7 +39,7 @@ const VirtualIP = "169.254.20.10"
 
 const (
 	image                    = "k8s.gcr.io/k8s-dns-node-cache"
-	tag                      = "1.15.12"
+	tag                      = "1.15.13"
 	componentLabel           = "nodelocaldns"
 	dnscacheCorefileTemplate = `
 __PILLAR__DNS__DOMAIN__:53 {
@@ -188,6 +188,14 @@ func dnscacheDaemonSet() *appsv1.DaemonSet {
 	hostPathFileOrCreate := corev1.HostPathFileOrCreate
 	terminationGracePeriodSeconds := int64(0)
 
+	// "sleep 10" is needed to avoid race-condition for iptables creation between node-local-cache and kube-proxy, to
+	// make sure that kube-proxy will insert its iptables rules first.
+	// see more at: https://github.com/kubermatic/kubeone/pull/1058
+	execScript := fmt.Sprintf(`
+sleep 10;
+exec /node-cache -localip %s -conf /etc/Corefile -upstreamsvc kube-dns-upstream`,
+		VirtualIP)
+
 	return &appsv1.DaemonSet{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "node-local-dns",
@@ -225,13 +233,10 @@ func dnscacheDaemonSet() *appsv1.DaemonSet {
 						{
 							Name:  "node-cache",
 							Image: fmt.Sprintf("%s:%s", image, tag),
-							Args: []string{
-								"-localip",
-								VirtualIP,
-								"-conf",
-								"/etc/Corefile",
-								"-upstreamsvc",
-								"kube-dns-upstream",
+							Command: []string{
+								"/bin/sh",
+								"-c",
+								execScript,
 							},
 							Resources: corev1.ResourceRequirements{
 								Requests: corev1.ResourceList{
--- a/pkg/clientutil/wait_pods.go
+++ b/pkg/clientutil/wait_pods.go
@@ -33,7 +33,7 @@ func PodsReadyCondition(ctx context.Context, c dynclient.Client, listOpts dyncli
 		podsList := corev1.PodList{}
 
 		if err := c.List(ctx, &podsList, &listOpts); err != nil {
-			return false, errors.Wrapf(err, "failed to list %s pods", listOpts.FieldSelector.String())
+			return false, errors.Wrapf(err, "failed to list pods")
 		}
 
 		return allPodsAreRunningAndReady(&podsList), nil


pkg/templates/nodelocaldns/nodelocaldns.go
```diff
@@ -187,6 +187,10 @@ func dnscacheDaemonSet() *appsv1.DaemonSet {
 	trueBool := true
 	hostPathFileOrCreate := corev1.HostPathFileOrCreate
 	terminationGracePeriodSeconds := int64(0)
+	execScript := fmt.Sprintf(`
+sleep 10;
+exec /node-cache -localip %s -conf /etc/Corefile -upstreamsvc kube-dns-upstream`,
+		VirtualIP)
```
xmudrii, line 193: Would it make sense that instead of sleeping 10 seconds, we poll for the kube-proxy DaemonSet/pods before deploying NodeLocalDNS and wait for all replicas to be ready? I think that would be a safer approach, especially if kube-proxy needs more than 10 seconds to come up for any reason.

If we decide to stick with this approach, can we add a comment describing why we did this?

kron4eg: No, because the racecondition also kicks in after node reboot, when there is no kubeone involved.

xmudrii: Okay, makes sense. Please add a comment describing that above this line.

kron4eg: added an explanatory comment

xmudrii (approved):
/lgtm

PR #1066, opened 2020-08-24
Update hetzner example to use hetzner loadbalancer service

cebor:

Hetzner now has officially load balancers as a service, i replaced the gobetween service with the hetzner load balancer service. Also added it to the same network and use targets via private ips.

Restrictions: Seems to only work with terraform >= 0.13. 

```release-note
NONE
```


examples/terraform/hetzner/main.tf
```diff
@@ -34,7 +34,7 @@ resource "hcloud_network_subnet" "kubeone" {
 resource "hcloud_server_network" "control_plane" {
   count       = 3
   server_id  = element(hcloud_server.control_plane.*.id, count.index)
-  network_id = hcloud_network.net.id
+  subnet_id = hcloud_network_subnet.kubeone.id
 }
```
xmudrii, line 38: Nit: can you please run `terraform fmt`?

examples/terraform/hetzner/versions.tf
```diff
@@ -1,4 +1,4 @@
 
 terraform {
-  required_version = ">= 0.12"
+  required_version = ">= 0.13"
```
xmudrii, line 3: I've successfully provisioned infra using this config and Terraform v0.12. What problems have you encountered with v0.12? Here's the output of `terraform version` for me:

```
Terraform v0.12.23
+ provider.hcloud v1.20.1

Your version of Terraform is out of date! The latest version
is 0.13.0. You can update by downloading from https://www.terraform.io/downloads.html
```

examples/terraform/hetzner/main.tf
```diff
@@ -54,61 +54,37 @@ resource "hcloud_server" "control_plane" {
...
-    content     = local.rendered_lb_config
-    destination = "/etc/gobetween.toml"
-  }
-
-  provisioner "remote-exec" {
-    inline = [
-      "systemctl restart gobetween",
-    ]
-  }
+resource "hcloud_load_balancer_service" "load_balancer_service" {
```
xmudrii, line 85: I think we should add a health check here like we have for other providers. Here's how it looks like for AWS: https://github.com/kubermatic/kubeone/blob/acbe987ec016e6db2aaff7a2f1c8cc3a431f20e8/examples/terraform/aws/main.tf#L207-L213

examples/terraform/hetzner/main.tf
```diff
@@ -54,61 +54,37 @@ resource "hcloud_server" "control_plane" {
...
-    script = "gobetween.sh"
-  }
 }
 
-locals {
-  rendered_lb_config = templatefile("./etc_gobetween.tpl", {
-    lb_targets = hcloud_server_network.control_plane.*.ip,
-  })
+resource "hcloud_load_balancer_target" "load_balancer_target" {
+  type             = "server"
```
xmudrii, line 74: I think that we should use the `label_selector` LB type. We can use these labels, but we can also add new ones if needed

https://github.com/kubermatic/kubeone/blob/dc1cb7ff6378bfda16d04f90f4cb202286514792/examples/terraform/hetzner/main.tf#L51-L54

xmudrii (commented):
@cebor Thank you for the PR! I've left a bunch of comments about things that could be eventually improved, but overall, it looks good to me!

examples/terraform/hetzner/versions.tf
cebor: If no providers are installed i got following error:

```
$ terraform init                                                                                 

Initializing the backend...

Initializing provider plugins...
- Finding latest version of hashicorp/hcloud...

Error: Failed to install providers

Could not find required providers, but found possible alternatives:

  hashicorp/hcloud -> terraform-providers/hcloud

If these suggestions look correct, upgrade your configuration with the
following command:
    terraform 0.13upgrade .
```

examples/terraform/hetzner/main.tf
cebor: done

examples/terraform/hetzner/main.tf
cebor: Health checks will be added by default depending on which protocol are you using, in this case tcp.

image

examples/terraform/hetzner/main.tf
cebor: I tried the following:

```
type                   = "label_selector"
label_selector   = "kubeone_cluster_name=${var.cluster_name},role=api"
```

But this will prevent manual adding of servers to the lb targets. Hetzner will automatically add new servers with matching label. For an initial infrastructure state both will do the same.

examples/terraform/hetzner/versions.tf
kron4eg: Yeah, I was able to provision it with 0.12 successfully too. So let's please not upgrade terraform. Especially given that we use 0.12 in e2e tests.

xmudrii: I agree with @kron4eg. Let's keep 0.12, as all other configs and E2E tests are still using 0.12.

examples/terraform/hetzner/main.tf
xmudrii: I think it would be better to use HTTP pointing to the `/healthz` endpoint if it works. @kron4eg do you have any opinions?

examples/terraform/hetzner/main.tf
xmudrii: > Hetzner will automatically add new servers with matching label. For an initial infrastructure state both will do the same.

That sounds okay to me. That's the same as we do for other providers as well. I think it's better because if any change happens, it will automatically update the load balancer as well. @kron4eg any opinions?

examples/terraform/hetzner/main.tf
kron4eg: Well, previous config with gobetween didn't done this too (http `GET /healthz`). At least we can try it like it's now and then later add some more stuff.

examples/terraform/hetzner/main.tf
kron4eg: Changes doesn't happen outside of the terraform, supposedly 😅. So it's OK to have `type = "server"`

examples/terraform/hetzner/versions.tf
kron4eg: OK, let's get rid of this `required_version = ">= 0.13"` and this PR is good to merge

cebor: done!

xmudrii (approved):
@cebor Thank you for the PR!

/lgtm
/approve
/test all

PR #1075, opened 2020-08-27
allow to use internal goproxy, add caching for go builds

xrstf:
This configures all Prow jobs to use our internal Go proxy, to gain a tiny bit more stability and speed. Shouldn't hurt anyway.
This has uncovered that KubeOne used a machine-controller version that uses a non-existing version for `openshift/api`, which was only resolvable by Google's GOPROXY. Our internal cache didn't know the version, so I bumbed the mchine-controller version.

The PR also adds caching for the `$GOCACHE` directory, similar to Kubermatic's approach but without a lockfile.

```release-note
NONE
```


kron4eg (changes requested):
I don't see the point of this PR

xmudrii (approved):
I think that it’s an okay change. In the worst case, nothing will happen, so I’m okay that we test it and see what happens.

/lgtm
/hold
Not sure is the PR ready yet.

xrstf:
> I don't see the point of this PR

Wasting less time on the CI cluster benefits everyone. :-)

kron4eg:
@xrstf please don't get me wrong, local GOPROXY might be beneficial (only for the sake of not depending directly on upsteam GOPROXY), but upload/download gocache is just useless in case of kubeone

xmudrii:
> @xrstf please don't get me wrong, local GOPROXY might be beneficial (only for the sake of not depending directly on upsteam GOPROXY), but upload/download gocache is just useless in case of kubeone.

Can you explain a little bit why is it useless? I think not depending on the upstream GOPROXY and using the internal one could definitely speed up jobs, so I think we should at least give it a try. It's not going to be some significant difference, as building KubeOne is already relatively fast, but we are going to reduce the network load a little bit and it may get somewhat faster. At least, we can always revert if there are no improvements.

> That's 2 additional bash scripts that I'd like not to have.

IMO, those two scripts are not bad at all. I think they are relatively easy to understand, and I don't see that there's anything to maintain there at all.

kron4eg:
On my very old laptop (2013 year of production) it takes 26 seconds to build kubeone with cold gobuild cache (not gomod, only gobuild, the one in `go env GOCACHE`), which translates to ~ 2% for total individual test time (from total 20 minutes). I'm pretty sure CI builds happens at least as twice as faster. Is 1% of build time worth the potential issues (like failed build because minio is dead yet again) those 2 scripts bring in? In my opinion — absolutely not.

kron4eg:
On the contrary, local GOPROXY is beneficial because it can reveal potential issues in our go.mod, like divergence between reality and upstream GOPROXY.

xmudrii:
> Is 1% of build time worth the potential issues (like failed build because minio is dead yet again) those 2 scripts bring in? In my opinion — absolutely not.

If Minio is down, our tests are going to fail anyway because Terraform depends on the Minio backend. Although, if something happens to the proxy, we are in the problem.

I'm not really sure how should we proceed. I'll leave it up to you and @xrstf to decide.

kron4eg:
In #1126 I've copied relevant bits regarding GOPROXY. Thanks @xrstf for inspiration :)

kron4eg:
Superseded by #1126

PR #1099, opened 2020-09-10
Don't stop Kubelet when upgrading Kubeadm on Flatcar

xmudrii:

For some reason, we stop Kubelet when upgrading Kubeadm and CNI binaries, but we don't start it until the upgrade is fully done. Usually, and as per [the upgrade instructions](https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/), restarting Kubelet should be done as the last step.

I don't know the reason why we introduced this, but it might be some leftover when copying code from another function.

/hold
for manual testing

```release-note
Don't stop Kubelet when upgrading Kubeadm on Flatcar
```

/assign @kron4eg

### DIFF
--- a/pkg/scripts/os.go
+++ b/pkg/scripts/os.go
@@ -281,7 +281,6 @@ sudo curl -L --remote-name-all \
 
 sudo mkdir -p /opt/bin
 cd /opt/bin
-sudo systemctl stop kubelet
 sudo mv /var/tmp/kube-binaries/kubeadm .
 sudo chmod +x kubeadm
 `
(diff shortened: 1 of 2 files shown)


kron4eg:
IIRC the binary was locked (if launched) and this prevented us from copying in place the new binary

xmudrii:
@kron4eg But we don't copy Kubelet in that step, we copy only Kubeadm in that function. I'll test and confirm does it work or not.

kron4eg:
Probably after refactoring the script was changed and now this stopping is not relevant anymore

kron4eg (approved):

xmudrii:
I've manually tested this PR and it seems to work well, so I'll remove the hold.
/hold cancel

Issue #1085, opened 2020-09-01
containerd support


kron4eg:
**What feature would you like to be added?**
containerd should be an option for Container Runtime as a docker replacement

**What are use cases of the feature?**
Lower system requirements, lower feature surface (equals to less bugs).

qwinkler:
Any updates/planned release date? Also, will there be an ability to specify the runtime version (containerd/docker)?

PR #1101, opened 2020-09-10
New v1beta1 .ContainerRuntime field

kron4eg:

Ref #1085 

```release-note
New v1beta1 .ContainerRuntime API field
```


pkg/apis/kubeone/types.go
```diff
@@ -29,35 +29,48 @@ import (
...
 	Addons *Addons `json:"addons,omitempty"`
-	// SystemPackages configure kubeone behaviour regarding OS packages
+	// SystemPackages configure kubeone behaviour regarding OS packages.
 	SystemPackages *SystemPackages `json:"systemPackages,omitempty"`
 }
 
+// ContainerRuntimeConfig
+type ContainerRuntimeConfig struct {
+	Docker     *ContainerRuntimeDocker     `json:"docker"`
+	Containerd *ContainerRuntimeContainerd `json:"containerd"`
```
xmudrii, line 65: Since those two fields are pointers, why they don't have omitempty?

pkg/apis/kubeone/validation/validation.go
```diff
@@ -177,6 +179,27 @@ func ValidateVersionConfig(version kubeone.VersionConfig, fldPath *field.Path) f
...
+	for _, x := range allCRs {
+		if !reflect.ValueOf(x).IsNil() {
+			if found {
+				allErrs = append(allErrs, field.Invalid(fldPath, x, "only 1 container runtime can be activated"))
+			}
+			found = true
+		}
+	}
+
+	return allErrs
```
xmudrii, line 200: As [validation happens after defaulting](https://github.com/kubermatic/kubeone/blob/3835bcac539cd4df037edd0b7505542d622f1fae/pkg/apis/kubeone/config/config.go#L165-L168), we should test is ContainerRuntimeConfig set, i.e. something like:

```
if !found {
    // err...
}
```

kron4eg: if validation happens after defaulting, it means that at least `cr.Docker` will be set (not nil), so `if !found` situation is ruled out.

pkg/apis/kubeone/types.go
kron4eg: Fixed

pkg/apis/kubeone/validation/validation.go
xmudrii: While I agree with that, I think validating could be useful to confirm that defaulting works (e.g. can help to catch potential bugs). However, I'm not going to hold on this.

xmudrii (approved):
/lgtm

PR #1127, opened 2020-10-01
Status logging

s-spindler:
Sorry but I just had to fix the message regarding the progress after reading it several times a day in the past days...

Regarding the ellipsis ligature: I know this is to some degree a matter of personal preference. I'm not a fan of it as it distracts my eyes from the code. It's just weird to see those special characters sprinkled around. Also, if I were to add a new log statement with ellipsis I would have to go and copy one from somewhere else since I have no idea how to type one out on my keyboard.
Lastly, if your terminal doesn't support unicode they will print some mush instead of the ellipses.

There are some cool code fonts that present ellipsis (and some special characters) as ligatures, e.g. [FiraCode](https://github.com/tonsky/FiraCode).

I can easily pop off the commit that changes the ellipses. Let me know what you think.

**Does this PR introduce a user-facing change?**: no

```release-note
NONE
```

### DIFF
--- a/pkg/addons/addons.go
+++ b/pkg/addons/addons.go
@@ -44,10 +44,10 @@ type TemplateData struct {
 
 func Ensure(s *state.State) error {
 	if s.Cluster.Addons == nil || !s.Cluster.Addons.Enable {
-		s.Logger.Infoln("Skipping applying addons because addons are not enabled…")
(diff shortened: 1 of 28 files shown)


kron4eg:
@xmudrii PTAL

xmudrii (approved):
/approve

kron4eg:
Thanks @s-spindler!

PR #1134, opened 2020-10-13
Use KubeletConfiguration Go type instead of a YAML template

xmudrii:

This PR proposes the usage of the KubeletConfiguration type, instead of using the template, to generate the Kubelet configuration manifest.

I've compared the Kubelet config manifests, and there is no difference in manifests between the current and the proposed approaches.

```release-note
NONE
```

/assign @kron4eg


kron4eg:
The reason why we copied kubeadm types is that they where located inside k/k repository and is was very painful to import it. But it's not the case with kubelet as it has a dedicated modules with kubeletConfiguration family of types. So I propose to better use k8s.io/kubelet  module

xmudrii:
@kron4eg I've thought the same, but I've decided to go for the consistency. Anyways, I've reworked the PR to vendor the Kubelet types instead of porting them. It should be much better now, so PTAL when you can.

kron4eg (approved):

PR #1139, opened 2020-10-16
Add the changelog for the v1.0.4 release

xmudrii:

Add the changelog for the v1.0.4 release.

```release-note
NONE
```

/assign @kron4eg

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,28 @@
 # Changelog
 
+# [v1.0.4](https://github.com/kubermatic/kubeone/releases/tag/v1.0.4) - 2020-10-16
+
+## Attention Needed
+
+* KubeOne now creates a dedicated secret with vSphere credentials for kube-controller-manager and vSphere Cloud Controller Manager (CCM). This is required because those components require the secret to adhere to the expected format.
+  * The new secret is called `vsphere-ccm-credentials` and is deployed in the `kube-system` namespace.
+  * This fix ensures that you can use all vSphere provider features, for example, volumes and having cloud provider metadata/labels applied on each node.
+  * If you're upgrading an existing vSphere cluster, you should take the following steps:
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
...
+
+## Attention Needed
+
+* KubeOne now creates a dedicated secret with vSphere credentials for kube-controller-manager and vSphere Cloud Controller Manager (CCM). This is required because those components require the secret to adhere to the expected format.
+  * The new secret is called `vsphere-ccm-credentials` and is deployed in the `kube-system` namespace.
+  * This fix ensures that you can use all vSphere provider features, for example, volumes and having cloud provider metadata/labels applied on each node.
+  * If you're upgrading an existing vSphere cluster, you should take the following steps:
+    * Upgrade the cluster without changing the cloud-config. Upgrading the cluster creates a new Secret for kube-controller-manager and vSphere CCM.
+    * Change the cloud-config in your KubeOne Configuration Manifest to refer to the new secret.
+    * Upgrade the cluster again to apply the new cloud-config, or change it manually on each control plane node and restart kube-controller-manager and vSphere CCM (if deployed).
```
xmudrii, line 13: The reasoning for the two-steps upgrade is that if there's no secret deployed (it's deployed after `kubeadm upgrade`) and the kube-controller-manager gets configured to use it, it might fail to mount volumes, which might cause problems with the workload.

However, we can probably assume that it was never possible to mount volumes when using the old secret, so it should be safe to upgrade it in a single step. But, if someone is switching from the username/password in a cloud-config to the secret approach, then it might cause problems.

Having a single-step upgrade would be much easier and more efficient, so I guess we can also recommend that, eventually with a warning about this.

kron4eg (approved):
@xmudrii two steps upgrade looks good.

PR #1155, opened 2020-11-02
Add support for insecure Docker registries

xmudrii:

This PR adds support for using the insecure Docker registry with the OverwriteRegistry functionality.

The RegistryConfiguration API has been extended with a new field `insecureRegistry`, that is used such as:

```yaml
registryConfiguration:
  overwriteRegistry: "127.0.0.1:5000"
  insecureRegistry: true
```

If `insecureRegistry` is set to `true`, the registry specified in the `overwriteRegistry` field will be configured as an insecure registry.

It's up to discussion do we want this field to be a boolean, or maybe a slice of strings, where users would be able to specify the custom insecure registries.

```release-note
Add support for insecure Docker registries
```

/assign @kron4eg

### DIFF
--- a/pkg/apis/kubeone/helpers.go
+++ b/pkg/apis/kubeone/helpers.go
@@ -113,3 +113,13 @@ func (r *RegistryConfiguration) ImageRegistry(defaultRegistry string) string {
 	}
 	return defaultRegistry
 }
+
+// InsecureRegistryAddress returns the registry that should be configured
+// as insecure if InsecureRegistry is enabled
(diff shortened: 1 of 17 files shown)


kron4eg:
How about `cluster.RegistryConfiguration.InsecureRegistry()` method that will incapsulate the logic, and simply using it directly in the Data{}?

xmudrii:
@kron4eg Fixed, PTAL.

kron4eg (approved):

kron4eg (approved):

PR #1161, opened 2020-11-06
Overwrite pause image on worker nodes

xmudrii:

Overwrite the pause image (`k8s.gcr.io/pause`) on worker nodes if overwriteRegistry is configured. The pause image version is determined using `kubeadm`.

The pause image on the control plane nodes is already overwritten by `kubeadm`.

```release-note
Overwrite pause image on worker nodes if overwriteRegistry is used
```

/assign @kron4eg

### DIFF
--- a/pkg/scripts/kubeadm.go
+++ b/pkg/scripts/kubeadm.go
@@ -63,6 +63,12 @@ rm -rf "{{ .WORK_DIR }}"
 
 	kubeadmUpgradeLeaderScriptTemplate = `
 sudo {{ .KUBEADM_UPGRADE }} --config={{ .WORK_DIR }}/cfg/master_0.yaml`
+
+	kubeadmPauseImageVersionScriptTemplate = `
+sudo kubeadm config images list --kubernetes-version={{ .KUBERNETES_VERSION }} |
+  grep "k8s.gcr.io/pause" |
+  cut -d ":" -f2
+`
 )
 
 func KubeadmJoin(workdir string, nodeID int, verboseFlag string) (string, error) {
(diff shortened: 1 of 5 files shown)


pkg/scripts/kubeadm.go
```diff
@@ -63,6 +63,13 @@ rm -rf "{{ .WORK_DIR }}"
 
 	kubeadmUpgradeLeaderScriptTemplate = `
 sudo {{ .KUBEADM_UPGRADE }} --config={{ .WORK_DIR }}/cfg/master_0.yaml`
+
+	kubeadmPauseImageVersionScriptTemplate = `
+sudo kubeadm config images list \
+  --kubernetes-version={{ .KUBERNETES_VERSION }} \
+  | grep "k8s.gcr.io/pause" \
+  | cut -d ":" -f2
+`
```
kron4eg, line 72:
```suggestion
sudo kubeadm config images list --kubernetes-version={{ .KUBERNETES_VERSION }} |
   grep "k8s.gcr.io/pause" |
   cut -d ":" -f2
`
```

pkg/templates/machinecontroller/deployment.go
```diff
@@ -765,6 +765,8 @@ func machineControllerDeployment(cluster *kubeoneapi.KubeOneCluster, credentials
 
 		args = append(args, "-node-hyperkube-image", hyperkubeImage)
 		args = append(args, "-node-kubelet-repository", poseidonKubeletImage)
+
+		args = append(args, "-node-pause-image", pauseImage)
```
kron4eg, line 769: when `if s.Cluster.RegistryConfiguration == nil || s.Cluster.RegistryConfiguration.OverwriteRegistry == "" {` is true, this pauseImage will be empty string, and machine-controller probably will blowup, this flag should be added under condition `if pauseImage != ""`.

xmudrii: If that condition is true, `if cluster.RegistryConfiguration != nil && cluster.RegistryConfiguration.OverwriteRegistry != "" {` on L762 would be false, so we wouldn't set the `-node-pause-image` flag (or any other relevant flag). I don't think that we need additional check.

pkg/scripts/kubeadm.go
xmudrii: Fixed.

kron4eg (approved):

PR #1166, opened 2020-11-18
Support for the static workers defined in terraform

### LINKED ISSUE #1021: Provide Example/Template/Document to Include Static Workers in Terraform Output

While the included terraform examples can write the control_plane nodes as static output, they don't cover static worker nodes. Instead they include templates/definitions for the cloud provider to be used with machine-controller. In case machine-controller is not used for any reasons, the user has to manually edit the kubeone manifest and add the static workers. Therefore, inclusion of static-workers output in the template would be very useful in ordere to avoid manual editing of config files.

A quick and dirty solution could be to include documentation on what kind of input can be provided from terraform output and how it has to be formatted in order to allow kubeone to use it during k8s-cluster deployment. 

- Testing of a future bare metal deployment in a cloud environment
- If a user does not want/cannot use machine-controller in its cloud deployment
- If a private cloud installation does not support some features

### LINKED ISSUE #1146: Parse staticWorkers from Terraform json input

In order to be able to deploy and update a cluster with Terraform managed worker (static) nodes, it would be great to have these also parsed from the Terraform json input. E.g. when called as:

```shell
kubeone apply -t tf.json
```

with tf.json including:

```yaml
staticWorkers:
  hosts:
   - publicAddress: 'X.Y.Z.W'
     hostname: 'worker0'
     sshUsername: ubuntu
     sshPrivateKeyFile: ssh-key
     sshPort: 22
```

As above.

Many thanks for considering.

kron4eg:
**What this PR does / why we need it**

Fixes #1021
Fixes #1146

```release-note
Support for the static workers defined in terraform
```

### DIFF
--- a/examples/terraform/aws/main.tf
+++ b/examples/terraform/aws/main.tf
@@ -292,6 +292,28 @@ resource "aws_instance" "control_plane" {
   )
 }
 
+resource "aws_instance" "static_workers1" {
+  count                  = var.static_workers_count
+  instance_type          = var.worker_type
+  iam_instance_profile   = aws_iam_instance_profile.profile.name
+  ami                    = local.ami
+  key_name               = aws_key_pair.deployer.key_name
+  vpc_security_group_ids = [aws_security_group.common.id]
+  availability_zone      = data.aws_availability_zones.available.names[count.index % length(data.aws_availability_zones.available.names)]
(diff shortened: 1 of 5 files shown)


xmudrii (approved):
/lgtm

PR #1174, opened 2020-11-25
Feature machine deployment annotations

### LINKED ISSUE #1133: Allow MachineDeployment annotations to be set via KubeOne

We would like the ability to set MachineDeployment annotations when provisioning a cluster with KubeOne (in our use case, specifically via the Terraform `kubeone_workers` outputs)

Among other possible use cases, we want to set the [cluster-autoscaler clusterapi annotations](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/clusterapi/README.md#enabling-autoscaling) alongside the other attributes for our worker node groups.

ffuerste:
**What this PR does / why we need it**: It allows to define `.meteadata.annotations` for `MachineDeployment` resources in a KubeOne configuration file (`.dynamicWorkers.providerSpec.annotations`).
Among other possible use cases, it allows to define the [cluster-autoscaler clusterapi annotations](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/clusterapi/README.md#enabling-autoscaling) alongside other attributes for worker node groups in a KubeOne config file.

Fixes #1133

The PR is based on the current kubeone master branch. The `zz_generated.*` files were created with the provided script `hack/update-codegen.sh` and verified with `hack/verify-codegen.sh`.

```release-note
No breaking changes are introduced with this PR.
User can optionally define annotations for `dynamicWorkers` elements in a KubeOne config file, which are used for the corresponding `MachineDeployment` resource.
```


pkg/apis/kubeone/v1alpha1/types.go
```diff
@@ -180,6 +180,7 @@ type WorkerConfig struct {
 // ProviderSpec describes a worker node
 type ProviderSpec struct {
 	CloudProviderSpec   json.RawMessage   `json:"cloudProviderSpec"`
+	Annotations         map[string]string `json:"annotations,omitempty"`
```
xmudrii, line 183: The v1alpha1 is frozen, so I don't think we should change it.

@kron4eg Any opinions on this?

kron4eg: Yeah, let's not touch frozen stuff

ffuerste: I included it for `v1alpha1` too, because of [this line](https://github.com/kubermatic/kubeone/blob/master/pkg/templates/machinecontroller/machines.go#L132). Here `v1alpha1` is used, so it's needed in the types.
Is it save to update this line to `v1beta1`?

xmudrii: Those are different APIs. The API you've linked is Cluster-API v1alpha1 ([imported from machine-controller](https://github.com/kubermatic/kubeone/blob/ec0e690a684fbbbac0473c73d571ea207238ea96/pkg/templates/machinecontroller/machines.go#L32)), while this is the KubeOne's API.

ffuerste: I included it for `v1alpha1` because of the conversions. Otherwise I get the following warning for `hack/update-codegen.sh`:
```
Generating deepcopy funcs
Generating defaulters
Generating conversions
E1125 15:24:08.526211    8987 conversion.go:755] Warning: could not find nor generate a final Conversion function for k8c.io/kubeone/pkg/apis/kubeone.ProviderSpec -> k8c.io/kubeone/pkg/apis/kubeone/v1alpha1.ProviderSpec
E1125 15:24:08.526277    8987 conversion.go:756]   the following fields need manual conversion:
E1125 15:24:08.526281    8987 conversion.go:758]       - Annotations
```
If this is fine, I can remove the `Annotations` field for `v1alpha`.

@xmudrii @kron4eg  What is your opinion about this?

ffuerste: You guys are too fast.

Yeah, I realized I mixed APIs here and deleted my comment. Sorry for the mess.

xmudrii: You can solve that by adding a custom conversion function that would call the autogenerated one. The autogenerated function doesn't have a conversion for the `Annotations` field, but we don't need it as that field doesn't exist in the `v1alpha1` API.

You can do it such as:

- Remove the Annotations field from the `v1alpha1` API
- Run the code generation script (`./hack/update-codegen.sh`)
- Add the following function to the [`pkg/apis/kubeone/v1alpha1/conversion.go` file](https://github.com/kubermatic/kubeone/blob/master/pkg/apis/kubeone/v1alpha1/conversion.go)
```go
func Convert_kubeone_ProviderSpec_To_v1alpha1_ProviderSpec(in *kubeoneapi.ProviderSpec, out *ProviderSpec, s conversion.Scope) error {
	if err := autoConvert_kubeone_ProviderSpec_To_v1alpha1_ProviderSpec(in, out, s); err != nil {
		return err
	}

	// The Annotations field is not available in the v1alpha1 API.

	return nil
}
```
- Re-run the code generation to ensure everything is okay

ffuerste: Done.

kron4eg (approved):

xmudrii (approved):
Thank you for the PR!

/lgtm
/approve

ffuerste:
Thanks for merging and the amazing fast replies.

PR #1175, opened 2020-11-26
Improve Amazon Linux 2 support and ensure apply command works

xmudrii:

* Improve Amazon Linux 2 OS scripts
  * Add remove binaries script
  * Don't use sudo for creating tmp folder and downloading binaries
  * Install CNI and binaries to /opt
  * Symlink kubelet, kubeadm, and kubectl to /usr/bin
* Add unit tests for Amazon Linux OS scripts
* Add apply version checks for Amazon Linux
* Update test fixtures

The reason for symlinking to /usr/bin instead of to /usr/local/bin is
that Amazon Linux 2 doesn't have /usr/local/bin in the PATH by default.
The breaks KubeOne scripts and kubeadm preflight checks.

```release-note
Improve Amazon Linux 2 support
```

/assign @kron4eg

### DIFF
--- a/pkg/scripts/os.go
+++ b/pkg/scripts/os.go
@@ -218,10 +218,10 @@ sudo yum install -y \
 	yum-plugin-versionlock \
 	device-mapper-persistent-data \
 	lvm2 \
-	conntrack \
+	conntrack-tools \
 	ebtables \
 	socat \
-	tc
+	iproute-tc
 
 {{- if or .FORCE .UPGRADE }}
 sudo yum versionlock delete docker || true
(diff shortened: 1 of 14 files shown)


pkg/scripts/os.go
```diff
@@ -231,28 +231,26 @@ sudo yum install -y \
...
-sudo tar xvf node.tar.gz
+curl -L --output /tmp/k8s-binaries/node.tar.gz {{ .NODE_BINARIES_URL }}
+tar xvf node.tar.gz
 {{- end }}
 
 {{- if and .KUBELET .NODE_BINARIES_URL }}
-sudo install --owner=0 --group=0 --mode=0755 /tmp/k8s-binaries/kubernetes/node/bin/kubelet /usr/bin/kubelet
-sudo rm /tmp/k8s-binaries/kubernetes/node/bin/kubelet
+sudo install --owner=0 --group=0 --mode=0755 /tmp/k8s-binaries/kubernetes/node/bin/kubelet /opt/bin/kubelet
+sudo ln -sf /opt/bin/kubelet /usr/bin/
```
kron4eg, line 252: Do we even need to link kubelet if in systemd unit file it's used like `ExecStart=/opt/bin/kubelet`?

kron4eg (approved):
✅

pkg/scripts/os.go
xmudrii: The reason for doing this is that Kubeadm requires Kubelet to be on PATH. Otherwise, the preflight checks would fail with the following error:

```
[preflight] Some fatal errors occurred:
	[ERROR KubeletVersion]: couldn't get kubelet version: cannot execute 'kubelet --version': executable file not found in $PATH
```

PR #1183, opened 2020-12-07
use a single shared ssh connection for proxying individual tcp connec…

### LINKED ISSUE #1181: kubectl proxy fails silently to tunnel connections when no ssh key is available in the ssh agent
**What happened**: I used `kubeone proxy` without having ssh keys in my ssh-agent and kubeone did not communicate that there is a problem. When I tried to contact the kubernetes api through that proxy kubectl would just hang and then print `Unable to connect to the server: Service Unavailable` and kubeone did not print any error in this stage.

**What is the expected behavior**: Fail with a desriptive error message, that tells the user that the SSH connection could not authenticate properly

**How to reproduce the issue**: Unload keys from ssh-agent, do not have a static path configured in kubeone.yaml and repeat what I stated in `What happened`

KubeOne version (`kubeone version`):  1.0.5
Operating system: 
Provider you're deploying cluster on:

erdii:
…tions

this also has the advantage of failing early when ssh authentication fails

Fixes #1181

```release-note
Use a single shared ssh connection for proxying individual tcp connections
```

### DIFF
--- a/pkg/cmd/proxy.go
+++ b/pkg/cmd/proxy.go
@@ -26,6 +26,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
+	"k8c.io/kubeone/pkg/ssh"
 	"k8c.io/kubeone/pkg/state"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/cmd/proxy.go
```diff
@@ -90,7 +98,9 @@ func setupProxyTunnel(opts *proxyOpts) error {
 
 	fmt.Println("SSH tunnel started, please open another terminal and setup environment")
 	fmt.Printf("export HTTPS_PROXY=http://%s\n", opts.ListenAddr)
-	return server.ListenAndServe()
+
+	err = server.ListenAndServe()
+	return err
```
kron4eg, line 103: The change here is no changing to good anything, please return back the `return server.ListenAndServe()`

kron4eg (commented):
In general looks good, left one microscopic comment to fix.

pkg/cmd/proxy.go
erdii: Nice catch! I somehow thought it would be needed for `defer tunn.Close()` to work. :see_no_evil:

kron4eg (approved):

PR #1190, opened 2020-12-14
Safeguard against sudden and dramatic cluster reconfigurations

### LINKED ISSUE #1185: Run probes to possibly detect container runtime
Run probes to possibly detect container runtime and compare this information to the config. If docker is already running but conrainerd is configured, probably user is trying to blow up the cluster and kubeone should prevent this.

A follow up based on https://github.com/kubermatic/kubeone/pull/1180#issuecomment-739939128

kron4eg:
Fixes #1185 

```release-note
Safeguard against sudden and dramatic cluster reconfigurations
```

### DIFF
--- a/pkg/apis/kubeone/helpers.go
+++ b/pkg/apis/kubeone/helpers.go
@@ -52,9 +52,10 @@ func (c KubeOneCluster) Followers() []HostConfig {
 	return followers
 }
 
-func (c *KubeOneCluster) OwnNode(name string) bool {
+// IsManagedNode reports whether given node name is known to the KubeOne configuration
+func (c *KubeOneCluster) IsManagedNode(nodename string) bool {
 	for _, host := range append(c.ControlPlane.Hosts, c.StaticWorkers.Hosts...) {
-		if host.Hostname == name {
+		if host.Hostname == nodename {
 			return true
 		}
 	}
--- a/pkg/apis/kubeone/helpers.go
+++ b/pkg/apis/kubeone/helpers.go
@@ -52,6 +52,16 @@ func (c KubeOneCluster) Followers() []HostConfig {
 	return followers
 }
 
+func (c *KubeOneCluster) OwnNode(name string) bool {
+	for _, host := range append(c.ControlPlane.Hosts, c.StaticWorkers.Hosts...) {
+		if host.Hostname == name {
+			return true
+		}
+	}
+
+	return false
+}
+
 // SetHostname sets the hostname for the given host
 func (h *HostConfig) SetHostname(hostname string) {
 	h.Hostname = hostname
--- a/pkg/state/cluster.go
+++ b/pkg/state/cluster.go
@@ -37,8 +37,9 @@ type Cluster struct {
 type Host struct {
 	Config *kubeone.HostConfig
 
-	ContainerRuntime ComponentStatus
-	Kubelet          ComponentStatus
+	ContainerRuntimeDocker     ComponentStatus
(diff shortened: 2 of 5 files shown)


pkg/scripts/os.go
```diff
@@ -25,295 +25,6 @@ const (
...
-
-sudo systemctl daemon-reload
-sudo systemctl enable --now kubelet
-
-{{- if or .FORCE .KUBELET }}
-sudo systemctl restart kubelet
-{{- end }}
-`
-
 	kubeadmCoreOSTemplate = `
```
xmudrii, line 28: Can you also move CoreOS/Flatcar templates to a dedicated file for consistency?

kron4eg:
/hold

I figured, the safeguard that iterates over all nodes is too strict since there also might be machine-controller controlled nodes that the container runtimes will be different

pkg/state/cluster.go
```diff
@@ -242,7 +244,17 @@ func (h *Host) WorkerHealthy() bool {
 }
 
 func (h *Host) healthy() bool {
-	return h.IsInCluster && h.ContainerRuntime.Healthy() && h.Kubelet.Healthy()
+	var crStatus bool
+
+	if h.ContainerRuntimeContainerd.IsProvisioned() {
```
xmudrii, line 249: I don't understand this check. How do we know that Docker is installed and supposed to be healthy if we just check is containerd provisioned? Shouldn't we check here is Docker provisioned instead?
```suggestion
	if h.ContainerRuntimeDocker.IsProvisioned() {
```

kron4eg: Yep, you're absolutely right, looks like I've messed up when was moving the code around. Nice catch!

pkg/scripts/os.go
kron4eg: I will actually move this extraction to the dedicated PR, as this move is not strictly related to the safeguard :)

pkg/apis/kubeone/helpers.go
```diff
@@ -52,6 +52,16 @@ func (c KubeOneCluster) Followers() []HostConfig {
 	return followers
 }
 
+func (c *KubeOneCluster) OwnNode(name string) bool {
```
xmudrii, line 55: `OwnNode` is a little bit confusing name to me. I would propose that we rename it to something like `NodeExists`. Also, please add a comment describing what this function does.

pkg/tasks/probes.go
```diff
@@ -37,20 +37,44 @@ import (
...
+	systemdShowExecStartCMD = `systemctl show %s -p ExecStart`
 
-	dockerVersionDPKG   = `dpkg-query --show --showformat='${Version}' docker-ce | cut -d: -f2 | cut -d~ -f1`
-	dockerVersionRPM    = `rpm -qa --queryformat '%{RPMTAG_VERSION}' docker-ce`
-	dockerVersionAmazon = `rpm -qa --queryformat '%{RPMTAG_VERSION}' docker`
+	kubeletInitializedCMD = `test -f /etc/kubernetes/kubelet.conf`
+)
 
-	containerdVersion = `containerd --version | awk '{print $3}'`
+func safeguard(s *state.State) error {
```
xmudrii, line 46: Should we allow users to skip this safeguard with the `--force` flag?

kron4eg: Not now, but maybe in the future when possibly we will figure out the way to migrate between runtimes.

xmudrii (approved):
/lgtm

PR #1192, opened 2020-12-15
containerd is default container runtime for new clusters

### LINKED ISSUE #1187: Containerd defaults
For clusters 1.21+ containerd should be a default (and no docker allowed) container runtime.

Kubernetes 1.21+ is expected to delete docker support from kubelet

kron4eg:
This PR in changes default container runtime from docker to containerd but only for NEW clusters.
Existing docker clusters will continue to use docker.

Fixes #1187 

```release-note
containerd is default container runtime for new clusters
```

### DIFF
--- a/pkg/templates/kubeadm/v1beta2/kubeadm.go
+++ b/pkg/templates/kubeadm/v1beta2/kubeadm.go
@@ -274,6 +274,7 @@ func NewConfigWorker(s *state.State, host kubeoneapi.HostConfig) ([]runtime.Obje
 
 	nodeRegistration := kubeadmv1beta2.NodeRegistrationOptions{
 		Name:      host.Hostname,
+		Taints:    host.Taints,
 		CRISocket: s.ContainerRuntimeConfig().CRISocket(),
 		KubeletExtraArgs: map[string]string{
 			"node-ip":           nodeIP,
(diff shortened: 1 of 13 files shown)


pkg/templates/kubeadm/v1beta2/kubeadm.go
```diff
@@ -272,7 +273,8 @@ func NewConfigWorker(s *state.State, host kubeoneapi.HostConfig) ([]runtime.Obje
 	}
 
 	nodeRegistration := kubeadmv1beta2.NodeRegistrationOptions{
-		Name: host.Hostname,
+		Name:      host.Hostname,
+		CRISocket: s.ContainerRuntimeConfig().CRISocket(),
```
xmudrii, line 277: Why we don't have Taints here as we do in the control plane config?

kron4eg: this is worker config

xmudrii: I know, but the behavior should be:

* If it's a control plane node, default taints to the control plane taint
* If it's a worker node, default taints to no taints

However, in both cases, users should be able to provide custom taints (using the `host.Taints` field).

xmudrii: This is not directly related to this PR, but if it's a bug, we should fix it or create a follow-up ticket.

kron4eg:
I will repurpose the diff of this PR in another PR.

PR #1204, opened 2020-12-17
Containerd on v1.23

kron4eg:
**What this PR does / why we need it**:
Forced use of containerd as a runtime for clusters version >= 1.23 (estimated version of dockershim removal).

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1187

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Containerd will be the only available container runtime on kubernetes >= 1.23
```

PR #1205, opened 2020-12-18
Containerd on kubernetes 1.21+

### LINKED ISSUE #1186: Containerd add E2E tests
At least 1 AWS test should be added (AWS to launch it every time).

A follow up based on https://github.com/kubermatic/kubeone/pull/1180#pullrequestreview-546048297

kron4eg:
Fixes #1186  #1187

Initial test uses upgrade from 1.16 for a reason:
* to check install of old version
* to check upgrade path

```release-note
NONE
```


kron4eg:
/hold 

until #1204 is merged

xmudrii:
Can you refactor this PR so it doesn’t depend on #1204? It might take some time until it’s not merged and I don’t think that we should rush it.

kron4eg:
No, I don't think it makes sense and is not worth the effort. #1204 is logical conclusion of containerd endeavor in KubeOne and so that's why E2E is based on top of it.

kron4eg:
Alternatively, we can close #1204 in favour of this one and close both open issues at once.

xmudrii:
/hold

We should discuss this a bit further and double-check what's the plan for removing dockershim.

pkg/apis/kubeone/helpers.go
```diff
@@ -89,6 +91,30 @@ func (crc ContainerRuntimeConfig) String() string {
 	return "unknown"
 }
 
+func (crc *ContainerRuntimeConfig) UnmarshalText(text []byte) error {
```
xmudrii, line 94: What is the reason for passing the byte slice as a parameter? Can we make it use string instead, especially because the function that uses this helper is using strings?

kron4eg: This is conform the https://pkg.go.dev/encoding#TextUnmarshaler interface.

pkg/scripts/node.go
```diff
@@ -16,16 +16,19 @@ limitations under the License.
...
+	hostnameScript = heredoc.Doc(`
+		fqdn=$(hostname -f)
+		[ "$fqdn" = localhost ] && fqdn=$(hostname)
+		echo "$fqdn"
+	`)
+
+	drainNodeScriptTemplate = heredoc.Doc(`
+		sudo KUBECONFIG=/etc/kubernetes/admin.conf \
+		kubectl drain {{ .NODE_NAME }} --ignore-daemonsets --delete-local-data
+	`)
```
xmudrii, line 31: Can we also use the Kubernetes API for draining the node, like we do for cordoning?

pkg/state/context.go
```diff
@@ -70,6 +71,28 @@ type State struct {
...
+
+	switch {
+	case crCfg.Docker != nil:
+	case crCfg.Containerd != nil:
+	default:
+		crCfg.Docker = &kubeoneapi.ContainerRuntimeDocker{}
+	}
+
+	return crCfg
+}
```
xmudrii, line 94: I don't really like the idea that we put this information in the state. This means that we have out-of-sync information between the state and the Cluster object. If someone would, by mistake, take `s.Cluster.ContainerRuntime`, they would get incorrect information and it would be hard to catch such a bug. It would make more sense to make this something like a defaulting function (it can't be a defaulting function, but we can re-use the mechanism), i.e. make it mutate the cluster object.

Additionally, the logic is not really correct because we don't check if the cluster is provisioned or not. I believe that this check:
```go
if condition.Check(s.LiveCluster.ExpectedVersion)
```
should become:
```go
if !s.LiveCluster.IsProvisioned() && condition.Check(s.LiveCluster.ExpectedVersion)
```

xmudrii (changes requested):

pkg/scripts/discovery.go
```diff
@@ -1,48 +0,0 @@
...
-# Check is Kubeadm installed
-if ! type kubeadm &>/dev/null; then exit 1; fi
-# Check do Kubernetes directories and files exist
-if [[ ! -d "/etc/kubernetes/manifests" ]]; then exit 1; fi
-if [[ ! -d "/etc/kubernetes/pki" ]]; then exit 1; fi
-if [[ ! -f "/etc/kubernetes/kubelet.conf" ]]; then exit 1; fi
-# Check are kubelet running
-if ! sudo systemctl is-active --quiet kubelet &>/dev/null; then exit 1; fi
-`
-)
```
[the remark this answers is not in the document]
kron4eg: No more of those check, the are not needed. This was a hack, and now it's time to remove it. 
Now it detects all the needed information during probes phase which became mandatory for everything.

pkg/scripts/node.go
kron4eg: Yeah, but it requires additional 200 lines of code for no good reason. Someday maybe 🤷‍♂️.

pkg/scripts/render.go
```diff
@@ -112,19 +112,21 @@ var (
...
+				sudo apt-key add -
+			sudo add-apt-repository "deb https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
+			{{ end }}
+
+			{{ if or .FORCE .UPGRADE }}
+			sudo apt-mark unhold containerd.io
+			{{ end }}
+
+			sudo apt-get install -y containerd.io=%s*
+			sudo apt-mark hold containerd.io
```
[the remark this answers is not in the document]
kron4eg: Well, package repo is always better over the direct download. It's checksummed, it's updating. It has dependencies that can change over time (this was the main driver to ditch the direct downloading).

After docker inc was acquired by the Mirantins, docker repositories became much better in terms of OS and versions support. And besides, we still use docker repos for docker itself. We already have the repo, so 🤷‍♂️ just use what we already have.

pkg/state/context.go
kron4eg: So you got me thinking again, and I think maybe we should ERROR OUT in case when the docker is specified (even of default), and plainly refuse until user will be conscious choice to change the config to indicate containerd as a container-runtime. Such approach will remove any double readings and accidents.

xmudrii: I agree, but we should start doing that only from 1.22+ because 1.21 still has Docker support. To recap, the plan should be something along:
1) Default to containerd for **new** 1.21+ clusters
2) Default to containerd for **existing** 1.22+ clusters (this should be done once we have migration in-place, not now)
3) Reject Docker for **all** clusters for Kubernetes 1.22+

kron4eg: OK, done, PTAL again

xmudrii (changes requested):

pkg/state/context.go
xmudrii: This is still not fixed, see https://github.com/kubermatic/kubeone/pull/1205#discussion_r561732844.

pkg/scripts/funcs.go
```diff
@@ -83,10 +83,11 @@ var (
...
 
 const (
-	defaultDockerVersion       = "19.03.14"
-	defaultAmazonDockerVersion = "19.03.13"
-	defaultLegacyDockerVersion = "18.09.9"
-	defaultContainerdVersion   = "1.4.3"
+	defaultDockerVersion           = "19.03.14"
+	defaultAmazonDockerVersion     = "19.03.13"
+	defaultLegacyDockerVersion     = "18.09.9"
+	defaultContainerdVersion       = "1.4.3"
```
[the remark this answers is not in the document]
kron4eg: No, luckily the CRI standard abstracts that away for us. The lowest version of kubernetes we support does support CRI (and even lesser versions). There is no such thing as "kubernetes VS containerd support matrix" as it defies the whole point of CRI.

pkg/apis/kubeone/helpers.go
```diff
@@ -89,6 +91,30 @@ func (crc ContainerRuntimeConfig) String() string {
...
+func (crc ContainerRuntimeConfig) CRISocket() string {
+	switch {
+	case crc.Containerd != nil:
+		return "/run/containerd/containerd.sock"
+	case crc.Docker != nil:
+		return "/var/run/dockershim.sock"
+	}
+
+	return ""
+}
```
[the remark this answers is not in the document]
kron4eg: I was experimenting with and without being explicit abut socket. With having it it works much better. So no, it's needed here.

pkg/apis/kubeone/v1beta1/defaults.go
```diff
@@ -118,13 +120,25 @@ func SetDefaults_Versions(obj *KubeOneCluster) {
...
+
+	k121OrMore, _ := semver.NewConstraint(">=1.21")
+	actualVer, err := semver.NewVersion(obj.Versions.Kubernetes)
+
+	if err == nil {
+		if k121OrMore.Check(actualVer) {
+			defaultContainerRuntime.Containerd = &ContainerRuntimeContainerd{}
+			defaultContainerRuntime.Docker = nil
+		}
+	}
```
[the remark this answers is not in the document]
kron4eg: It seems much better alternative will be:
* always default container-runtime to docker, preserving behaviour
* always require containerd to be explicitly configured as container-runtime
* fail validation for kubernetes 1.22+ if docker is indicated as container-runtime

P.S.
flatcar is working on containerd support at the moment, but it's not yet there.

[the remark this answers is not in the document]
kron4eg: The aforesaid is to avoid any "magic" jumps in logic and behaviour and keep being predictable.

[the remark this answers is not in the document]
xmudrii: The idea behind switching to containerd for the new clusters as of 1.21 is to encourage users to try and test it before we (and Kubernetes) make it mandatory as of 1.22. Of course, users could do it manually, but if we will already do it, the idea was to do it earlier (in this case 1.21).

Also, the idea was that users don't have to change their manifest (keeping backward compatibility), so if a cluster is already provisioned with Docker, we would use Docker while it's supported.

Another case that came to my mind is that if some user would stamp a custom image that has Docker preinstalled, and we default to containerd, then we might/would fail to provision the cluster. The same applies in the other direction, i.e. someone has containerd preinstalled and we try to use Docker.

I think the best scenario would be that we determine the container runtime dynamically like we determine the hostname and the operating systems. Determining container runtime would happen right after those two tasks and it would work in the following way:

* If container runtime is configured in the manifest, skip determining the CR
* Check is there Docker or containerd already installed and default to the installed CR
* If nothing is installed:
  * For Kubernetes 1.21 and newer: use containerd
  * For Kubernetes 1.20 and older: use Docker
* At the end, validate is the configuration supported (we already have that, so nothing to do from that side)

I think that this should be relatively easy to implement and would solve all our problems, while we would still be able to follow the original plan.

pkg/tasks/probes.go
```diff
@@ -103,7 +104,32 @@ func runProbes(s *state.State) error {
...
+	}
+
+	gteKube121Condition, _ := semver.NewConstraint(">= 1.21")
+
+	switch {
+	case gteKube121Condition.Check(s.LiveCluster.ExpectedVersion) && !s.LiveCluster.IsProvisioned():
+		s.Cluster.ContainerRuntime.Containerd = &kubeoneapi.ContainerRuntimeContainerd{}
+	default:
+		s.Cluster.ContainerRuntime.Docker = &kubeoneapi.ContainerRuntimeDocker{}
 	}
```
xmudrii, line 133: This logic is not going to work for newly created 1.21 clusters.

For example:
* User creates 1.21.0 cluster
* The cluster gets provisioned with containerd
* User attempts to upgrade the cluster to 1.21.1+
* This logic would default to Docker because the cluster is provisioned, so `!s.LiveCluster.IsProvisioned()` would return false
* KubeOne would install Docker and eventually break the cluster

xmudrii (changes requested):

pkg/tasks/probes.go
kron4eg: That's only seems like so, because you've missed the [static defaulting](https://github.com/kubermatic/kubeone/pull/1205/files#diff-8e3eafbecc2f7f4dbfd402187a37180eb206b8fe2128ddf3a60bb7fd2717bf07R135-R137)

kron4eg: OK, I've misread this one. Added some live checks

xmudrii (approved):
/lgtm
/approve

Issue #769, opened 2019-12-22
Support for Encryption Providers


containerpope:
**What feature would you like to be added?**
KubeOne should implement support for providing a Kubernetes Encryption provider to ensure, that no Kubernetes secrets are stored in cleartext in etcd.

This can be achieved by setting the --encryption-provider-config flag on the Kubernetes API Server and pointing it to the location of an EncryptionConfiguration yaml file. 

This is an example config from the docs:
```yaml
apiVersion: apiserver.config.k8s.io/v1
kind: EncryptionConfiguration
resources:
  - resources:
    - secrets
    providers:
    - aescbc:
        keys:
        - name: key1
          secret: <BASE 64 ENCODED SECRET>
    - identity: {}
```

Editing the config allows enabling encryption, disabling encryption and rotating keys. No further logic is required beside changing the content of the file.

**How could this be implemented?**
Implementing this would require KubeOne to copy the config to the /etc/kubernetes folder and setting the flag on the api server. If a KubeOne upgrade is triggered, the hashes of the config files could be compared to detect changes or the file is just copied without checking, KubeOne refreshes the config on the masters and restarts the api servers. 

A good approach might be the following structure under the KubeOne feature config:

```yaml
features:
  # Enables the KubernetesEncryptionProvider that ensures, that secrets are only 
  # stored encrypted inside etcd
  secretEncryption:
    enabled: true
    encryptionConfig: "/path/to/config.yaml"
```

This would ensure, that a user can provide the config from outside easily and is able to rotate keys and other stuff, because he has control over the config.

**What are use cases of the feature?**
* All secrets are stored encrypted in etcd
* Users can easily plugin in other KMS systems for secret encryption
* Users can use the key based secret encryption option
* Users have full control over how the secrets are encrypted, when the keys are rotated or which provider is used

**Useful Links:**
* K8s Docs: https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/
* Use KMS for encryption: https://kubernetes.io/docs/tasks/administer-cluster/kms-provider/

moelsayed:
We are already working on this, but with a different perspective. The proposal for the implementation is tracked in #1212.

PR #1213, opened 2021-01-12
Add encryption providers proposal

### LINKED ISSUE #1212: Proposal for adding Encryption Providers support in KubeOne.

moelsayed:
A proposal draft to add Encryption Providers support in KubeOne.
Fixes #1212

```release-note
NONE
```

### DIFF
--- a/docs/proposals/20210112-encryption-roviders.md
+++ b/docs/proposals/20210112-encryption-roviders.md
@@ -18,12 +18,10 @@ KubeOne needs to support this feature natively. Meaning the user should be able 
 * Provide a safe path to enable/disable Encryption Providers.
 * Support atomic(?) rotation for existing keys.
 * Rewriting all secret resources (no just secrets) after enable/disable/rotate operations.
-* Support using custom EncryptionProviders configuration file to allow the use of external KMS.
 
 ## Non-Goals
 
-* Deploy External KMS.
-* Safely manage (disable/enable/rotate) configuration when a custom configuration file is used. 
+* Support external KMS providers.
 
 ## Challenges
 
@@ -31,7 +29,7 @@ The feature has a lot of moving parts; as it requires performing a specific sequ
 
 ## Implementation
 
-Unfortunately, it's not possible to simply update the KubeAPI configuration and expect the configuration to reconcile. KubeOne will have to _read_ the _current_ configuration on the custer, _mutate_ it based on the _required_ state and then apply it. Additionally, KubeOne will have to be able to revert changes on any errors and recover safely if the process is interrupted at any point.
+Unfortunately, it's not possible to simply update the KubeAPI configuration and expect the configuration to reconcile. KubeOne will have to _read_ the _current_ configuration on the cluster, _mutate_ it based on the _required_ state and then apply it. Additionally, KubeOne will have to be able to revert changes on any errors and recover safely if the process is interrupted at any point.
 
 The configuration for this will be added under `features` in the KubeOneCluster spec:
 
@@ -42,36 +40,22 @@ features:
   encryptionProviders:
     # enable/disable 
     enabled: true
-    # set to true to automatically rotate keys on apply command. doesn't work with `customProviderFile` set.
-    rotate: true
-    # setting this value will disable automated rotation.
-    customProviderFile: |
-      apiVersion: apiserver.config.k8s.io/v1
-      kind: EncryptionConfiguration
-      resources:
-      - resources:
-        - secrets
-        providers:
-        - identity: {}
-        - aescbc:
-            keys:
-            - name: key1
-            secret: <BASE 64 ENCODED SECRET>
 ```
 
 ### pre-flight checks
+
  * Cluster is healthy.
  * Current Encryption Providers state/configuration is valid and identical on all control plane nodes.
- * Ensure Custom Encryption Providers configuration file is valid (if any). 
 
 ### Enable Encryption Providers for new cluster
+
 * Generate a valid configuration file with the `identity` provider set last.
 * Sync the configuration file to all Control Plane nodes. 
 * Set the required KubeAPI configuration and deploy KubeAPI.
 
 ### Enable Encryption Providers for existing cluster
 
-* Ensure there is no Encryption Provider Config (manually added by the user, broken previos enable process, etc..) present.
+* Ensure there is no Encryption Provider Config (manually added by the user, broken previous enable process, etc..) present.
 * Generate a valid configuration file with the `identity` provider set last.
 * Sync the configuration file to all Control Plane nodes. 
 * Update and restart KubeAPI on all nodes.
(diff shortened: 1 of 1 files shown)


docs/proposals/20210112-encryption-roviders.md
```diff
@@ -0,0 +1,112 @@
...
+* Rewrite all secrets to ensure they are encrypted with the new key.
+* Mutate the configuration file again to remove the old key.
+* Sync the updated configuration file to all control plane nodes and restart KubeAPI.
+
+### Apply Custom Encryption Provider file
+This use case is useful for users who would like to utilize an external KMS provider or specify resources other than secrets for encryption. In this case, KubeOne will not manage the content of the file, it will only validate it to make sure it's syntactically valid. Additionally, KubeOne will not rewrite the resources in this case. 
+
+* Ensure the configuration file is valid. 
+* Sync the configuration file to all control plane nodes.
+* Restart KubeAPI on all nodes.
```
kron4eg, line 105: Custom Encryption Provider require kubeapi-server to have unix socket mounted inside the kubeapi pod (to let kubeapi communicate over it with KMS). Yet, user has no chance to influence on this, so I'm not really sure if we need this.

Let's not provide possibility to specify custom file. This will give us time to figure out a way to have custom KMS providers without regrets on our API.

docs/proposals/20210112-encryption-roviders.md
```diff
@@ -0,0 +1,112 @@
...
+ * Ensure Custom Encryption Providers configuration file is valid (if any). 
+
+### Enable Encryption Providers for new cluster
+* Generate a valid configuration file with the `identity` provider set last.
+* Sync the configuration file to all Control Plane nodes. 
+* Set the required KubeAPI configuration and deploy KubeAPI.
+
+### Enable Encryption Providers for existing cluster
+
+* Ensure there is no Encryption Provider Config (manually added by the user, broken previos enable process, etc..) present.
```
kron4eg, line 74: How will you ensure this?

docs/proposals/20210112-encryption-roviders.md
moelsayed: Good point :+1:

docs/proposals/20210112-encryption-roviders.md
moelsayed: Probe the server to check if the kubeapi has the option enabled.

xmudrii (approved):
/approve

docs/proposals/20210112-encryption-roviders.md
```diff
@@ -0,0 +1,95 @@
...
+Unfortunately, it's not possible to simply update the KubeAPI configuration and expect the configuration to reconcile. KubeOne will have to _read_ the _current_ configuration on the cluster, _mutate_ it based on the _required_ state and then apply it. Additionally, KubeOne will have to be able to revert changes on any errors and recover safely if the process is interrupted at any point.
+
+The configuration for this will be added under `features` in the KubeOneCluster spec:
+
+```yaml
+apiVersion: kubeone.io/v1beta1
+kind: KubeOneCluster
+features:
+  encryptionProviders:
+    enabled: true
```
kron4eg, line 41: Keeping in mind future API extensions that will be inevitably required, simple boolean flag is not enough. Flags are needed per provider type, and not only flags but possible config options too.

So we are looking at 3 possible providers:
* none
* static (the one for being implemented now)
* kms (the one for being implemented later)

Those should be expressed as additional objects inside `encryptionProviders`, like

```
#### VERY DRAFT, MUCH FANTASY ####
####### only for demo purpose #######
kind: KubeOneCluster
features:
  encryptionProviders:
    - kind: none
    - kind: static
    - kind: kms
      endpoint: unix:///tmp/socketfile.sock
      cachesize: 100
      timeout: 3s
```

Alternatively, we could not reimplement parts of what's already done in upstream and "just" provide the

```
kind: KubeOneCluster
features:
  encryptionConfiguration: |
    apiVersion: apiserver.config.k8s.io/v1
    kind: EncryptionConfiguration
    resources:
    - resources:
      - secrets
      providers:
      - kms:
          name: myKmsPlugin
          endpoint: unix:///tmp/socketfile.sock
          cachesize: 100
          timeout: 3s
      - identity: {}
```

Given that second option, we can actually extract the `endpoint` from KMS providers and automatically mount them to the kube-apiserver pod ( with extra mounts in kubeadm config).

Provided that we can detect changes in `encryptionConfiguration` (compare actual file to encryptionConfiguration`, we can automate and orchestrate kubeapi pod restarts.

kron4eg: Providing ONLY `encryptionConfiguration` we could give all encryption options to the end user, including KMS support in one shot.

xmudrii: I agree with what @kron4eg said, but in that case, I'd leave the `enable` field in the API and implement it like @moelsayed initially proposed. For example:

```yaml
features:
  encryptionProviders:
    # enable/disable 
    enable: true
    # setting this value will disable automated rotation.
    customProviderFile: |
      apiVersion: apiserver.config.k8s.io/v1
      kind: EncryptionConfiguration
      ...
```

The difference between the originally proposed API and this API is that I've removed the `rotate` field. Instead of the `rotate` field, we would keep the CLI flag as it's currently proposed.

* If `enable` is set to `false`
  * Do nothing if the feature has never been enabled
  * Otherwise, disable the feature
* If `enable` is set to `true`:
  * If `customProviderFile` is not set, generate it ourselves and handle key management
  * If `customProviderFile` is set, upload it to the instance. Additionally, if KMS is configured, create the appropriate volumes on the API server pods
  * Configure the API server and everything else as described in the proposal

Note: if `customProviderFile` is set and the CLI flag is set, KubeOne would either show a warning or fail.

In this case, we can quickly implement support for all setups. Technically, users could use Terraform/Ansible to deploy KMS and with such API, they could actually use it in their clusters. I think that it wouldn't require too much additional effort on our side, but we could solve everything (besides deploying KMS, which should be discussed) in a single run.

moelsayed: @kron4eg @xmudrii I updated the proposal accordingly. Please check.

For the initial implementation, I will stop at implementing the the custom config basic functionality. I will create a new issue/PR to address deploying the KMS and creating the required volume mounts for the API server. I am considering that out of scope for the initial implementation.

xmudrii: @moelsayed Can you create an Epic for this?

xmudrii (approved):
/lgtm
/approve
/hold
if @kron4eg wants to take a look as well

docs/proposals/20210112-encryption-roviders.md
moelsayed: @xmudrii there is already an epic to track the proposal and the implementation: https://github.com/kubermatic/kubeone/issues/769

xmudrii: @moelsayed I was thinking about an Epic for addressing the KMS deployment and the volume creation. You can also extend the current Epic. I only see issues relevant to the initial implementation in the Epic, but please correct me if I'm wrong.

kron4eg: What happens if:
* `enable: true`, but no `customProviderFile` is given (e.g. it's empty)?
* `enable: true`, but no `customProviderFile` is given and install/apply is run, and then `customProviderFile` becomes defined?

moelsayed: > enable: true, but no customProviderFile is given (e.g. it's empty)?
We manage everything automatically. A new config file is created and pushed to the nodes. In this case key rotation is supported. 
> enable: true, but no customProviderFile is given and install/apply is run, and then customProviderFile becomes defined?
K1 should push the defined custom file but it will refuse to the rotate the keys if the rotate flag is set.

moelsayed: @xmudrii I updated the epic with an issue to add KMS #1242.

kron4eg (approved):
/lgtm
/approve

PR #1235, opened 2021-02-02
Fix confusing defaults in OpenIDConnect

### LINKED ISSUE #1189: Confusing "default" values
**What happened**:
I tried to configure OIDC for my cluster. For this, I looked at the output of `kubeone config print -f`, which tells me

```yaml
features:
  openidConnect:
    enable: false
    config:
      # The URL of the OpenID issuer, only HTTPS scheme will be accepted. If
      # set, it will be used to verify the OIDC JSON Web Token (JWT).
      issuerUrl: ""
      # The client ID for the OpenID Connect client, must be set if
      # issuer_url is set.
      clientId: "kubernetes"
      # The OpenID claim to use as the user name. Note that claims other than
      # the default ('sub') is not guaranteed to be unique and immutable. This
      # flag is experimental in kubernetes, please see the kubernetes
      # authentication documentation for further details.
      usernameClaim: "sub"
      # If provided, all usernames will be prefixed with this value. If not
      # provided, username claims other than 'email' are prefixed by the issuer
      # URL to avoid clashes. To skip any prefixing, provide the value '-'.
      usernamePrefix: "oidc:"
      # If provided, the name of a custom OpenID Connect claim for specifying
      # user groups. The claim value is expected to be a string or array of
      # strings. This flag is experimental in kubernetes, please see the
      # kubernetes authentication documentation for further details.
      groupsClaim: "groups"
      # If provided, all groups will be prefixed with this value to prevent
      # conflicts with other authentication strategies.
      groupsPrefix: "oidc:"
```

This is what I assumed to be the defaults, so I went ahead and configured this in my `kubeone.yaml`:

```yaml
features:
  openidConnect:
    enable: true
    config:
      issuerUrl: "https://example.com/dex"
      usernameClaim: "preferred_username"
```

But this left me then with only half the configuration applied, and I had to actually configure _all_ values to make things work.

Everything from the kubeone output is actually the default.

See above.

KubeOne version (`kubeone version`): 1.1.0
Operating system: Linux

kron4eg:
Fixes #1189 

```release-note
Fix confusing defaults in OpenIDConnect
```

### DIFF
--- a/pkg/apis/kubeone/v1alpha1/defaults.go
+++ b/pkg/apis/kubeone/v1alpha1/defaults.go
@@ -90,9 +90,7 @@ func SetDefaults_APIEndpoints(obj *KubeOneCluster) {
 		}
 		obj.APIEndpoint.Host = obj.Hosts[0].PublicAddress
 	}
-	if obj.APIEndpoint.Port == 0 {
-		obj.APIEndpoint.Port = 6443
-	}
+	obj.APIEndpoint.Port = defaulti(obj.APIEndpoint.Port, 6443)
 }
 
 func SetDefaults_Versions(obj *KubeOneCluster) {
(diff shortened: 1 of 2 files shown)


xmudrii (approved):
/lgtm

PR #1251, opened 2021-02-21
Fix missing confirmation for reset

shibumi:

This PR adds a request for confirmation for the `reset` subcommand.
We have a request for confirmation already in the `apply` subcommand, hence it should not be much work
adding the same question to `reset`. I really think `reset` should have such a question, because it destroys
the whole cluster.

None

The functionality works so far, but I am not sure about the implementation. Actually it would be nicer to have access to the
`LiveCluster` object inside of `reset`, because the `Cluster` object doesn't seem to be "Dynamic Worker"-aware.
Furthermore it looks like the hostnames are missing when I test it:

```
The following nodes will be resetted: 
        + reset control plane node "" (10.88.10.2)
        + reset control plane node "" (10.88.10.3)
        + reset control plane node "" (10.88.10.4)
        + reset static worker nodes "" (10.88.10.11)
        + reset static worker nodes "" (10.88.10.12)
        + reset static worker nodes "" (10.88.10.13)
Do you want to proceed (yes/no): no

INFO[01:49:46 CET] Operation canceled.     
```

Does this have something to do with missing network access? I tested this without the VPN connection to my cluster (I did not have another cluster to test right now and I did not want to delete one of our company clusters out of a mistake).

```release-note
[BREAKING/ACTION REQUIRED] The kubeone reset command requires an explicit confirmation like the apply command. The command can be automatically approved by using the --auto-approve flag.
```

### DIFF
--- a/pkg/cmd/apply.go
+++ b/pkg/cmd/apply.go
@@ -312,7 +312,7 @@ func runApplyInstall(s *state.State, opts *applyOpts) error { // Print the expec
 	}
 
 	fmt.Println()
-	confirm, err := confirmApply(opts.AutoApprove)
+	confirm, err := confirmCommand(opts.AutoApprove)
 	if err != nil {
 		return err
 	}
@@ -416,7 +416,7 @@ func runApplyUpgradeIfNeeded(s *state.State, opts *applyOpts) error {
 	}
 
 	fmt.Println()
-	confirm, err := confirmApply(opts.AutoApprove)
+	confirm, err := confirmCommand(opts.AutoApprove)
 	if err != nil {
 		return err
 	}
--- a/pkg/cmd/apply.go
+++ b/pkg/cmd/apply.go
@@ -18,6 +18,10 @@ package cmd
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
+	"reflect"
+
 	"github.com/MakeNowJust/heredoc/v2"
 	"github.com/Masterminds/semver/v3"
 	"github.com/pkg/errors"
@@ -26,9 +30,6 @@ import (
 	"k8c.io/kubeone/pkg/credentials"
 	"k8c.io/kubeone/pkg/state"
 	"k8c.io/kubeone/pkg/tasks"
-	"os"
-	"path/filepath"
-	"reflect"
 
 	apiserverconfigv1 "k8s.io/apiserver/pkg/apis/config/v1"
 	kyaml "sigs.k8s.io/yaml"
--- a/pkg/cmd/apply.go
+++ b/pkg/cmd/apply.go
@@ -17,29 +17,22 @@ limitations under the License.
 package cmd
 
 import (
-	"bufio"
 	"fmt"
-	"os"
-	"path/filepath"
-	"reflect"
-	"strings"
-
 	"github.com/MakeNowJust/heredoc/v2"
 	"github.com/Masterminds/semver/v3"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
-	"golang.org/x/term"
-
 	"k8c.io/kubeone/pkg/credentials"
 	"k8c.io/kubeone/pkg/state"
 	"k8c.io/kubeone/pkg/tasks"
+	"os"
+	"path/filepath"
+	"reflect"
 
 	apiserverconfigv1 "k8s.io/apiserver/pkg/apis/config/v1"
 	kyaml "sigs.k8s.io/yaml"
 )
-
-const yes = "yes"
 
 type applyOpts struct {
 	globalOptions
@@ -450,7 +443,7 @@ func runApplyRotateKey(s *state.State, opts *applyOpts) error {
 	}
 
 	fmt.Println()
-	confirm, err := confirmApply(opts.AutoApprove)
+	confirm, err := confirmCommand(opts.AutoApprove)
 	if err != nil {
 		return err
 	}
@@ -460,28 +453,6 @@ func runApplyRotateKey(s *state.State, opts *applyOpts) error {
 		return nil
 	}
 	return errors.Wrap(tasksToRun.Run(s), "failed to reconcile the cluster")
-}
-
-func confirmApply(autoApprove bool) (bool, error) {
-	if autoApprove {
-		return true, nil
-	}
-
-	if !term.IsTerminal(int(os.Stdin.Fd())) || !term.IsTerminal(int(os.Stdout.Fd())) {
-		return false, errors.New("not running in the terminal")
-	}
-
-	reader := bufio.NewReader(os.Stdin)
-	fmt.Print("Do you want to proceed (yes/no): ")
-
-	confirmation, err := reader.ReadString('\n')
-	if err != nil {
-		return false, err
-	}
-
-	fmt.Println()
-
-	return strings.Trim(confirmation, "\n") == yes, nil
 }
 
 func printHostInformation(host state.Host) {
--- a/pkg/cmd/apply.go
+++ b/pkg/cmd/apply.go
@@ -39,6 +39,8 @@ import (
 	kyaml "sigs.k8s.io/yaml"
 )
 
+const yes = "yes"
+
 type applyOpts struct {
 	globalOptions
 	AutoApprove bool `longflag:"auto-approve" shortflag:"y"`
@@ -479,7 +481,7 @@ func confirmApply(autoApprove bool) (bool, error) {
 
 	fmt.Println()
 
-	return strings.Trim(confirmation, "\n") == "yes", nil
+	return strings.Trim(confirmation, "\n") == yes, nil
 }
 
 func printHostInformation(host state.Host) {
@@ -518,7 +520,7 @@ func componentStatusReport(component state.ComponentStatus) {
 
 func boolStr(b bool) string {
 	if b {
-		return "yes"
+		return yes
 	}
 	return "no"
 }
--- a/pkg/cmd/apply.go
+++ b/pkg/cmd/apply.go
@@ -27,6 +27,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+
 	"k8c.io/kubeone/pkg/credentials"
 	"k8c.io/kubeone/pkg/state"
 	"k8c.io/kubeone/pkg/tasks"
--- a/pkg/cmd/reset.go
+++ b/pkg/cmd/reset.go
@@ -20,11 +20,11 @@ import (
 	"fmt"
 
 	"github.com/MakeNowJust/heredoc/v2"
-	"github.com/kubermatic/machine-controller/pkg/machines/v1alpha1"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
+	"github.com/kubermatic/machine-controller/pkg/machines/v1alpha1"
 	"k8c.io/kubeone/pkg/kubeconfig"
 	"k8c.io/kubeone/pkg/state"
 	"k8c.io/kubeone/pkg/tasks"
@@ -120,7 +120,7 @@ func runReset(opts *resetOpts) error {
 			s.Logger.Warnln("Failed to list Machines. Worker nodes will not be deleted. If there are worker nodes in the cluster, you might have to delete them manually.")
 		}
 		for _, machine := range machines.Items {
-			fmt.Printf("\t+ reset machine-controller managed machines %q\n", machine.Name)
+			fmt.Printf("\t- reset machine-controller managed machines %q\n", machine.Name)
 		}
 	} else {
 		s.Logger.Warnln("Failed to list Machines. Worker nodes will not be deleted. If there are worker nodes in the cluster, you might have to delete them manually.")
--- a/pkg/cmd/reset.go
+++ b/pkg/cmd/reset.go
@@ -19,6 +19,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"github.com/kubermatic/machine-controller/pkg/apis/cluster/v1alpha1"
 	"k8c.io/kubeone/pkg/kubeconfig"
 
 	"github.com/MakeNowJust/heredoc/v2"
@@ -28,8 +29,6 @@ import (
 
 	"k8c.io/kubeone/pkg/state"
 	"k8c.io/kubeone/pkg/tasks"
-
-	corev1 "k8s.io/api/core/v1"
 )
 
 type resetOpts struct {
@@ -123,13 +122,13 @@ func runReset(opts *resetOpts) error {
 
 	// Gather information about machine-controller managed nodes
 	ctx := context.Background()
-	nodes := &corev1.NodeList{}
-	if err := s.DynamicClient.List(ctx, nodes); err != nil {
-		return errors.Wrap(err, "unable to list nodes")
+	machines := v1alpha1.MachineList{}
+	if err := s.DynamicClient.List(ctx, &machines); err != nil {
+		s.Logger.Debugln("unable to list machines: ", err.Error())
 	}
 
-	for _, node := range nodes.Items {
-		fmt.Printf("\t+ reset machine-controller managed nodes %q\n", node.Name)
+	for _, machine := range machines.Items {
+		fmt.Printf("\t+ reset machine-controller managed machines %q\n", machine.Name)
 	}
 
 	confirm, err := confirmCommand(opts.AutoApprove)
--- a/pkg/cmd/reset.go
+++ b/pkg/cmd/reset.go
@@ -114,10 +114,6 @@ func runReset(opts *resetOpts) error {
 		fmt.Printf("\t+ reset static worker nodes %q (%s)\n", node.Hostname, node.PrivateAddress)
 	}
 
-	if !s.Cluster.MachineController.Deploy {
-		s.Logger.Info("Skipping deleting workers because machine-controller is disabled in configuration.")
-	}
-
 	if s.DynamicClient == nil {
 		err = kubeconfig.BuildKubernetesClientset(s)
 		if err != nil {
--- a/pkg/cmd/reset.go
+++ b/pkg/cmd/reset.go
@@ -24,10 +24,11 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
-	"github.com/kubermatic/machine-controller/pkg/machines/v1alpha1"
 	"k8c.io/kubeone/pkg/kubeconfig"
 	"k8c.io/kubeone/pkg/state"
 	"k8c.io/kubeone/pkg/tasks"
+
+	"github.com/kubermatic/machine-controller/pkg/machines/v1alpha1"
 )
 
 type resetOpts struct {
--- a/pkg/cmd/reset.go
+++ b/pkg/cmd/reset.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 
 	"github.com/MakeNowJust/heredoc/v2"
+	"github.com/kubermatic/machine-controller/pkg/machines/v1alpha1"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -27,8 +28,6 @@ import (
 	"k8c.io/kubeone/pkg/kubeconfig"
 	"k8c.io/kubeone/pkg/state"
 	"k8c.io/kubeone/pkg/tasks"
-
-	"github.com/kubermatic/machine-controller/pkg/apis/cluster/v1alpha1"
 )
 
 type resetOpts struct {
@@ -113,21 +112,18 @@ func runReset(opts *resetOpts) error {
 		fmt.Printf("\t- reset static worker nodes %q (%s)\n", node.Hostname, node.PrivateAddress)
 	}
 
-	if s.DynamicClient == nil {
-		err = kubeconfig.BuildKubernetesClientset(s)
-		if err != nil {
-			s.Logger.Warnln("Failed to build Kubernetes client set.")
+	err = kubeconfig.BuildKubernetesClientset(s)
+	if err == nil {
+		// Gather information about machine-controller managed nodes
+		machines := v1alpha1.MachineList{}
+		if err := s.DynamicClient.List(s.Context, &machines); err != nil {
+			s.Logger.Warnln("Failed to list Machines. Worker nodes will not be deleted. If there are worker nodes in the cluster, you might have to delete them manually.")
 		}
-	}
-
-	// Gather information about machine-controller managed nodes
-	machines := v1alpha1.MachineList{}
-	if err := s.DynamicClient.List(s.Context, &machines); err != nil {
+		for _, machine := range machines.Items {
+			fmt.Printf("\t+ reset machine-controller managed machines %q\n", machine.Name)
+		}
+	} else {
 		s.Logger.Warnln("Failed to list Machines. Worker nodes will not be deleted. If there are worker nodes in the cluster, you might have to delete them manually.")
-	}
-
-	for _, machine := range machines.Items {
-		fmt.Printf("\t+ reset machine-controller managed machines %q\n", machine.Name)
 	}
 
 	confirm, err := confirmCommand(opts.AutoApprove)
--- a/pkg/cmd/reset.go
+++ b/pkg/cmd/reset.go
@@ -17,6 +17,9 @@ limitations under the License.
 package cmd
 
 import (
+	"context"
+	"fmt"
+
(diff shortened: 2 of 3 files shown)


kron4eg:
This is breaking CLI change. But on the other hand it's precautionary measure. I'm not sure how to proceed.
@xmudrii WDYT?

shibumi:
@kron4eg it is indeed a breaking CLI change, but I really think we should have such functionality :) My solution may not be the best, but it's definitely something the reset sub command should have. I was really surprised when I used `kubeone reset` the first time.

pkg/cmd/reset.go
```diff
@@ -90,5 +99,23 @@ func runReset(opts *resetOpts) error {
 		return errors.Wrap(err, "failed to initialize State")
 	}
 
+	fmt.Println("The following nodes will be resetted: ")
+	for _, node := range s.Cluster.ControlPlane.Hosts {
+		fmt.Printf("\t+ reset control plane node %q (%s)\n", node.Hostname, node.PrivateAddress)
+	}
+	for _, node := range s.Cluster.StaticWorkers.Hosts {
+		fmt.Printf("\t+ reset static worker nodes %q (%s)\n", node.Hostname, node.PrivateAddress)
+	}
```
kron4eg, line 108: Would be also good to list the Nodes that are managed by the machine-controller and will be destroyed before the cluster

xmudrii:
@shibumi Thank you for the PR! We've discussed it internally, and we agree that there should be a confirmation for the `reset` command because it's very destructive. However, this is a breaking CLI change and it should be handle as such. We considered the following two approaches:

1. Create a new command that would replace the `reset` command, and that would have the confirmation. The new command would deprecate the `reset` command, which we would remove after one or two releases
  * The problem is that `reset` very well describes what's done by the command, so it would be hard to come up with a new name
  * It's not that big breaking change that requires a new command
1. Starting with the upcoming 1.2 release (planned for the upcoming weeks), we would show a warning when users run `kubeone reset` that starting with the 1.3 release, the command will require confirmation. We would also announce this in the changelog. In this case, we would merge your PR after creating the release branch for the 1.2 release (that's going to happen in the upcoming weeks)

We've decided to go with the **second approach** (to show a warning) because it seems like the best approach for moving forward.

I'll put this PR on an explicit hold until we don't create a release branch. In meanwhile, I'll leave some comments which you can address.
/hold

If you want, you can create another PR that would add a warning to the `kubeone reset` command to say that the command will ask for a confirmation starting with the 1.3 release. It can say something along:

> This command will require an explicit confirmation starting with the next release

pkg/cmd/reset.go
```diff
@@ -90,5 +99,23 @@ func runReset(opts *resetOpts) error {
...
+
+	confirm, err := confirmApply(opts.AutoApprove)
```
xmudrii, line 110: Nit: since we're now using this function for other commands as well, we should probably rename it to something like `confirmOperation` and move it to a more generic place (e.g. to the [`shared.go` file](https://github.com/kubermatic/kubeone/blob/master/pkg/cmd/shared.go))

pkg/cmd/reset.go
```diff
@@ -17,17 +17,19 @@ limitations under the License.
 package cmd
 
 import (
+	"fmt"
+
 	"github.com/MakeNowJust/heredoc/v2"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
-
```
xmudrii, line 24: This new line should still stay there because we put `k8c.io` imports in a dedicated group.

pkg/cmd/reset.go
shibumi: @kron4eg  do you have a hint on how I would do this? Do I need the `LiveCluster` API for this?

pkg/cmd/reset.go
shibumi: Makes sense! I absolutely agree with you there :)

pkg/cmd/reset.go
shibumi: Sure

shibumi:
>    * It's not that big breaking change that requires a new command
> 2. Starting with the upcoming 1.2 release (planned for the upcoming weeks), we would show a warning when users run `kubeone reset` that starting with the 1.3 release, the command will require confirmation. We would also announce this in the changelog. In this case, we would merge your PR after creating the release branch for the 1.2 release (that's going to happen in the upcoming weeks)
> 
> We've decided to go with the **second approach** (to show a warning) because it seems like the best approach for moving forward.

Full ACK. The second option makes more sense. Would be a shame to deprecate the `reset` command only for a confirmation flag.


> 
> I'll put this PR on an explicit hold until we don't create a release branch. In meanwhile, I'll leave some comments which you can address.

Will have a look on your suggestions after work :) Until tomorrow you should have an updated version.

> /hold
> 
> If you want, you can create another PR that would add a warning to the `kubeone reset` command to say that the command will ask for a confirmation starting with the 1.3 release. It can say something along:
> 
> > This command will require an explicit confirmation starting with the next release

Will have a look on this, too.

xmudrii:
@shibumi You'll also need to update E2E tests to run the `reset` command with the `--auto-approve` flag here: https://github.com/kubermatic/kubeone/blob/168338870a5eb01a4c4d2831ad84ac415ccd479e/test/e2e/kubeone.go#L185-L197

pkg/cmd/reset.go
kron4eg: @shibumi you will need to:
* get the list of nodes
* every MC-managed Node has Machine object reference in its metadata. If it doesn't it means — node is not managed by the machine-controller

https://github.com/kubermatic/kubeone/blob/168338870a5eb01a4c4d2831ad84ac415ccd479e/pkg/templates/machinecontroller/helper.go#L104-L110

pkg/cmd/reset.go
shibumi: This should be solved

pkg/cmd/reset.go
shibumi: @kron4eg do you want to have another look on my current solution? I have the feeling this should work., but I want to be sure it's okay.

shibumi:
@xmudrii I made sure that the tests are running through now..

pkg/cmd/apply.go
```diff
@@ -17,22 +17,17 @@ limitations under the License.
...
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
-	"golang.org/x/term"
-
 	"k8c.io/kubeone/pkg/credentials"
 	"k8c.io/kubeone/pkg/state"
 	"k8c.io/kubeone/pkg/tasks"
+	"os"
+	"path/filepath"
```
xmudrii, line 30: Those two imports should be in the group (along with `fmt`).

pkg/cmd/reset.go
```diff
@@ -17,17 +17,22 @@ limitations under the License.
...
+	"fmt"
+
 	"github.com/MakeNowJust/heredoc/v2"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
 	"k8c.io/kubeone/pkg/state"
 	"k8c.io/kubeone/pkg/tasks"
+	corev1 "k8s.io/api/core/v1"
```
xmudrii, line 30: This should be a dedicated imports group.

xmudrii (commented):
Tiny nits.

shibumi:
@xmudrii  done

shibumi:
rebased and fixed a minor bug:

I removed the `return nil` from this part here:

```go
	if !s.Cluster.MachineController.Deploy {
		s.Logger.Info("Skipping deleting workers because machine-controller is disabled in configuration.")
                return nil
	}
```

With `return nil` reset would exit if the cluster has no machine Controller.

xmudrii:
@shibumi Considering that the KubeOne 1.2 release went out a while ago, I think that we can proceed with this PR. Can you please rebase it so that we can merge it?

pkg/cmd/reset.go
```diff
@@ -100,5 +105,43 @@ func runReset(opts *resetOpts) error {
...
+		fmt.Printf("\t+ reset static worker nodes %q (%s)\n", node.Hostname, node.PrivateAddress)
+	}
+
+	if !s.Cluster.MachineController.Deploy {
+		s.Logger.Info("Skipping deleting workers because machine-controller is disabled in configuration.")
+	}
+
+	if s.DynamicClient == nil {
+		return errors.New("kubernetes client not initialized")
+	}
```
xmudrii, line 122: It's mostly like that this will always be `nil`. If it's `nil`, you should initialize the DynamicClient by calling the following function: https://github.com/kubermatic/kubeone/blob/af4f8c57d1b5d6fabd01c9694422d2347349274f/pkg/kubeconfig/kubeconfig.go#L52-L74

pkg/cmd/reset.go
```diff
@@ -100,5 +105,43 @@ func runReset(opts *resetOpts) error {
...
+	// Gather information about machine-controller managed nodes
+	ctx := context.Background()
+	nodes := &corev1.NodeList{}
+	if err := s.DynamicClient.List(ctx, nodes); err != nil {
+		return errors.Wrap(err, "unable to list nodes")
+	}
+
+	for _, node := range nodes.Items {
+		fmt.Printf("\t+ reset machine-controller managed nodes %q\n", node.Name)
+	}
```
xmudrii, line 133: There are two problems with this approach:

1. This will also list nodes not managed by machine-controller (e.g. static worker nodes)
2. Someone might run `kubeone reset` on instances that are not provisioned yet. In that case, reset will fail, however, that shouldn't happen at all.

The first point can be solved by filtering by labels or by listing Machine objects (instead of Node objects). The second point can be solved by ignoring errors instead of returning them.

pkg/cmd/reset.go
```diff
@@ -100,5 +105,43 @@ func runReset(opts *resetOpts) error {
...
+	for _, node := range s.Cluster.ControlPlane.Hosts {
+		fmt.Printf("\t+ reset control plane node %q (%s)\n", node.Hostname, node.PrivateAddress)
+	}
+	for _, node := range s.Cluster.StaticWorkers.Hosts {
+		fmt.Printf("\t+ reset static worker nodes %q (%s)\n", node.Hostname, node.PrivateAddress)
+	}
+
+	if !s.Cluster.MachineController.Deploy {
+		s.Logger.Info("Skipping deleting workers because machine-controller is disabled in configuration.")
+	}
```
xmudrii, line 118: We already print that message, so I don't think this is needed here: https://github.com/kubermatic/kubeone/blob/af4f8c57d1b5d6fabd01c9694422d2347349274f/pkg/templates/machinecontroller/helper.go#L91-L95

pkg/cmd/reset.go
```diff
@@ -100,5 +105,43 @@ func runReset(opts *resetOpts) error {
 
 	s.Logger.Warnln("this command will require an explicit confirmation starting with the next minor release (v1.3)")
 
+	fmt.Println("The following nodes will be resetted: ")
```
xmudrii, line 108: I think that we might want to explicitly say what this means. For example, we can say something along:

```suggestion
	fmt.Println("The following nodes will be reset. The Kubernetes cluster running on those nodes will be permanently destroyed:")
```

shibumi:
Hey @xmudrii  thansk for your feedback. I will resolve the issues asap and then we can get this merged, I guess :D

pkg/cmd/reset.go
shibumi: done

pkg/cmd/reset.go
```diff
@@ -100,5 +105,43 @@ func runReset(opts *resetOpts) error {
...
+	if err := s.DynamicClient.List(ctx, nodes); err != nil {
+		return errors.Wrap(err, "unable to list nodes")
+	}
+
+	for _, node := range nodes.Items {
+		fmt.Printf("\t+ reset machine-controller managed nodes %q\n", node.Name)
+	}
+
+	confirm, err := confirmCommand(opts.AutoApprove)
+
```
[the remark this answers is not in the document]
shibumi: done

pkg/cmd/reset.go
shibumi: @xmudrii  does this resolve this issue? I wasn't sure if v1alpha1.MachineList{} is here correct :)
https://github.com/kubermatic/kubeone/pull/1251/commits/cc83fc094ccbc6f93c11135832fe715f28a5cb0b

pkg/cmd/reset.go
shibumi: solved in: https://github.com/kubermatic/kubeone/pull/1251/commits/35dfd6ef9da62a0bac09e58db6b7c7aab36e0544

please have a look

pkg/cmd/reset.go
shibumi: @xmudrii  has been something like this in your mind? https://github.com/kubermatic/kubeone/pull/1251/commits/c0022c072e866a06d3acbce021d26478164c0b33

shibumi:
@xmudrii  sorry forgot to highlight you. I think I changed everything according to your feedback.

pkg/cmd/reset.go
```diff
@@ -17,6 +17,11 @@ limitations under the License.
 package cmd
 
 import (
+	"context"
+	"fmt"
+	"github.com/kubermatic/machine-controller/pkg/apis/cluster/v1alpha1"
+	"k8c.io/kubeone/pkg/kubeconfig"
```
xmudrii, line 23: This import should be moved along with other `k8c.io` imports.

pkg/cmd/reset.go
```diff
@@ -17,6 +17,11 @@ limitations under the License.
 package cmd
 
 import (
+	"context"
+	"fmt"
+	"github.com/kubermatic/machine-controller/pkg/apis/cluster/v1alpha1"
```
xmudrii, line 22: This import should also be moved to a dedicated group. You can move it to a new group below `k8c.io` imports, like here:

https://github.com/kubermatic/kubeone/blob/a21344b6b32458cf088efb1d71a1a55c0507b685/pkg/templates/machinecontroller/machines.go#L19-L38

pkg/cmd/reset.go
```diff
@@ -100,5 +105,41 @@ func runReset(opts *resetOpts) error {
...
+		if err != nil {
+			return err
+		}
+	}
+
+	// Gather information about machine-controller managed nodes
+	ctx := context.Background()
+	machines := v1alpha1.MachineList{}
+	if err := s.DynamicClient.List(ctx, &machines); err != nil {
+		s.Logger.Debugln("unable to list machines: ", err.Error())
```
xmudrii, line 127: Let's put a more user-friendly error message here. We can say something along:

```suggestion
		s.Logger.Warnln("Failed to list Machines. Worker nodes will not be deleted. If there are worker nodes in the cluster, you might have to delete them manually.")
```

pkg/cmd/reset.go
```diff
@@ -100,5 +105,41 @@ func runReset(opts *resetOpts) error {
...
+		fmt.Printf("\t+ reset control plane node %q (%s)\n", node.Hostname, node.PrivateAddress)
+	}
+	for _, node := range s.Cluster.StaticWorkers.Hosts {
+		fmt.Printf("\t+ reset static worker nodes %q (%s)\n", node.Hostname, node.PrivateAddress)
+	}
+
+	if s.DynamicClient == nil {
+		err = kubeconfig.BuildKubernetesClientset(s)
+		if err != nil {
+			return err
```
xmudrii, line 119: We shouldn't return an error here. For example, if the cluster is broken or if there's no cluster, it'll make `kubeone reset`  fail, however, it should still work.

Instead, we can show a warning like the one proposed above.

pkg/cmd/reset.go
```diff
@@ -100,5 +105,41 @@ func runReset(opts *resetOpts) error {
...
+		}
+	}
+
+	// Gather information about machine-controller managed nodes
+	ctx := context.Background()
+	machines := v1alpha1.MachineList{}
+	if err := s.DynamicClient.List(ctx, &machines); err != nil {
```
xmudrii, line 126: Let's use `s.Context` instead of `ctx`:

```suggestion
	if err := s.DynamicClient.List(s.Context, &machines); err != nil {
```

pkg/cmd/reset.go
```diff
@@ -100,5 +105,41 @@ func runReset(opts *resetOpts) error {
 
 	s.Logger.Warnln("this command will require an explicit confirmation starting with the next minor release (v1.3)")
 
+	fmt.Println("The following nodes will be reset. The Kubernetes cluster running on those nodes will be permanently destroyed ")
+	for _, node := range s.Cluster.ControlPlane.Hosts {
+		fmt.Printf("\t+ reset control plane node %q (%s)\n", node.Hostname, node.PrivateAddress)
+	}
+	for _, node := range s.Cluster.StaticWorkers.Hosts {
+		fmt.Printf("\t+ reset static worker nodes %q (%s)\n", node.Hostname, node.PrivateAddress)
+	}
```
xmudrii, line 114: Since we're resetting the cluster, I think it makes sense to show the minus sign instead of the plus sign.

```suggestion
	for _, node := range s.Cluster.ControlPlane.Hosts {
		fmt.Printf("\t- reset control plane node %q (%s)\n", node.Hostname, node.PrivateAddress)
	}
	for _, node := range s.Cluster.StaticWorkers.Hosts {
		fmt.Printf("\t- reset static worker nodes %q (%s)\n", node.Hostname, node.PrivateAddress)
	}
```

pkg/cmd/reset.go
```diff
@@ -100,5 +105,41 @@ func runReset(opts *resetOpts) error {
...
+
+	// Gather information about machine-controller managed nodes
+	ctx := context.Background()
+	machines := v1alpha1.MachineList{}
+	if err := s.DynamicClient.List(ctx, &machines); err != nil {
+		s.Logger.Debugln("unable to list machines: ", err.Error())
+	}
+
+	for _, machine := range machines.Items {
+		fmt.Printf("\t+ reset machine-controller managed machines %q\n", machine.Name)
```
xmudrii, line 131: Similar as above, since we're removing the Machine, we can show the minus sign instead.

```suggestion
		fmt.Printf("\t- reset machine-controller managed machines %q\n", machine.Name)
```

xmudrii (commented):
Just a few more nits.

shibumi:
@xmudrii applied your nitpicks :) hit me if you need more adjustments

pkg/cmd/reset.go
```diff
@@ -100,5 +105,40 @@ func runReset(opts *resetOpts) error {
...
+		fmt.Printf("\t- reset control plane node %q (%s)\n", node.Hostname, node.PrivateAddress)
+	}
+	for _, node := range s.Cluster.StaticWorkers.Hosts {
+		fmt.Printf("\t- reset static worker nodes %q (%s)\n", node.Hostname, node.PrivateAddress)
+	}
+
+	if s.DynamicClient == nil {
+		err = kubeconfig.BuildKubernetesClientset(s)
+		if err != nil {
+			s.Logger.Warnln("Failed to build Kubernetes client set.")
```
kron4eg, line 119:
```suggestion
			return err
```

kron4eg: Seem like there was previous discussion about this exact line and change. So please ignore mine message.

pkg/cmd/reset.go
```diff
@@ -100,5 +105,40 @@ func runReset(opts *resetOpts) error {
...
+
+	// Gather information about machine-controller managed nodes
+	machines := v1alpha1.MachineList{}
+	if err := s.DynamicClient.List(s.Context, &machines); err != nil {
+		s.Logger.Warnln("Failed to list Machines. Worker nodes will not be deleted. If there are worker nodes in the cluster, you might have to delete them manually.")
+	}
+
+	for _, machine := range machines.Items {
+		fmt.Printf("\t+ reset machine-controller managed machines %q\n", machine.Name)
+	}
```
xmudrii, line 131: If we fail to build the Kubernetes clientset on L117 and try to list machines on L125, the List function might throw an NPE. That's not something that we want, and we can do something along this to solve it:

1. Try to build the clientset
2. `if err == nil` -> list machines
3. `if err != nil` -> show a warning saying that we failed to list Machines

```suggestion
	err = kubeconfig.BuildKubernetesClientset(s)
	if err == nil {
		// Gather information about machine-controller managed nodes
		machines := v1alpha1.MachineList{}
		if err := s.DynamicClient.List(s.Context, &machines); err != nil {
			s.Logger.Warnln("Failed to list Machines. Worker nodes will not be deleted. If there are worker nodes in the cluster, you might have to delete them manually.")
		}

		for _, machine := range machines.Items {
			fmt.Printf("\t+ reset machine-controller managed machines %q\n", machine.Name)
		}
	} else {
		s.Logger.Warnln("Failed to list Machines. Worker nodes will not be deleted. If there are worker nodes in the cluster, you might have to delete them manually.")
	}
```

shibumi: fixed in: https://github.com/kubermatic/kubeone/pull/1251/commits/90cccce01661e7def9b492ba2361bd416baba4d3

pkg/cmd/shared.go
```diff
@@ -17,10 +17,15 @@ limitations under the License.
...
 import (
+	"bufio"
 	"context"
+	"fmt"
+	"os"
 	"reflect"
 	"strings"
 
+	"golang.org/x/term"
+
```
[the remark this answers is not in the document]
shibumi: ah sorry, you meant the import order. Of course. I applied a patch for this as well :)

shibumi:
Mh ok the linting fails, because of two reasons:

```
pkg/cmd/apply.go:482:45: string `yes` has 3 occurrences, make it a constant (goconst)
	return strings.Trim(confirmation, "\n") == "yes", nil
	                                           ^
pkg/cmd/reset.go:119:6: shadow: declaration of "err" shadows declaration at line 100 (govet)
		if err := s.DynamicClient.List(s.Context, &machines); err != nil {
```

The latter maybe makese sense to change, but the first.. what do you think? @xmudrii  @kron4eg

kron4eg:
@shibumi yeah, please fix them both (add yes constant, replace := to = in `err :=`)

shibumi:
@kron4eg  done. Waiting for the ci now.

xmudrii (commented):
Some final nits, but otherwise looks good to me.

shibumi:
@xmudrii  done

pkg/cmd/apply.go
```diff
@@ -17,23 +17,18 @@ limitations under the License.
...
+	"reflect"
```
xmudrii, line 31: Please move those imports to the first group along with the `fmt` import.

pkg/cmd/reset.go
```diff
@@ -17,11 +17,15 @@ limitations under the License.
...
 
 import (
+	"fmt"
+
 	"github.com/MakeNowJust/heredoc/v2"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
+	"github.com/kubermatic/machine-controller/pkg/machines/v1alpha1"
```
xmudrii, line 27: This import should go into a dedicated group below `k8c.io` imports. For example: https://github.com/kubermatic/kubeone/blob/0328596b6317603087b50ea8777816c0707bd039/pkg/templates/machinecontroller/machines.go#L19-L38

shibumi:
@xmudrii  done. Shall I squash commits or is the commit history fine? My feeling is it's a little bit too atomic.

xmudrii:
@shibumi We automatically squash all PRs on merge.

/test all

shibumi:
@xmudrii  what does this mean? 

pkg/cmd/apply.go:29: File is not `goimports`-ed with -local k8c.io/kubeone (goimports)
	"github.com/spf13/pflag"

pkg/cmd/apply.go
```diff
@@ -17,20 +17,16 @@ limitations under the License.
...
 	"os"
 	"path/filepath"
 	"reflect"
-	"strings"
 
 	"github.com/MakeNowJust/heredoc/v2"
 	"github.com/Masterminds/semver/v3"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
```
xmudrii, line 29: There should be a new line between those two import groups.

```suggestion
	"github.com/spf13/pflag"

```

xmudrii: For some reason, GitHub is not rendering this properly. The new line should be between the `github.com` and `k8c.io` import groups (lines 29-30).

xmudrii (approved):
@shibumi Thank you for the contribution!

/lgtm
/approve

@kron4eg PTAL and remove the hold.

shibumi:
@xmudrii  /retest please :) it was a timeout of your Ci system.

shibumi:
The merge label is blocking the merge

PR #1252, opened 2021-02-22
warn of explicit confirmation for reset

shibumi:

KubeOne 1.3 will have a new confirmation flag for the `reset` subcommand, hence we should warn users about this.

More information can be found in the corresponding PR: https://github.com/kubermatic/kubeone/pull/1251

```release-note
Warn of explicit confirmation for kubeone reset in the future
```

### DIFF
--- a/pkg/cmd/reset.go
+++ b/pkg/cmd/reset.go
@@ -28,6 +28,7 @@ import (
 
 type resetOpts struct {
 	globalOptions
+	AutoApprove    bool `longflag:"auto-approve" shortflag:"y"`
 	DestroyWorkers bool `longflag:"destroy-workers"`
 	RemoveBinaries bool `longflag:"remove-binaries"`
 }
@@ -68,6 +69,13 @@ func resetCmd(rootFlags *pflag.FlagSet) *cobra.Command {
 		},
 	}
 
+	cmd.Flags().BoolVarP(
+		&opts.AutoApprove,
+		longFlagName(opts, "AutoApprove"),
+		shortFlagName(opts, "AutoApprove"),
+		false,
+		"auto approve reset (NO-OP/NOT YET ENABLED)")
+
 	cmd.Flags().BoolVar(
 		&opts.DestroyWorkers,
 		longFlagName(opts, "DestroyWorkers"),
@@ -89,6 +97,8 @@ func runReset(opts *resetOpts) error {
 	if err != nil {
 		return errors.Wrap(err, "failed to initialize State")
 	}
+
+	s.Logger.Warnln("this command will require an explicit confirmation starting with the next minor release (v1.3)")
 
 	return errors.Wrap(tasks.WithReset(nil).Run(s), "failed to reset the cluster")
 }
--- a/test/e2e/kubeone.go
+++ b/test/e2e/kubeone.go
@@ -186,6 +186,7 @@ func (k1 *Kubeone) Kubeconfig() ([]byte, error) {
 func (k1 *Kubeone) Reset() error {
 	err := k1.run("reset",
 		"-v",
+		"--auto-approve",
 		"--tfjson", "tf.json",
 		"--destroy-workers",
 		"--manifest", k1.ConfigurationFilePath)


shibumi:
@xmudrii  is this warning sufficient? I am not sure if I should have used the logger or used a specific formatting.

kron4eg:
There is a linker-time assigned variable `k8c.io/kubeone/pkg/cmd.version`
```
$ make
go version go1.16 darwin/amd64
go install -ldflags='-s -w -extldflags=-zrelro -extldflags=-znow -X k8c.io/kubeone/pkg/cmd.version=v1.2.0-beta.1-1-g1683388 -X k8c.io/kubeone/pkg/cmd.commit=168338870a5eb01a4c4d2831ad84ac415ccd479e -X k8c.io/kubeone/pkg/cmd.date=2021-02-23T14:27:11+02:00' -v .
```

So it's possible to use it and check if current version is less than v1.3, and then display the warning.

shibumi:
@kron4eg  nice idea. WIll have a look on this after my work :)

shibumi:
@kron4eg  done. We need to use semver.Compare for our comparison instead of semver.Constraints, because [constraints do not seem to support pre-release versions](https://github.com/Masterminds/semver#user-content-working-with-prerelease-versions:~:text=SemVer%20comparisons%20using%20constraints%20without%20a,which%20is%20what%20the%20spec%20specifies.) and I wanted to make sure that users will get a warning even if they are using v1.2.0-beta.

kron4eg:
@shibumi I must ask to excuse me for confusing you. We had an internal discussion and reconsidered if this condition is a good idea. 

So the plan would be:
* remove the version based condition
* add a `-auto-approve / -y` NOOP flag to reset subcommand (but please make it clean in the flag description it's a NOOP until 1.3)

pkg/cmd/reset.go
```diff
@@ -85,6 +88,21 @@ func resetCmd(rootFlags *pflag.FlagSet) *cobra.Command {
...
+	c, err := semver.NewVersion("1.3")
+	if err != nil {
+		return err
+	}
+	v, err := semver.NewVersion(version)
+	if err != nil {
+		return err
+	}
+	if result := c.Compare(v); result > 0 {
+		fmt.Println("WARNING: This command will require an explicit confirmation starting with KubeOne version 1.3")
```
xmudrii, line 103: I have two comments regarding this line:
1) I think that we should use the logger instead of `fmt.Println`. You should be able to use the logger such as `s.Logger.Warnln` after running `opts.BuildState`
2) Let's change the message to make it clear that 1.3 is the next minor version
```suggestion
		fmt.Println("This command will require an explicit confirmation starting with the next minor release (v1.3)")
```

shibumi: @xmudrii  I am a little bit confused now, because @kron4eg  says I should remove the version check entirely and instead add an auto-approve NOOP flag to the reset command.

Can one of you tell me what I shall do now? :D

kron4eg: I mean, the message about 1.3 should be in place, it's just the check that should be removed. Plus a NOOP flag (to show it off in the --help)

xmudrii: @shibumi We should do both. There should be a no-op flag, but also a warning that we would show unconditionally (without version check). The warning should say that this command will require an explicit confirmation and that we added a no-op flag that will be used as of the next version to skip the confirmation.

shibumi: @kron4eg  @xmudrii  ah ok, now I got it. Sorry for the confusion folks :)

shibumi:
@xmudrii  @kron4eg  I have modified the PR according to your suggestions:

* we make use of the logger now
* we do not calculate the version anymore
* We provide the flag already as no-op
* we check for the flag existence in our tests

kron4eg (approved):

kron4eg:
@xmudrii PTAL

xmudrii (approved):
/lgtm
/approve
/test all

PR #1300, opened 2021-03-30
Automate certificates renew

### LINKED ISSUE #1297: kubeone upgrade --force without changing version doesn't renew certificates
**What happened**:
Despite we run
```
sudo kubeadm upgrade apply -y --certificate-renewal=true 1.17.3 --config=./kubeone/cfg/master_0.yaml
```
on the node, it doesn't actually renew soon-to-be-expired certificates. This is by [design of kubeadm](https://github.com/kubernetes/kubeadm/issues/2308#issuecomment-700185286).

Certificates to be renewed in case when they will soon expire.

KubeOne version (`kubeone version`): any
Operating system: any
Provider you're deploying cluster on: any
Operating system you're deploying on: any

kron4eg:
Fixes #1297 

```release-note
certificates renew on upgrades
```

### DIFF
--- a/pkg/state/cluster.go
+++ b/pkg/state/cluster.go
@@ -78,7 +78,7 @@ const (
 */
 
 const (
-	x90Days = time.Minute * 24 * 90
+	x90Days = time.Hour * 24 * 90
 )
 
 // CertsToExpireInLessThen90Days will return true if any of the control plane certificates are to be expired soon (90
--- a/pkg/state/cluster.go
+++ b/pkg/state/cluster.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2020 The KubeOne Authors.
+Copyright 2021 The KubeOne Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -19,6 +19,7 @@ package state
 import (
 	"errors"
 	"sync"
+	"time"
 
 	"github.com/Masterminds/semver/v3"
 	"k8s.io/client-go/rest"
@@ -44,6 +45,8 @@ type Host struct {
 	// Applicable only for CP nodes
 	APIServer ContainerStatus
 	Etcd      ContainerStatus
+
+	EarliestCertExpiry time.Time
 
 	IsInCluster bool
 	Kubeconfig  []byte
@@ -73,6 +76,27 @@ const (
 /*
 	Cluster level checks
 */
+
+const (
+	x90Days = time.Minute * 24 * 90
+)
+
+// CertsToExpireInLessThen90Days will return true if any of the control plane certificates are to be expired soon (90
+// days).
+func (c *Cluster) CertsToExpireInLessThen90Days() bool {
+	var (
+		now       = time.Now()
+		needRenew bool
+	)
+
+	for _, host := range c.ControlPlane {
+		if !host.EarliestCertExpiry.IsZero() && host.EarliestCertExpiry.Sub(now) <= x90Days {
+			needRenew = true
+		}
+	}
+
+	return needRenew
+}
 
 // IsProvisioned returns is the target cluster provisioned.
 // The cluster is consider provisioned if there is at least one initialized host
--- a/pkg/tasks/certs.go
+++ b/pkg/tasks/certs.go
@@ -46,10 +46,11 @@ func deployCAOnNode(s *state.State, node *kubeoneapi.HostConfig, conn ssh.Connec
 func renewCerts(s *state.State) error {
 	if !s.ForceUpgrade {
 		s.Logger.Warn("Your control-plane certificates are about to expire in less then 90 days")
+		s.Logger.Warn("To renew them without changing kubernetes version run `kubeone apply --force-upgrade`")
 		return nil
 	}
 	s.Logger.Warn("Your control-plane certificates are about to expire in less then 90 days")
-	s.Logger.Warn("Forced renew")
+	s.Logger.Warn("Force renewing Kubernetes certificates")
 
 	// /etc/kubernetes/admin.conf will be changed after certificates renew, so we have to initialize client again
 	s.Logger.Infoln("Resetting Kubernetes clientset...")
--- a/pkg/tasks/certs.go
+++ b/pkg/tasks/certs.go
@@ -17,7 +17,18 @@ limitations under the License.
 package tasks
 
 import (
+	"bytes"
+	"crypto/x509"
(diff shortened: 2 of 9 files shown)


pkg/state/cluster.go
```diff
@@ -74,6 +77,27 @@ const (
 	Cluster level checks
 */
 
+const (
+	x90Days = time.Minute * 24 * 90
```
xmudrii, line 81: You're missing `* 60` for the number of minutes in one hour. You can do something like this instead:

```suggestion
	x90Days = time.Hour * 24 * 90
```

pkg/tasks/certs.go
```diff
@@ -31,3 +42,99 @@ func deployCAOnNode(s *state.State, node *kubeoneapi.HostConfig, conn ssh.Connec
 	s.Logger.Infoln("Uploading PKI files...")
 	return s.Configuration.UploadTo(conn, s.WorkDir)
 }
+
+func renewCerts(s *state.State) error {
+	if !s.ForceUpgrade {
+		s.Logger.Warn("Your control-plane certificates are about to expire in less then 90 days")
```
xmudrii, line 48: Let's mention that you can use `--force-upgrade` to force renew certificates. That can be on a new line, i.e. another `s.Logger.Warn` call.

pkg/tasks/certs.go
```diff
@@ -31,3 +42,99 @@ func deployCAOnNode(s *state.State, node *kubeoneapi.HostConfig, conn ssh.Connec
...
+		return nil
+	}
+	s.Logger.Warn("Your control-plane certificates are about to expire in less then 90 days")
+	s.Logger.Warn("Forced renew")
```
xmudrii, line 52: Let's make this message more explicit, such as:
```suggestion
	s.Logger.Warn("Force renewing Kubernetes certificates")
```

pkg/state/cluster.go
kron4eg: ups :D good you’re paying attention :) 👍
done

pkg/tasks/certs.go
kron4eg: done

pkg/tasks/certs.go
kron4eg: done

pkg/tasks/tasks.go
```diff
@@ -116,20 +116,36 @@ func WithFullInstall(t Tasks) Tasks {
...
+func WithResources(t Tasks) Tasks {
 	return t.append(
 		Tasks{
+			{
+				Fn:         renewCerts,
+				ErrMsg:     "failed to renew certificates",
+				Desciption: "renew all certificates",
+				Predicate: func(s *state.State) bool {
+					return s.LiveCluster.CertsToExpireInLessThen90Days()
+				},
```
xmudrii, line 134: If I'm not mistaken, this function uses the cluster state that was generated at the beginning. If a user runs `kubeone apply -y` with changing the Kubernetes version, and with certificates that will expire soon, we can end up in the following situation:

1. Apply probes determine that the certificates will expire soon
2. kubeadm upgrades the cluster and renews all certificates
3. Since we're not refreshing the state after the upgrade, this predicate will be true as certificates were about to expire
4. KubeOne will show a message that the user should run `kubeone apply --force-upgrade`, however, certificates are already renewed by KubeOne when upgrading the cluster
    - This can be misleading to the user and the user will probably think that they need to run force upgrade, while that's not the case

This can also be a problem if the user runs `kubeone apply --force-upgrade` with changing the Kubernetes version. In that case, the certificates will be renewed two times.

pkg/tasks/tasks.go
```diff
@@ -116,20 +116,36 @@ func WithFullInstall(t Tasks) Tasks {
...
 			Task{Fn: createMachineDeployments, ErrMsg: "failed to create worker machines"},
 		)
 }
 
-func WithRefreshResources(t Tasks) Tasks {
+func WithResources(t Tasks) Tasks {
 	return t.append(
 		Tasks{
+			{
+				Fn:         renewCerts,
```
xmudrii, line 129: The kubeconfig will be invalidated if the certificates are renewed, so we should download the new kubeconfig as well.

pkg/tasks/certs.go
```diff
@@ -31,3 +42,100 @@ func deployCAOnNode(s *state.State, node *kubeoneapi.HostConfig, conn ssh.Connec
...
+		renewCmd = "sudo kubeadm certs renew all"
+	}
+
+	err := s.RunTaskOnControlPlane(
+		func(s *state.State, node *kubeoneapi.HostConfig, conn ssh.Connection) error {
+			_, _, _, err := conn.Exec(renewCmd)
+			return err
+		},
+		state.RunParallel,
+	)
```
xmudrii, line 71: What about certificates on the static worker nodes? Do we maybe need to renew those as well?

xmudrii (commented):
We might be missing one important step here. What about restarting the API server, controller manager, etcd, and other relevant components? AFAIK, if we don't restart those components, they will still use the old certificates.

pkg/tasks/certs.go
kron4eg: There are no certificates to renew with kubeadm on worker nodes.
At the moment we have kubelet server self-signed cert (which is already "not-valid"). Plus kubelet-client cert that is rotated by kubelet itself. Everything else works based on tokens.

pkg/tasks/tasks.go
kron4eg: Running --force-upgrade with changed version is so pointless and marginal that I'd say that it doesn't matter that certs will be changed twice.

pkg/tasks/tasks.go
kron4eg: Now we are refreshing downloaded kubeconfig every time.

pkg/tasks/probes.go
```diff
@@ -154,69 +154,76 @@ func kubeletVersionCmdGenerator(execPath string) string {
...
 	if err != nil {
 		return err
 	}
 
-	if err := detectKubeletInitialized(h, conn); err != nil {
+	if err = detectKubeletInitialized(foundHost, conn); err != nil {
 		return err
 	}
 
+	if foundHost.Initialized() && controlPlane {
```
kron4eg, line 215: here's actual bugfix missed: the `&& controlPlane` prevents for checking certificates on worker nodes where the certs are missing all the time.

xmudrii (approved):
/lgtm
/approve

PR #1304, opened 2021-04-02
 Disable repo_gpgcheck for the Kubernetes yum repo

xmudrii:

Installing or upgrading Kubernetes packages on CentOS 7 fails with the following error:

```
Retrieving key from https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg
kubernetes/x86_64/signature                                                                                                                                     | 1.4 kB  00:00:01 !!! 
https://packages.cloud.google.com/yum/repos/kubernetes-el7-x86_64/repodata/repomd.xml: [Errno -1] repomd.xml signature could not be verified for kubernetes
Trying other mirror.
```

As per [Google Known Issues doc](https://cloud.google.com/compute/docs/troubleshooting/known-issues#keyexpired), this issue is affecting CentOS 7 and RHEL. The recommended solution is to disable the `repo_gpgcheck` option. From the docs:

> Yum repositories do not usually use GPG keys for repository validation. Instead, the https endpoint is trusted.

CentOS 8 is not affected, however, this PR disables the `repo_gpgcheck` option for CentOS 8 as well. The reason for doing this is that we're using the same scripts for CentOS 7 and 8, so it's easier to implement the fix this way.

Reference: https://github.com/kubernetes/release/issues/1982#issuecomment-812449935

I've tested this fix manually and confirmed that it works.

```release-note
Disable repo_gpgcheck for the Kubernetes yum repository. This fixes the cluster provisioning and upgrading failures for CentOS/RHEL caused by yum failing to install Kubernetes packages.
```

/assign @kron4eg

### DIFF
--- a/pkg/scripts/os.go
+++ b/pkg/scripts/os.go
@@ -135,7 +135,7 @@ name=Kubernetes
 baseurl=https://packages.cloud.google.com/yum/repos/kubernetes-el7-x86_64
 enabled=1
 gpgcheck=1
-repo_gpgcheck=1
+repo_gpgcheck=0
 gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg
 EOF
 {{ end }}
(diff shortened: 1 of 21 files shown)


kron4eg (approved):
🔥 💩

kron4eg:
The more and more I think we should avoid all those pesky repositories as they are loosing least usefulness.

PR #1306, opened 2021-04-05
[Azure] Added 30 second delay in output so that publicIP for Azure VM is available

dharapvj:
Also added variable for machine counts for master and worker nodes

Currently, for Azure, we must always do `terraform refresh` to get public IP in the output before we can run `kubeone apply`. This step is not properly documented and is unnecessary.  This PR adds a 30 second delay in finishing the output of terraform and thereby ensures that terraform output _ALWAYS_ contains public ip for control plane VMs.

PR also adds variables for count of control plane and worker nodes for easy control of these from variable instead of controlling via terraform files.

NA

NA

```release-note
NONE
```

### DIFF
--- a/examples/terraform/azure/main.tf
+++ b/examples/terraform/azure/main.tf
@@ -130,8 +130,8 @@ resource "azurerm_lb" "lb" {
 }
 
 resource "azurerm_lb_backend_address_pool" "backend_pool" {
-  loadbalancer_id     = azurerm_lb.lb.id
-  name                = "ApiServers"
+  loadbalancer_id = azurerm_lb.lb.id
+  name            = "ApiServers"
 }
 
 resource "azurerm_lb_rule" "lb_rule" {
@@ -228,3 +228,17 @@ resource "azurerm_virtual_machine" "control_plane" {
   }
 }
 
+# Hack to ensure we get access to public ip in first attempt
+resource "time_sleep" "wait_30_seconds" {
+  depends_on      = [azurerm_virtual_machine.control_plane]
+  create_duration = "30s"
+}
+
+data "azurerm_public_ip" "control_plane" {
+  depends_on = [
+    time_sleep.wait_30_seconds
+  ]
+  count               = var.control_plane_vm_count
+  name                = "${var.cluster_name}-cp-${count.index}"
+  resource_group_name = azurerm_resource_group.rg.name
+}
--- a/examples/terraform/azure/main.tf
+++ b/examples/terraform/azure/main.tf
@@ -18,6 +18,9 @@ provider "azurerm" {
   features {}
 }
 
+provider "time" {
+}
+
 resource "azurerm_resource_group" "rg" {
   name     = "${var.cluster_name}-rg"
   location = var.location
@@ -97,7 +100,7 @@ resource "azurerm_public_ip" "lbip" {
 }
 
 resource "azurerm_public_ip" "control_plane" {
-  count = 3
+  count = var.control_plane_vm_count
 
   name                = "${var.cluster_name}-cp-${count.index}"
   location            = var.location
@@ -127,7 +130,6 @@ resource "azurerm_lb" "lb" {
 }
 
 resource "azurerm_lb_backend_address_pool" "backend_pool" {
-  resource_group_name = azurerm_resource_group.rg.name
   loadbalancer_id     = azurerm_lb.lb.id
   name                = "ApiServers"
 }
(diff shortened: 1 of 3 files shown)


examples/terraform/azure/variables.tf
```diff
@@ -69,3 +69,12 @@ variable "worker_vm_size" {
   default     = "Standard_B2s"
 }
 
+variable "control_plane_vm_count" {
+  description = "VM Size for worker machines"
+  default     = 3
+}
+
+variable "worker_node_vm_count" {
```
xmudrii, line 77: This field is called `initial_machinedeployment_replicas` for AWS. I think that we should keep it consistent for all providers/examples.

```suggestion
variable "initial_machinedeployment_replicas" {
```

examples/terraform/azure/variables.tf
```diff
@@ -69,3 +69,12 @@ variable "worker_vm_size" {
...
+  description = "VM Size for worker machines"
+  default     = 3
```
xmudrii, line 79: The default value for other providers is 1, so I think that we should also keep it here for consistency.

examples/terraform/azure/variables.tf
```diff
@@ -69,3 +69,12 @@ variable "worker_vm_size" {
   default     = "Standard_B2s"
 }
 
+variable "control_plane_vm_count" {
+  description = "VM Size for worker machines"
```
xmudrii, line 73: We should adjust the description to make it clear that it's about the number of virtual machines and not about the instance size/type. For example:

```suggestion
  description = "Number of control plane instances"
```

examples/terraform/azure/variables.tf
```diff
@@ -69,3 +69,12 @@ variable "worker_vm_size" {
...
+  default     = 3
+}
+
+variable "worker_node_vm_count" {
+  description = "VM Size for worker machines"
```
xmudrii, line 78: Similar as other comment:

```suggestion
  description = "number of replicas per MachineDeployment"
```

examples/terraform/azure/output.tf
```diff
@@ -22,6 +22,21 @@ output "kubeone_api" {
...
+}
+
+data "azurerm_public_ip" "control_plane" {
+  depends_on = [
+    time_sleep.wait_30_seconds
+  ]
+  count = var.control_plane_vm_count
+  name                = "${var.cluster_name}-cp-${count.index}"
+  resource_group_name = azurerm_resource_group.rg.name
+}
```
xmudrii, line 38: Can you run [`terraform fmt`](https://www.terraform.io/docs/cli/commands/fmt.html) to format the file?

examples/terraform/azure/main.tf
```diff
@@ -127,7 +130,6 @@ resource "azurerm_lb" "lb" {
 }
 
 resource "azurerm_lb_backend_address_pool" "backend_pool" {
-  resource_group_name = azurerm_resource_group.rg.name
```
xmudrii, line 130: Why is it needed to remove this?

examples/terraform/azure/output.tf
xmudrii, line 38: Is it possible to define those resources in `main.tf` or is it required to put them in `output.tf`?

xmudrii (changes requested):
Thank you for the PR! I've left some comments.

/assign @kron4eg 
Please take a look at this PR as well.

examples/terraform/azure/main.tf
dharapvj: If we keep this, we get a deprecation notice for this attribute and that it will be removed in future versions of terraform.

examples/terraform/azure/output.tf
dharapvj: I can move the resources in main.tf. I kept them here as they hacks only specifically added for the ensuring right output. But I will move them.

examples/terraform/azure/output.tf
```diff
@@ -47,7 +47,7 @@ output "kubeone_workers" {
     # following outputs will be parsed by kubeone and automatically merged into
     # corresponding (by name) worker definition
     "${var.cluster_name}-pool1" = {
-      replicas = 1
+      replicas = var.worker_node_vm_count
```
xmudrii, line 50: This should also be changed to `initial_machinedeployment_replicas`.

xmudrii (commented):
One more comment, otherwise LGTM.

examples/terraform/azure/output.tf
dharapvj: Done

kron4eg (approved):

xmudrii (approved):
/approve

PR #1310, opened 2021-04-08
Use spot instances for E2E on AWS

kron4eg:
We have quite cool feature in machine-controller "spot instances" flag. 
This PR adds a demonstration of how to use it + we save some funds on our e2e tests.

Fixes #

terraform conversion to string with `format("%t"` is needed at otherwise it's treated as real bool and this is an error of json Unmarshal to `map[string]string`

```release-note
Turn on and off AWS spot instances in terraform config example
```

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -146,7 +146,7 @@ presubmits:
         - name: CONTAINER_RUNTIME
           value: containerd
         - name: TF_VAR_initial_machinedeployment_spotinstances
-          value: "true"
+          value: true
         - name: TEST_CLUSTER_TARGET_VERSION
           value: "1.18.17"
         - name: KUBEONE_TEST_RUN
@@ -174,7 +174,7 @@ presubmits:
             - name: PROVIDER
               value: "aws"
             - name: TF_VAR_initial_machinedeployment_spotinstances
-              value: "true"
+              value: true
             - name: TEST_CLUSTER_TARGET_VERSION
               value: "1.18.17"
             - name: KUBEONE_TEST_RUN
@@ -202,7 +202,7 @@ presubmits:
             - name: PROVIDER
               value: "aws"
             - name: TF_VAR_initial_machinedeployment_spotinstances
-              value: "true"
+              value: true
             - name: TEST_CLUSTER_TARGET_VERSION
               value: "1.19.9"
             - name: KUBEONE_TEST_RUN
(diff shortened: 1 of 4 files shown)


kron4eg:
/hold

I'd like to merge #1241 before this

examples/terraform/aws/variables.tf
```diff
@@ -166,3 +166,8 @@ variable "static_workers_count" {
   description = "number of static workers"
   default     = 0
 }
+
+variable "initial_machinedeployment_spotinstances" {
+  description = "use spot instances for initil machine-deployment"
```
xmudrii, line 171:
```suggestion
  description = "use spot instances for initial machine-deployment"
```

examples/terraform/aws/variables.tf
```diff
@@ -166,3 +166,8 @@ variable "static_workers_count" {
...
+  default     = true
```
xmudrii, line 172: I think that we should set the default value to false and then override the value in E2E tests.

```suggestion
  default     = false
```

kron4eg: Why?

examples/terraform/aws/variables.tf
kron4eg: 👍

examples/terraform/aws/variables.tf
kron4eg: OK done

.prow.yaml
```diff
@@ -145,6 +145,8 @@ presubmits:
           value: "aws"
         - name: CONTAINER_RUNTIME
           value: containerd
+        - name: TF_VAR_initial_machinedeployment_spotinstances
+          value: "true"
```
xmudrii, line 149: This looks okay to me, but we also need to change periodics.

An alternative approach would be that we apply this flag when provisioning the cluster, such as [here](https://github.com/kubermatic/kubeone/blob/a6096111313da857cfafd64c1a8b035df7b2c821/test/e2e/conformance_test.go#L199-L201) and [here](https://github.com/kubermatic/kubeone/blob/a6096111313da857cfafd64c1a8b035df7b2c821/test/e2e/upgrade_test.go#L179-L181). In that case, we don't need to change all jobs.

kron4eg: In indicated places we are forced to have hacks like this because we have to dynamically change variable. In general such configs should be explicitly shown.

xmudrii (approved):
/lgtm

PR #1328, opened 2021-04-26
Fix OWNERS file

kron4eg:
@eqrx has left us so I propose to replace him with @moelsayed 

```release-note
NONE
```

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,13 +1,13 @@
 # See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
 
 approvers:
-  - xmudrii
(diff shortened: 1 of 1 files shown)


OWNERS
```diff
@@ -4,10 +4,10 @@ approvers:
   - xmudrii
   - kron4eg
   - xrstf
-  - eqrx
+  - moelsayed
```
xmudrii, line 7: Nit: can we sort this file alphabetically?

kron4eg: done

xmudrii (approved):

PR #1334, opened 2021-04-28
kubeone list-images subcommand

### LINKED ISSUE #1277: Images list subcommand

`kubeone list-images` should reveal the images list that will be used in the given deployment, that take all the configurations in the account, including images from the addons.

At the moment we have static image-loader script which doesn't have complete images list and we have to maintain it manually.

kron4eg:
Fixes #1277

```release-note
kubeone list-images subcommand
```

### DIFF
--- a/pkg/cmd/list-images.go
+++ b/pkg/cmd/list-images.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/MakeNowJust/heredoc/v2"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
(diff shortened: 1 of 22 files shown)


pkg/cmd/list-images.go
```diff
@@ -0,0 +1,83 @@
...
+		},
+	}
+
+	return cmd
+}
+
+func listImages(opts *listImagesOpts) error {
+	var imgopts []images.Opt
+
+	configBuf, err := os.ReadFile(opts.ManifestFile)
```
xmudrii, line 60: We should add a comment briefly describing why we don't use the existing function for parsing the manifest. I guess the reason is that the existing function automatically validates the manifest, which we don't need here at all.

xmudrii (commented):
Two nits, but otherwise LGTM. Also, please address the lint error.

pkg/cmd/list-images.go
```diff
@@ -0,0 +1,83 @@
...
+	ManifestFile string `longflag:"manifest" shortflag:"m"`
+}
+
+func listImagesCmd(rootFlags *pflag.FlagSet) *cobra.Command {
+	opts := &listImagesOpts{}
+
+	cmd := &cobra.Command{
+		Use:     "list-images",
+		Short:   "List images that will be used",
+		Example: `kubeone list-images -m mycluster.yaml -t terraformoutput.json`,
```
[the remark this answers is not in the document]
kron4eg: 👍

pkg/cmd/list-images.go
kron4eg: 👍 
Yep, you are exactly right. Existing mechanism to "load" the config would also validate it and basically I'm forced to ether write full config manually for just image-list, or eve worse to create actual instances with terraform :D

xmudrii (approved):

xmudrii (approved):

PR #1335, opened 2021-04-29
Ignore preexisting static manifests kubeadm preflight error

### LINKED ISSUE #1261: Allow ignoring kubeadm preflight check for preexisting static manifests

In some deployment scenarios, users might use Terraform to create static manifests that would be used once the cluster is provisioned. However, kubeadm is failing a preflight check if there are preexisting manifests. We should add a feature that allows users to ignore that preflight check. Alternatively, we can ignore it uncoditionally.

xmudrii:

The preexisting static manifests kubeadm preflight error (`DirAvailable--etc-kubernetes-manifests`) is now unconditionally ignored. This allows users to use Terraform to create static manifests that would be used once the cluster is provisioned.

Fixes #1261 

Using the kubeadm configuration to ignore preflight errors works only with the kubeadm v1beta2 API. The v1beta2 API is used for 1.15+ clusters, which means that those preflight errors will not be ignored for clusters running Kubernetes 1.14 or older. We've discussed this internally and we believe this is okay because Kubernetes 1.14 has reached EOL a long time ago (December 2019).

```release-note
Ignore the preexisting static manifests kubeadm preflight error
```

/assign @kron4eg

### DIFF
--- a/pkg/scripts/kubeadm.go
+++ b/pkg/scripts/kubeadm.go
@@ -21,8 +21,7 @@ const (
 if [[ -f /etc/kubernetes/admin.conf ]]; then exit 0; fi
 
 sudo kubeadm join {{ .VERBOSE }} \
-	--config={{ .WORK_DIR }}/cfg/master_{{ .NODE_ID }}.yaml \
-	--ignore-preflight-errors=DirAvailable--var-lib-etcd,ImagePull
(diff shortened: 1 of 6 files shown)


kron4eg (approved):

toschneck:
cc @chrkl wasn't this request by one customer? Could you forward the info?

PR #1340, opened 2021-05-06
Containerd support for flatcar

### LINKED ISSUE #1333: Support containerd on Flatcar Linux

Flatcar now enables the CRI plugin by default, which means that we can use containerd for Kubernetes clusters. More details about this can be found in the [Flatcar docs](https://kinvolk.io/docs/flatcar-container-linux/latest/container-runtimes/switching-from-docker-to-containerd-for-kubernetes/).

mlavacca:
Since Flatcar now enables the CRI plugin by default, this PR introduces its support to containerd.

Fixes #1333 

```release-note
NONE
```

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -147,6 +147,41 @@ presubmits:
           resources:
             requests:
               cpu: 1
+  - name: pull-kubeone-e2e-aws-flatcar-containerd-conformance-1.18
+    run_if_changed: "(pkg/|examples/terraform/aws|hack/|test/)"
+    decorate: true
+    clone_uri: "ssh://git@github.com/kubermatic/kubeone.git"
+    labels:
+      preset-goproxy: "true"
+      preset-aws: "true"
+    spec:
+      containers:
+        - image: kubermatic/kubeone-e2e:v0.1.15
+          imagePullPolicy: Always
+          command:
+            - make
+          args:
+            - e2e-test
+          env:
+            - name: PROVIDER
+              value: "aws"
+            - name: CONTAINER_RUNTIME
+              value: containerd
+            - name: TEST_OS_CONTROL_PLANE
+              value: "flatcar"
+            - name: TEST_OS_WORKERS
+              value: "flatcar"
+            - name: TF_VAR_ssh_username
+              value: "core"
+            - name: TF_VAR_bastion_user
+              value: "core"
+            - name: TEST_CLUSTER_TARGET_VERSION
+              value: "1.18.17"
+            - name: KUBEONE_TEST_RUN
+              value: "TestClusterConformance"
+          resources:
+            requests:
+              cpu: 1
   - name: pull-kubeone-e2e-aws-conformance-1.18
     run_if_changed: "(pkg/|examples/terraform/aws|hack/|test/)"
     decorate: true
--- a/pkg/scripts/os.go
+++ b/pkg/scripts/os.go
@@ -22,6 +22,7 @@ import (
 
 const (
 	defaultKubernetesCNIVersion = "0.8.7"
+	defaultCriToolsVersion      = "1.21.0"
 )
 
 const (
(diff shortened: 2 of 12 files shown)


pkg/scripts/os.go
```diff
@@ -313,18 +313,27 @@ sudo systemctl restart kubelet
...
 
-sudo systemctl restart docker
-
 sudo mkdir -p /opt/cni/bin /etc/kubernetes/pki /etc/kubernetes/manifests
 curl -L "https://github.com/containernetworking/plugins/releases/download/v{{ .KUBERNETES_CNI_VERSION }}/cni-plugins-linux-${HOST_ARCH}-v{{ .KUBERNETES_CNI_VERSION }}.tgz" |
 	sudo tar -C /opt/cni/bin -xz
 
 RELEASE="v{{ .KUBERNETES_VERSION }}"
 
+curl -L https://github.com/kubernetes-sigs/cri-tools/releases/download/${RELEASE}/crictl-${RELEASE}-linux-${HOST_ARCH}.tar.gz |
```
crinjes, line 325: There is no 1:1 mapping between kubernetes and cri-tools releases, so I don't think this is correct.

pkg/scripts/testdata/TestKubeadmFlatcar-force.golden
```diff
@@ -19,21 +19,6 @@ aarch64)
...
-	],
-	"storage-driver": "overlay2",
-	"log-driver": "json-file",
-	"log-opts": {
-		"max-size": "100m"
-	}
-}
-EOF
-
-
```
kron4eg, line 36: this config should not be gone

pkg/scripts/os.go
mlavacca: Yes, you are right, CRI-tools are released only with the major ones. I'm going to fix it

kron4eg:
@mlavacca the PR itself looks good 👍 

New MC release v1.29.0 has been just released. Release that include kubermatic/machine-controller#958. Please integrate that new version as well an this PR will feature complete support for containerd on flatcar.

kron4eg:
> @mlavacca the PR itself looks good 👍
> 
> New MC release v1.29.0 has been just released. Release that include [kubermatic/machine-controller#958](https://github.com/kubermatic/machine-controller/pull/958). Please integrate that new version as well an this PR will feature complete support for containerd on flatcar.

It means simply changing version here to v1.29.0:

https://github.com/kubermatic/kubeone/blob/2bd1393f44b0a185a24165e1c4be986365adcabd/pkg/templates/images/images.go#L56

kron4eg (approved):

kron4eg:
@xmudrii PTAL

.prow.yaml
```diff
@@ -147,6 +147,39 @@ presubmits:
           resources:
             requests:
               cpu: 1
+  - name: pull-kubeone-e2e-aws-flatcar-containerd-conformance-1.18
```
xmudrii, line 150: I'm not sure should we run Flatcar E2E tests on each PR. Currently, we're testing Flatcar periodically (each workday in the morning), and that seems to be enough. Historically, Flatcar tests were flakier than Ubuntu ones, although, they seem much more stable nowadays.

Should we move this test to periodic along with other Flatcar E2E tests?

@kron4eg WDYT?

.prow.yaml
```diff
@@ -147,6 +147,39 @@ presubmits:
...
+            - make
+          args:
+            - e2e-test
+          env:
+            - name: PROVIDER
+              value: "aws"
+            - name: CONTAINER_RUNTIME
+              value: containerd
+            - name: TEST_OS_CONTROL_PLANE
+              value: "flatcar"
```
xmudrii, line 171: Besides this variable, you also need to set OS for worker nodes, like this:

```yaml
- name: TEST_OS_WORKERS
  value: "flatcar"
```

pkg/scripts/testdata/TestKubeadmFlatcar-force.golden
```diff
@@ -62,13 +47,43 @@ EOF
...
 sudo mkdir -p /opt/cni/bin /etc/kubernetes/pki /etc/kubernetes/manifests
 curl -L "https://github.com/containernetworking/plugins/releases/download/v0.8.7/cni-plugins-linux-${HOST_ARCH}-v0.8.7.tgz" |
 	sudo tar -C /opt/cni/bin -xz
 
 RELEASE="v1.17.4"
+MAJOR_RELEASE="$(cut -d '.' -f 1,2 <<< ${RELEASE}).0"
+
+curl -L https://github.com/kubernetes-sigs/cri-tools/releases/download/${MAJOR_RELEASE}/crictl-${MAJOR_RELEASE}-linux-${HOST_ARCH}.tar.gz |
+	sudo tar -C /opt/bin -xz
```
xmudrii, line 58: The crictl (cri-tools) are not part of the Kubernetes release lifecycle (i.e. they are released separately). This approach will not work for newly released Kubernetes versions. For example, once Kubernetes 1.22 goes out, it might take some time until the cri-tools team doesn't release cri-tools v1.22. If someone would try to provision Kubernetes 1.22 before cri-tools v1.22 is out, the cluster provisioning would fail.

As you can use any cri-tools version with any Kubernetes version, I would prefer if we would bind this version, instead of depending on the Kubernetes version.

xmudrii (changes requested):

.prow.yaml
kron4eg: I'd keep it, for that we could spontaneously verify that it's still working when we might have doubts.

kron4eg (approved):

xmudrii (approved):
/approve

PR #1341, opened 2021-05-09
ensure consistent package aliases

xrstf:
gimps 0.4 has learned to do this. Usage is the same, just run `gimps .`

```release-note
NONE
```

### DIFF
--- a/.gimps.yaml
+++ b/.gimps.yaml
@@ -23,3 +23,22 @@ sets:
     patterns:
       - 'k8s.io/**'
       - '*.k8s.io/**'
+exclude:
+  - "pkg/apis/kubeadm/**"
+  - "vendor/**"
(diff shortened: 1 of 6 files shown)


pkg/apis/kubeadm/v1beta1/types.go
```diff
@@ -17,7 +17,7 @@ limitations under the License.
 package v1beta1
 
 import (
-	v1 "k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
```
kron4eg, line 20: This package is copied from the upstream, not written by us so I'd like to avoid changing them.

xrstf: No problemo, fixed.

PR #1358, opened 2021-05-21
Use latest available (wildcard) docker and containerd version

kron4eg:
Since those latest versions usually bring in security fixes, version constraints are relaxed and "latest available" version will be installed automatically.

We already doing the same in machine-controller

```release-note
Use latest available (wildcard) docker and containerd version
```

### DIFF
--- a/pkg/scripts/funcs.go
+++ b/pkg/scripts/funcs.go
@@ -83,11 +83,10 @@ var (
 )
 
 const (
-	defaultDockerVersion           = "19.03.14"
-	defaultAmazonDockerVersion     = "19.03.13"
-	defaultLegacyDockerVersion     = "18.09.9"
-	defaultContainerdVersion       = "1.4.3"
-	defaultAmazonContainerdVersion = "1.4.1"
+	defaultDockerVersion           = "19.03.*"
+	defaultLegacyDockerVersion     = "18.09.*"
+	defaultContainerdVersion       = "1.4.*"
+	defaultAmazonContainerdVersion = "1.4.*"
 	defaultAmazonCrictlVersion     = "1.13.0"
 )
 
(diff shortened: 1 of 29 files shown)


pkg/scripts/funcs.go
```diff
@@ -83,11 +83,10 @@ var (
 )
 
 const (
-	defaultDockerVersion           = "19.03.14"
-	defaultAmazonDockerVersion     = "19.03.13"
-	defaultLegacyDockerVersion     = "18.09.9"
-	defaultContainerdVersion       = "1.4.3"
-	defaultAmazonContainerdVersion = "1.4.1"
+	defaultDockerVersion           = "19.03.*"
```
xmudrii, line 86: Can we bump the Docker version for 1.21+ clusters? As per the `dependencies.yaml` file, the recommended versions are:

* Docker 19.03 for Kubernetes 1.17+ ([`dependencies.yaml` for 1.17](https://github.com/kubernetes/kubernetes/blob/1d9e1e6d56cbd5c9d04ba25cccf184a973056a79/build/dependencies.yaml#L40-L44))
* Docker 20.10 for Kubernetes 1.21+ ([`dependencies.yaml` for 1.21](https://github.com/kubernetes/kubernetes/blob/1348146e68ffa14262beeaddfa6f7c61a359fff2/build/dependencies.yaml#L72-L77))

xmudrii (changes requested):

pkg/scripts/funcs.go
kron4eg: Docker 20.10 for Kubernetes 1.21+, but none (dockershim no more) for kubernetes 1.22+ :D

kron4eg: it's done

xmudrii (approved):

PR #1361, opened 2021-05-26
Add worker_deploy_ssh_key Terraform variable to the AWS configs

xmudrii:

Add `worker_deploy_ssh_key` Terraform variable to the AWS configs. This variable controls should the provided SSH public key be embedded into the MachineDeployment objects. The default value is `true`, i.e. the key will be embedded.

The main use case for this feature is to disable embedding SSH keys for Flatcar MachineDeploments. That's because the Flatcar Machines can reach the AWS user-data limit if the SSH key is provided. This is a known issue and we're working on resolving it. In meanwhile, this approach can be used as a workaround.

```release-note
Add `worker_deploy_ssh_key` Terraform variable to the AWS configs. This variable controls should the provided SSH public key be embedded into the MachineDeployment objects. The default value is `true`, i.e. the key will be embedded.
```

/assign @kron4eg

### DIFF
--- a/examples/terraform/aws/output.tf
+++ b/examples/terraform/aws/output.tf
@@ -73,7 +73,7 @@ output "kubeone_workers" {
     "${var.cluster_name}-${local.zoneA}" = {
       replicas = var.initial_machinedeployment_replicas
       providerSpec = {
-        sshPublicKeys   = [aws_key_pair.deployer.public_key]
+        sshPublicKeys   = local.worker_deploy_ssh_key
         operatingSystem = local.worker_os
         operatingSystemSpec = {
           distUpgradeOnBoot = false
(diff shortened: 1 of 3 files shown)


examples/terraform/aws/output.tf
```diff
@@ -73,7 +73,7 @@ output "kubeone_workers" {
     "${var.cluster_name}-${local.zoneA}" = {
       replicas = var.initial_machinedeployment_replicas
       providerSpec = {
-        sshPublicKeys   = [aws_key_pair.deployer.public_key]
+        sshPublicKeys   = var.worker_deploy_ssh_key ? [aws_key_pair.deployer.public_key] : []
```
kron4eg, line 76: To avoid repeating the same thing, please define a local var (which will equal to this exact ternary expression) and use it in the rest of the config.

xmudrii: Good point, changed. PTAL.

kron4eg (approved):

PR #1362, opened 2021-06-01
Docker to containerd automated migration

### LINKED ISSUE #1350: automate Docker -> Containerd migration

kron4eg:

Fixes #1350

```release-note
Docker to containerd automated migration
```

### DIFF
--- a/pkg/scripts/render.go
+++ b/pkg/scripts/render.go
@@ -39,7 +39,7 @@ var (
 			{{ end }}
 
 			{{- if or .FORCE .UPGRADE }}
-			sudo apt-mark unhold docker-ce docker-ce-cli
+			sudo apt-mark unhold docker-ce docker-ce-cli containerd.io || true
 			{{- end }}
 
 			{{- $DOCKER_VERSION_TO_INSTALL := "%s" }}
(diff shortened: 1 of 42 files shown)


pkg/scripts/render.go
```diff
@@ -58,19 +79,24 @@ var (
...
 			defaultDockerVersion,
 			defaultLegacyDockerVersion,
 			latestDockerVersion,
+			defaultContainerdVersion,
 		),
 
 		"yum-docker-ce-amzn": heredoc.Docf(`
 			{{- if or .FORCE .UPGRADE }}
-			sudo yum versionlock delete docker cri-tools || true
+			sudo yum versionlock delete docker cri-tools containerd
```
xmudrii, line 99: I think that we should keep `|| true` here because of:

* older clusters that might not have `versionlock` installed
* if some package isn't version-locked (I'm not sure is `versionlock` failing in that case -- IIRC, it doesn't, but we should double-check)

pkg/tasks/tasks.go
```diff
@@ -292,6 +292,35 @@ func WithReset(t Tasks) Tasks {
...
+				},
+				ErrMsg: "failed to download Kubernetes PKI from the leader",
+			},
+			{
+				Fn: func(s *state.State) error {
+					if err := machinecontroller.Ensure(s); err != nil {
+						return err
+					}
+
+					s.Logger.Warn("roll-over your machineDeployments to get containerd")
```
xmudrii, line 315: Tiny nit: we usually start log messages with a capital letter. Also, what do you think about rephrasing this message to include some more details?

```suggestion
					s.Logger.Warn("Roll-over your MachineDeployments to migrate the worker nodes to containerd")
```

pkg/scripts/render.go
```diff
@@ -184,79 +215,32 @@ var (
 			sudo yum-config-manager --save --setopt=docker-ce-stable.module_hotfixes=true
 			{{ end }}
 
-			sudo yum install -y yum-plugin-versionlock
```
xmudrii, line 187: I don't think it's guaranteed that the `versionlock` plugin will be installed by default. We should still install it explicitly to be sure.

xmudrii (commented):
Left some tiny nits, but otherwise, it looks great.

I just have two additional remarks:

* We should document this feature before closing the epic. Please create a subtask for documentation.
* Have you tested this on all supported operating systems? If not, we should do that before merging the PR.

pkg/scripts/render.go
kron4eg: * the `yum-plugin-versionlock` is already installed at this point so it's always present.
* `yum versionlock delete` is very liberal and if you pass it even non-existing packages it wouldn’t complain (e.g. giving you non-zero exist status)

pkg/scripts/render.go
kron4eg: It's removed here because it's the second time it's installed. So we still have versionlock plugin.

kron4eg:
>Have you tested this on all supported operating systems? If not, we should do that before merging the PR.

There might be a problem with flatcar.

pkg/tasks/tasks.go
kron4eg: OK, changed that to a link to doc with how to launch a rolling restart of machineDeployment.

kron4eg:
/hold

until docs are ready

kron4eg:
@xmudrii despite it's onhold (waiting for docs), it's ready for a review.

kron4eg:
docs PR: https://github.com/kubermatic/docs/pull/712

PR #1390, opened 2021-06-16
Fix hetzner ssh_port terraform type

kron4eg:
`ssh_port` should be a number (like in other providers) in hetzner terraform config, otherwise it blows up at JSON unmarshal time in kubeone.

```release-note
Fix hetzner ssh_port terraform type
```


xmudrii (approved):
/lgtm
/approve

PR #1407, opened 2021-06-29
Migrate WeaveNet to YAML addons

### LINKED ISSUE #1353: Convert WeaveNet Go-based template to YAML addon

kron4eg:
Fixes #1353 

```release-note
Migrate WeaveNet to YAML addons
```

### DIFF
--- a/pkg/tasks/cni.go
+++ b/pkg/tasks/cni.go
@@ -33,22 +33,19 @@ func ensureCNI(s *state.State) error {
 			return err
 		}
 	case s.Cluster.ClusterNetwork.CNI.WeaveNet != nil:
-		return ensureCNIWeaveNet(s)
+		if s.Cluster.ClusterNetwork.CNI.WeaveNet.Encrypted {
+			if err := weave.EnsureSecret(s); err != nil {
+				return err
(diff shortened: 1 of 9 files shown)


pkg/tasks/cni.go
```diff
@@ -33,22 +33,19 @@ func ensureCNI(s *state.State) error {
 			return err
 		}
 	case s.Cluster.ClusterNetwork.CNI.WeaveNet != nil:
-		return ensureCNIWeaveNet(s)
+		if s.Cluster.ClusterNetwork.CNI.WeaveNet.Encrypted {
+			if err := weave.EnsureSecret(s); err != nil {
+				return err
+			}
+		}
```
xmudrii, line 40: What do you think about replacing this with YAML/addon instead? I think it should be possible. We would probably have to add a template function for generating the password.

kron4eg: weave.EnsureSecret() will not override secret in case if it's already exists, this is not possible with YAML, and probably there is no point to even try.

xmudrii (approved):
/lgtm
/approve

PR #1412, opened 2021-07-07
Migrate machine-controller to new addons mechanism

### LINKED ISSUE #1356: Convert machine-controller Go-based template to YAML addon

kron4eg:
Fixes #1356

```release-note
Migrate machine-controller to new addons mechanism
```


addons/machinecontroller/machine-controller.yaml
```diff
@@ -0,0 +1,672 @@
...
+            - -node-no-proxy={{ . }}
+            {{ end }}
+            {{ with .Config.RegistryConfiguration.InsecureRegistryAddress }}
+            - -node-insecure-registries={{ . }}
+            {{ end }}
+            {{ if .Config.CABundle }}
+            - -ca-bundle={{ .Resources.CABundleSSLCertFilePath }}
+            {{ end }}
+            # TODO: handle dynamic pause-image
+            # - -node-pause-image=
```
xmudrii, line 497: Can we create a follow-up issue for this?

pkg/certificate/certificate.go
```diff
@@ -62,3 +64,37 @@ func CAKeyPair(config *configupload.Configuration) (*rsa.PrivateKey, *x509.Certi
 
 	return rsaKey, certs[0], nil
 }
+
+func NewSignedWebhookCert(name string, namespace string, caKey crypto.Signer, caCert *x509.Certificate) (map[string]string, error) {
+	serviceCommonName := fmt.Sprintf("%s.%s.svc", name, namespace)
+	serviceFQDNCommonName := fmt.Sprintf("%s.cluster.local.", serviceCommonName)
```
xmudrii, line 70: Shouldn't we use FQDN from the KubeOneCluster object instead of defaulting to `.cluster.local.`?

pkg/templates/machinecontroller/helper.go
```diff
@@ -22,36 +22,43 @@ import (
...
+	mcAppLabelValue = "machine-controller"
+)
+
+// machineControllerWebhook related constants
+const (
+	whName          = "machine-controller-webhook"
+	whAppLabelKey   = mcAppLabelKey
+	whAppLabelValue = whName
+	whNamespace     = mcNamespace
+)
```
xmudrii, line 54: Don't we define some of those consts in the resources package as well? Maybe we could reuse some of the consts from that package instead of redefining them.

kron4eg: good point

pkg/certificate/certificate.go
kron4eg: yeah (it was originally like that).

addons/machinecontroller/machine-controller.yaml
kron4eg: I've fixed it

xmudrii (approved):
/lgtm
/approve

PR #1413, opened 2021-07-09
machine-controller: Inject EnvVarBindings only for needed credentials

xmudrii:

This PR fixes two bugs:

1) Inject EnvVarBindings only for the needed credentials. Currently, we inject bindings for all credentials EnvVars found in the environment. However, the Secret only contains the credentials for the cloud provider. This can cause the machine-controller to fail to start because it has envVarBindings for credentials that are not present in the Secret.
1) Use envVar names required by machine-controller. Currently, we use the same envVar names as used by KubeOne, however, machine-controller has different names for some variables.

```release-note
NONE
```

/assign @kron4eg

### DIFF
--- a/pkg/addons/applier.go
+++ b/pkg/addons/applier.go
@@ -32,6 +32,8 @@ import (
 	"k8c.io/kubeone/pkg/state"
 	"k8c.io/kubeone/pkg/templates/images"
 	"k8c.io/kubeone/pkg/templates/resources"
+
+	"sigs.k8s.io/yaml"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pkg/addons/applier.go
```diff
@@ -50,11 +52,12 @@ type applier struct {
...
 type templateData struct {
-	Config         *kubeoneapi.KubeOneCluster
-	Certificates   map[string]string
-	Credentials    map[string]string
-	InternalImages *internalImages
-	Resources      map[string]string
+	Config             *kubeoneapi.KubeOneCluster
+	Certificates       map[string]string
+	Credentials        map[string]string
+	CredentialsEnvVars string
```
kron4eg, line 58:
```suggestion
	MachineControllerCredentialsEnvVars string
```

I know it's quite long name, but it makes the purpose much more transparent.

kron4eg (commented):
looks good, but let's consider renaming the template date variable to make it even more awesome :D

pkg/addons/applier.go
xmudrii: @kron4eg Changed, PTAL.

kron4eg (approved):

PR #1448, opened 2021-08-16
Extend restart API server script to handle failing crictl logs

xmudrii:

This PR extends the restart API server script to handle the failing `crictl logs` command. The `crictl logs` command can fail on subsequent script run if Kubelet failed to create symlink when restarting the container on the previous script run.

This issue is often triggered on upgrade tests where the `crictl logs` command fails when upgrading the cluster. For example: https://public-prow.loodse.com/view/gs/prow-dev-public-data/pr-logs/pull/kubermatic_kubeone/1447/pull-kubeone-e2e-aws-upgrade-1.20-1.21/1427238137196711936

```release-note
Extend restart API server script to handle failing "crictl logs" due to missing symlink
```

/assign @kron4eg

### DIFF
--- a/pkg/scripts/node.go
+++ b/pkg/scripts/node.go
@@ -26,6 +26,13 @@ var (
 	`)
 
 	restartKubeAPIServerCrictlTemplate = heredoc.Doc(`
+		# Disable exit immediately if a command in a pipeline fails.
+		# crictl logs can fail if kubelet fails to set up symlink for the API
+		# server container logs. This usually happens on subsequent script run,
+		# for example when upgrading the cluster.
+		# This is a known issue and it's described here:
+		# https://github.com/kubernetes/kubernetes/issues/52172
+		set +e
 		apiserver_id=$(sudo crictl ps --name=kube-apiserver -q)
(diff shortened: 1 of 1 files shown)


kron4eg (approved):

pkg/scripts/node.go
```diff
@@ -26,14 +26,21 @@ var (
 	`)
 
 	restartKubeAPIServerCrictlTemplate = heredoc.Doc(`
+		# Disable exit immediately if a command in a pipeline fails.
+		# crictl logs can fail if kubelet fails to set up symlink for the API
+		# server container logs. This usually happens on subsequent script run,
+		# for example when upgrading the cluster.
+		# This is a known issue and it's described here:
+		# https://github.com/kubernetes/kubernetes/issues/52172
+		set +e
```
xmudrii, line 35: Do we need to restart it back to `set -xeu pipefail` or KubeOne will automatically do that when running the next step?

kron4eg: next script invocation will be prefixed with `set -xeu pipefail` automatically.

PR #1462, opened 2021-08-19
New Addons API

### LINKED ISSUE #1459: Implement the Addons API
We have implemented a mechanism for embedding addons into the KubeOne binary. We need to implement an API that can be used to enable optional addons embedded in the binary, such as `cluster-autoscaler` and `unattended-upgrades`. That would make it much easier to use those addons because users wouldn't have to download them from GitHub and put them in their addons directory.

kron4eg:

xref #1459 

```release-note
New Addons API
```


pkg/apis/kubeone/v1beta1/types.go
```diff
@@ -590,12 +590,31 @@ type OpenIDConnectConfig struct {
...
+
 	// Path on the local file system to the directory with addons manifests.
-	Path string `json:"path"`
+	Path string `json:"path,omitempty"`
+
+	// GlobalParams to the addon, to render all addons using text/template
+	GlobalParams map[string]string `json:"globalParams,omitempty"`
+
+	// Addons is a list of config options for named addon
+	Addons []Addon `json:"addons,omitempty"`
```
xmudrii, line 617: Have you implemented enabling embedded addons via this API yet?

kron4eg: No, this is out of scope.

xmudrii: This is actually the key part of the Epic. Passing parameters to addons was supposed to be only a nice addition if we have time left, but not the key part. Quoting part of the Epic description:

> We need to implement an API that can be used to enable optional addons embedded in the binary, such as `cluster-autoscaler` and `unattended-upgrades`.

Can you please extend the PR to handle this as well? I think it should be fairly easy, but it might take some time to test it properly.

PR #1488, opened 2021-09-03
Add default-storage-class optional addon

xmudrii:

This PR ports the `default-storage-class` addon from KKP: https://github.com/kubermatic/kubermatic/tree/master/addons/default-storage-class

This is an optional addon that can be deployed using the new Addons API:

```yaml
addons:
  enable: true
  addons:
  - name: default-storage-class
```

This should make it easier to test the CCM/CSI migration. 

```release-note
Add an optional default-storage-class addon used to create StorageClass for AWS, Azure, GCP, OpenStack, vSphere, and Hetzner clusters. The addon can be enabled using the Addons API.
```


mlavacca (approved):
/approve
/lgtm

PR #1517, opened 2021-09-13
Default MTU in calico-vxlan addon

kron4eg:
Otherwise it causes crashloop since MTU is not defaulted and results to JSON error

```release-note
Default MTU in calico-vxlan addon
```


moadqassem (commented):
/approve
/lgtm

Issue #1307, opened 2021-04-05
Azure ingress is not accessible via Load Balancer


dharapvj:
**What happened**:
When I install nginx-ingress-controller on kubeone based cluster in Azure, the ingress URLs timeout.

**What is the expected behavior**:
Ingress based app access should work.

**How to reproduce the issue**:
* Used latest kubeone (v1.12) terraform example for Azure to create infrastructure and subsequently used kubeone apply to create kubernetes cluster
* Used nginx-ingress-controller helm chart / direct url from [guide](https://kubernetes.github.io/ingress-nginx/deploy/#azure) to install nginx ingress controller. Controller gets installed without any errors.
* Try to access Load Balancer IP from browser - it is not accessible.
* Try to install simple app and ingress definition, the ingress based app is not accessible.

**Anything else we need to know?**
I observed that if I 
1. Create another availability set in terraform
2. move the worker node to different availability set
2. change the `primaryAvailabilitySetName` to this new avset in kubeone.yaml
ingress becomes available. But I do not know reason for this behavior. Also I am not sure the impact of changing value of `primaryAvailabilitySetName` in kubeone.yaml

**Information about the environment**:
KubeOne version (`kubeone version`): 
```
{
  "kubeone": {
    "major": "1",
    "minor": "2",
    "gitVersion": "1.2.0-rc.1",
    "gitCommit": "fde1f267769e04acc07bbdb94c09b0d6a18ea4cc",
    "gitTreeState": "",
    "buildDate": "2021-03-12T10:25:53Z",
    "goVersion": "go1.16.1",
    "compiler": "gc",
    "platform": "linux/amd64"
  },
  "machine_controller": {
    "major": "1",
    "minor": "25",
    "gitVersion": "v1.25.0",
    "gitCommit": "",
    "gitTreeState": "",
    "buildDate": "",
    "goVersion": "",
    "compiler": "",
    "platform": "linux/amd64"
  }
}
```
Operating system: `linux`
Provider you're deploying cluster on: `Azure`
Operating system you're deploying on: `Ubuntu`

JoaquimFreitas:
Hi,
[...]

PR #1528, opened 2021-09-17
azure-terraform: allow nodeports

kron4eg:
Open nodeports by default for azure cluster

Found during v1.3.0 testing

Potentially fixes `Azure ingress is not accessible via Load Balancer` #1307

```release-note
* azure-terraform: open azure nodeports in firewall
* azure-terraform: default VM type is changed to Standard_F2
```

### DIFF
--- a/examples/terraform/azure/variables.tf
+++ b/examples/terraform/azure/variables.tf
@@ -69,13 +69,13 @@ variable "location" {
 
 variable "control_plane_vm_size" {
   description = "VM Size for control plane machines"
-  default     = "Standard_B2s"
+  default     = "Standard_F2"
   type        = string
 }
 
 variable "worker_vm_size" {
   description = "VM Size for worker machines"
-  default     = "Standard_B2s"
+  default     = "Standard_F2"
   type        = string
 }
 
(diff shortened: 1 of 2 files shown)


examples/terraform/azure/variables.tf
```diff
@@ -69,13 +69,13 @@ variable "location" {
...
   description = "VM Size for control plane machines"
-  default     = "Standard_B2s"
+  default     = "Standard_F2"
   type        = string
 }
 
 variable "worker_vm_size" {
   description = "VM Size for worker machines"
-  default     = "Standard_B2s"
+  default     = "Standard_F2"
```
xmudrii, line 78: Is this going to cause the VMs to be recreated or only updated in-place? Regardless, I'd mention this in the release notes.

kron4eg: That will force-replace VMs yes. However during pre-release testing `Standard_F2` proved to be much faster. Also `Standard_F2` is used in MC e2e tests and our account has wide-big quota for `Standard_F2`s.

xmudrii: Can you add a note about this in the release notes? This sounds like a breaking change, so users should be aware of this.

kron4eg: ✅

xmudrii (approved):

PR #1534, opened 2021-09-20
Update Go to 1.17.1

kron4eg:
Move forward with new Go version since there is a difference in codegen

xref #1540

```release-note
NONE
```


kron4eg:
For now this PR should not be merged. I had impression that we are moving kubermatic to 1.17 as well but seems it's not the case. So holding this for now.

xmudrii (approved):
/lgtm

PR #1544, opened 2021-09-28
CCM/CSI migration support for clusters with static worker nodes

### LINKED ISSUE #1450: Implement the CCM/CSI migration mechanism
Kubernetes has deprecated the in-tree cloud providers in favor of external CCMs (cloud controller managers) and CSI drivers. The in-tree cloud providers will be removed in the future, so we need to ensure users can easily migrate to CCM/CSI. In addition to deprecation, many providers are already providing new features and fixes only to external CCMs.

This is a collector issue (Epic) for tasks related to implementing the CCM/CSI migration mechanism.

In the initial phase, we'll support the CCM/CSI migration only for OpenStack and potentially vSphere. We're planning to add support for other providers (e.g. Azure, AWS, GCP) in one of the future versions.

**Related issues:**

#1451
#1452
#1454
#1455
https://github.com/kubermatic/docs/issues/754

### LINKED ISSUE #1495: Support the CCM/CSI migration for static worker nodes
Currently, the CCM/CSI migration is not supported for clusters with static worker nodes. In such a case, validation would fail and KubeOne will not proceed with the migration.

This is because `kubeadm join phase kubelet-start` works differently compared to `kubeadm init phase kubelet-start` used for control plane nodes. The `init` command just generates the config and restarts Kubelet. However, the `join` command also tries to TLS bootstrap the node and then join it to a cluster, but that fails because the node is already joined.

One possible approach to this problem could be:

* Download Kubelet config from the node
* Unmarshal it to the Go struct
* Add the required Go flags
* Marshal it back to YAML
* Cordon and drain the node
* Upload the new configuration file
* Restart Kubelet and wait for it to become ready
* Uncrodon the node

xmudrii:

This PR introduces CCM/CSI migration support for clusters with the static worker nodes. As described in #1495, `kubeadm` cannot be used to handle updating kubelet configuration and flags. This is because `kubeadm init` (used for control plane nodes) works differently than `kubeadm join` (used for static worker nodes).

To mitigate that, we manually update the required files, which we already do for Docker to containerd migration.

The workflow is:

* Cordon and drain a static worker node
* Download the kubelet config file (`/var/lib/kubelet/config.yaml`)
* Add the required feature gates
* Upload the new kubelet config file
* Download the `kubeadm-flags.env` file (`/var/lib/kubelet/kubeadm-flags.env`)
* Set `--cloud-provider=external` and remove the `--cloud-config` flag
* Upload the updated `kubeadm-flags.env`
* Restart kubelet and wait up to 2 minutes for it to become running
* Uncordon the node

This PR has been already tested on an OpenStack cluster.

Fixes #1495
Fixes #1450

```release-note
CCM/CSI migration support for clusters with static worker nodes
```

### DIFF
--- a/pkg/tasks/ccm_csi_migration.go
+++ b/pkg/tasks/ccm_csi_migration.go
@@ -17,7 +17,9 @@ limitations under the License.
 package tasks
 
 import (
+	"bytes"
 	"fmt"
+	"io"
 	"strconv"
 	"time"
 
@@ -27,6 +29,7 @@ import (
 	"k8c.io/kubeone/pkg/nodeutils"
 	"k8c.io/kubeone/pkg/scripts"
 	"k8c.io/kubeone/pkg/ssh"
+	"k8c.io/kubeone/pkg/ssh/sshiofs"
 	"k8c.io/kubeone/pkg/state"
 
 	"github.com/kubermatic/machine-controller/pkg/apis/cluster/common"
@@ -44,7 +47,7 @@ const (
 	provisionedByOpenStackCSICinder    = "cinder.csi.openstack.org"
 )
 
-func validateExternalCloudProviderConfig(s *state.State) error {
+func ccmMigrationValidateConfig(s *state.State) error {
 	if !s.Cluster.CloudProvider.External {
 		return errors.New(".cloudProvider.external must be enabled to start the migration")
 	}
@@ -59,14 +62,11 @@ func validateExternalCloudProviderConfig(s *state.State) error {
 	if s.Cluster.CloudProvider.Vsphere != nil && s.Cluster.CloudProvider.CSIConfig == "" {
 		return errors.New("the ccm/csi migration for vsphere requires providing csi configuration using .cloudProvider.csiConfig field")
 	}
-	if len(s.Cluster.StaticWorkers.Hosts) > 0 {
-		return errors.New("the ccm/csi migration for cluster with static worker nodes is currently unsupported")
-	}
 
 	return nil
 }
 
-func readyToCompleteMigration(s *state.State) error {
+func readyToCompleteCCMMigration(s *state.State) error {
 	if s.DynamicClient == nil {
 		return errors.New("clientset not initialized")
 	}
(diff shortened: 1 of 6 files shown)


pkg/tasks/ccm_csi_migration.go
```diff
@@ -189,6 +233,107 @@ func updateKubeletConfigInternal(s *state.State, node *kubeoneapi.HostConfig, co
...
+	if _, err = fw.Seek(0, io.SeekStart); err != nil {
+		return err
+	}
+
+	if _, err = io.Copy(fw, bytes.NewBuffer(buf)); err != nil {
+		return err
+	}
+
+	return nil
+}
```
xrstf, line 335: Not sure if it's worth it, but `ccmMigrationUpdateKubeletFlags()` and `ccmMigrationUpdateKubeletConfigFile()` are practically identical and only have very slight differences in behaviour. Maybe it make sense to extrapolate into a `updateRemoteFile(state *State, file string, modifier func(content string) string) error` (pseudocode)?

pkg/tasks/common.go
```diff
@@ -0,0 +1,100 @@
...
+	if len(s1) != 2 {
+		return nil, errors.New("can't parse: wrong split length")
+	}
+
+	envValue := strings.Trim(s1[1], `"`)
+	flagsvalues := strings.Split(envValue, " ")
+	kubeletflagsMap := map[string]string{}
+
+	for _, flg := range flagsvalues {
+		fl := strings.Split(flg, "=")
```
xrstf, line 49: No SplitN here?

pkg/tasks/common.go
```diff
@@ -0,0 +1,100 @@
...
+	var buf bytes.Buffer
+	fmt.Fprintf(&buf, `%s="`, kubeletKubeadmArgsEnv)
+
+	for i, val := range kvpairs {
+		format := "%s "
+		if i == len(kvpairs)-1 {
+			format = "%s"
+		}
+		fmt.Fprintf(&buf, format, val)
+	}
```
xrstf, line 76: Is this the same as `strings.Join(kvpairs, " ")`? If so, this can be reduced to

```
return fmt.Sprintf(`%s="%s"`, kubeletKubeadmArgsEnv, strings.Join(kvpairs, " "))
```

xrstf (commented):
Consider yourself reviewed!

pkg/tasks/ccm_csi_migration.go
xmudrii: I really like this idea. Addressed in a follow-up commit.

pkg/tasks/common.go
xmudrii: Fixed and added a test case.

xmudrii:
/hold
to manually test it

xrstf (approved):
/approve

xmudrii:
/hold cancel
tested manually and everything works fine

PR #1545, opened 2021-09-28
Fix Go build tags

xmudrii:

We didn't fix Go build tags for all files when updating to Go 1.17.1, so this PR fixes the build tags for the remaining files.

```release-note
NONE
```


xrstf (approved):
/approve

PR #1552, opened 2021-09-29
Default image in the cluster-autoscaler addon

xmudrii:

Currently, the cluster-autoscaler addon cannot be used as an embedded addon because we don't default the image/version. Instead, users still have to copy the addon to their addons directory and then manually modify the image.

The embedded addons are supposed to be used without having to copy and modify them, and this PR modifies the cluster-autoscaler addon to support this use case.

There are two changes in this PR:

* The image is now defaulted depending on the Kubernetes version of the cluster
* The image can be overridden via addon parameters:

```yaml
...
addons:
  enable: true
  path: "./addons"
  addons:
  - name: cluster-autoscaler
    params:
      CLUSTER_AUTOSCALER_IMAGE: "k8s.gcr.io/autoscaling/cluster-autoscaler:v1.21.0"
```

```release-note
Default image in the cluster-autoscaler addon and allow the image to be overridden using addon parameters
```

### DIFF
--- a/addons/cluster-autoscaler/cluster-autoscaler.yaml
+++ b/addons/cluster-autoscaler/cluster-autoscaler.yaml
@@ -147,7 +147,7 @@ spec:
         app: cluster-autoscaler
     spec:
       containers:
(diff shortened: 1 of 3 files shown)


xrstf (approved):
/approve

PR #1557, opened 2021-10-03
Use gocache in CI when building KubeOne

xmudrii:

This PR proposes using the gocache mechanism in the CI to speed up the build jobs. The build times right now are on average 2-3 minutes, but depending on the load, the build times can go up to 5-10 minutes. Building all the dependencies each time also requires more resources.

The gocache mechanism should solve this and reduce build times and resource consumption. Right now, we're only using it for building the binary, but if it works, we'll use it for other tasks as well.

The gocache scripts are based on work @xrstf did in #1075 and for KKP and other projects.

This PR also brings some utility functions from KKP which we'll use for other scripts in the future.

```release-note
NONE
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ terraform.tfvars
 *kubeconfig
 !*kubeconfig/
 /vendor
+download-gocache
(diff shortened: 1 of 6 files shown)


xrstf (approved):
/approve

let's try it!

xrstf:
/lgtm cancel

please add the download-gocache file to the .gitignore

PR #1558, opened 2021-10-03
Disable colors for Terraform and Ginkgo in CI

xmudrii:

The colors in the CI are making Spyglass load much slower while previewing raw logs is hard because of all the ANSI escape characters. This PR disables colors for Terraform and Ginkgo (Kubernetes conformance tests) to fix this. 

```release-note
NONE
```


xrstf (approved):
/approve

PR #1559, opened 2021-10-03
Use gocache in test and e2e-test Make targets

xmudrii:

We started using gocache for the build job with #1557. Since everything seems fine, this PR makes `test` and `e2e-test` Make targets download the cache before running tests.

```release-note
NONE
```


xrstf (approved):
/approve

PR #1560, opened 2021-10-04
Cilium

### LINKED ISSUE #462: Add Cilium as a Overlay Network
Cilium as a overlay

Cilium has lot of benefits over a calico / kube-proxy setup, described in detail here: https://cilium.readthedocs.io/en/stable/intro/.
For me the 3 outstanding features are:
- FQDN support for networkpolicies 
- BPF Support
- Removal of kube-proxy and iptables
- Much better monitoring / insights

WeirdMachine:

Fixes #462

Support for hubble and kube-proxy free deployments

```release-note
Support Cilium CNI
```


pkg/apis/kubeone/types.go
```diff
@@ -301,6 +303,15 @@ type CanalSpec struct {
 	MTU int `json:"mtu,omitempty"`
 }
 
+// CiliumSpec defines the Cilium CNI plugin
+type CiliumSpec struct {
+	// Disable kube-proxy to use Cilium eBPF replacement
+	// default value is false
```
xmudrii, line 309: Please make this a valid Go comment and format it a little bit better.

```suggestion
	// DisableKubeProxy disables kube-proxy to use Cilium eBPF replacement.
	// Default value is false
```

pkg/apis/kubeone/types.go
```diff
@@ -301,6 +303,15 @@ type CanalSpec struct {
...
+	DisableKubeProxy bool `json:"disableKubeProxy,omitempty"`
```
xmudrii, line 310: This sort of conflicts with the KubeProxy struct that we have in the ClusterNetworkConfig struct. For example, what if someone configures kube-proxy in IPVS mode via ClusterNetworkConfig, but disable it here? We should have a validation rule to reject KubeOneCluster config if this option is enabled, but the kube-proxy configuration is provided.

I'm also wondering should we put this option in the KubeProxyConfig struct instead. It might make sense to have all kube-proxy options grouped in one place. We could eventually have a validation rule to reject disabling kube-proxy if CNI is not Cilium.

pkg/apis/kubeone/types.go
```diff
@@ -301,6 +303,15 @@ type CanalSpec struct {
...
+	// Weather to deploy Hubble relay and UI
```
xmudrii, line 311: Please make this a valid Go comment.

```suggestion
	// EnableHubble is used to deploy Hubble relay and UI
```

pkg/scripts/kubeadm.go
```diff
@@ -44,7 +44,7 @@ var (
 			exit 0;
 		fi
 
-		sudo kubeadm {{ .VERBOSE }} init --config={{ .WORK_DIR }}/cfg/master_{{ .NODE_ID }}.yaml
+		sudo kubeadm {{ .VERBOSE }} init --skip-phases={{ .SKIP_PHASE }} --config={{ .WORK_DIR }}/cfg/master_{{ .NODE_ID }}.yaml
```
xmudrii, line 47: I think we should skip providing this flag if it's empty to be on the safe side if kubeadm changes how's this flag handled someday (e.g. if they change the default value). The simplest way would be just to check if this field is not empty, although, there might be more elegant ways as well. I think we can go with the simplest way for now.

```suggestion
		sudo kubeadm {{ .VERBOSE }} init {{ if .SKIP_PHASE }}--skip-phases={{ .SKIP_PHASE }}{{ end}} --config={{ .WORK_DIR }}/cfg/master_{{ .NODE_ID }}.yaml
```

examples/terraform/gce/variables.tf
```diff
@@ -90,7 +90,7 @@ variable "control_plane_volume_size" {
 }
 
 variable "control_plane_image_family" {
-  default     = "ubuntu-1804-lts"
+  default     = "ubuntu-2004-lts"
```
xmudrii, line 93: This is a breaking change because it might cause all instances to be recreated for those who don't bind the image in `terraform.tfvars`. If this is required, I suggest putting a warning in the release note.

xmudrii (changes requested):
Overall looks great to me, but I have a couple of comments.

examples/terraform/gce/variables.tf
WeirdMachine: Okay i put this into another PR than.

pkg/apis/kubeone/types.go
WeirdMachine: I agree on the Validation.

We could to the following:
1. It is possible to skip installation of kube-proxy for any reason.
2. Cilium needs to be configured in order to run in "without kube-proxy mode"

We need to add proper validation for this as well and it increases complexity a bit. In general i am fine to do this, however not sure if this is needed.

xmudrii: I'm okay with going with the first approach — allow disabling kube-proxy for any reason regardless of the CNI. Out of curiosity, what would happen if someone disables kube-proxy and deploy Cilium, but then afterward enable kube-proxy? Would that break the cluster?

examples/terraform/gce/variables.tf
xmudrii: It's okay to leave the change in this PR, just make sure the release note reflects the change.

pkg/apis/kubeone/types.go
WeirdMachine: Works fine if cilium gets configured to use kube-proxy again. But if you go the other way it will break the cluster. When kube-proxy got installed once it creates iptable rules that require cleanup on each node. For more information: https://docs.cilium.io/en/v1.10/gettingstarted/kubeproxy-free/

PR #1597, opened 2021-10-27
Add CSI driver for AWS EBS

### LINKED ISSUE #1565: Add a CSI driver addon for AWS

embik:
Adds the CSI driver for AWS EBS volumes to kubeone, offering an out-of-tree alternative to the built-in driver. 

Fixes #1565 

- [The documentation for the CSI driver says that the csi-snapshotter needs to be installed as a prerequisite](https://github.com/kubernetes-sigs/aws-ebs-csi-driver#prerequisites) if the snapshotting feature is supposed to be used. It's not included in the YAML files published by the CSI driver. Should we embed it into this addon?
- The resources have been adapted to a) run on the control plane nodes only and b) use the system's identity. I can change this to using the AWS credentials already passed to machine-controller.

```release-note
Support for AWS EBS via CSI driver added
```


xmudrii:
/assign
I’ll take a look and test this on Monday.

xmudrii:
> The documentation for the CSI driver says that the csi-snapshotter needs to be installed as a prerequisite if the snapshotting feature is supposed to be used. It's not included in the YAML files published by the CSI driver. Should we embed it into this addon?

That's not all. The docs state that we should also enable a bunch of feature gates on Kubelet and API server. Please take care of those feature gates as well.

Regarding the CSI snapshotter, we deploy it for some other providers, so I think it makes sense to do it for AWS too. Take a look at Azure or vSphere CSI addons as an example.

> The resources have been adapted to a) run on the control plane nodes only and b) use the system's identity. I can change this to using the AWS credentials already passed to machine-controller.

I think this is fine.

As a side note, the problem with hostnames for the worker nodes will be resolved by https://github.com/kubermatic/machine-controller/pull/1087.

embik:
@xmudrii the feature gates listed there should be all enabled by default in the Kubernetes versions we support. I assume the documentation lists them for completeness' sake or hasn't been updated:

https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

- `CSINodeInfo` enabled by default since 1.14 as Beta (GA since 1.17)
- `CSIDriverRegistry` enabled by default since 1.14 as Beta (GA since 1.18)
- `CSIBlockVolume` enabled by default since 1.14 as Beta (GA since 1.18)
- `VolumeSnapshotDataSource` enabled by default since 1.17 as Beta (GA since 1.20)

So I think we don't need to change anything about the feature flags.

xmudrii (approved):

PR #1606, opened 2021-11-01
Add flag for API endpoint alternative names

ahmedwaleedmalik:
Add flag for API endpoint alternative names

Fixes #

```release-note
New flag `--api-endpoint-alternative-names` for kubeone
```

### DIFF
--- a/pkg/cmd/config.go
+++ b/pkg/cmd/config.go
@@ -57,9 +57,9 @@ type printOpts struct {
 
 	ControlPlaneHosts string `longflag:"control-plane-hosts"`
 
-	APIEndpointHost             string `longflag:"api-endpoint-host"`
-	APIEndpointPort             int    `longflag:"api-endpoint-port"`
-	APIEndpointAlternativeNames string `longflag:"api-endpoint-alternative-names"`
+	APIEndpointHost             string   `longflag:"api-endpoint-host"`
+	APIEndpointPort             int      `longflag:"api-endpoint-port"`
+	APIEndpointAlternativeNames []string `longflag:"api-endpoint-alternative-names"`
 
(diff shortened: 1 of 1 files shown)


pkg/cmd/config.go
```diff
@@ -871,7 +877,8 @@ addons:
 # The API server can also be overwritten by Terraform. Provide the
 # external address of your load balancer or the public addresses of
 # the first control plane nodes.
-# apiEndpoint:
+apiEndpoint:
+  alternativeNames: {{ .APIEndpointAlternativeNames }}
```
xmudrii, line 881: Can we have it commented like host and port? Also, very tiny nit, but I think it might make sense to put this after host and port because those two fields are more important.

xmudrii (changes requested):

pkg/cmd/config.go
ahmedwaleedmalik: Done

pkg/cmd/config.go
```diff
@@ -874,6 +880,7 @@ addons:
 # apiEndpoint:
 #   host: '{{ .APIEndpointHost }}'
 #   port: {{ .APIEndpointPort }}
+#	alternativeNames: {{ .APIEndpointAlternativeNames }}
```
xmudrii, line 883: Looks like indentation doesn't match host and port fields.

xmudrii (changes requested):
Just one more nit.

pkg/cmd/config.go
ahmedwaleedmalik: Fixed

xmudrii (approved):

kron4eg (approved):
/approve

PR #1607, opened 2021-11-01
Enable ip_tables modules and disable nm-cloud-setup for rhel on aws

### LINKED ISSUE #1573: Networking (CNI) is broken on newly-created clusters running RHEL 8.3 and newer
While doing pre-release tests (#1478), we found out that networking (CNI) is broken on newly-created clusters running RHEL 8.3 and newer. Some more details can be found in #1514. We should invest some time to investigate and eventually fix this.

moadqassem:
Signed-off-by: Moath Qasim <moad.qassem@gmail.com>

Enable ip_tables related kernel modules and disable nm-cloud-setup tool on aws for rhel machines  in order to fix canal support in KubeOne for RHEL machines on AWS. 

Fixes #1573

```release-note
Enable ip_tables related kernel modules and disable nm-cloud-setup tool on aws for rhel machines 
```

### DIFF
--- a/pkg/scripts/funcs.go
+++ b/pkg/scripts/funcs.go
@@ -57,6 +57,7 @@ var (
 		EOF
 		sudo modprobe overlay
 		sudo modprobe br_netfilter
+        sudo modprobe ip_tables
 		sudo mkdir -p /etc/sysctl.d
 		cat <<EOF | sudo tee /etc/sysctl.d/k8s.conf
 		fs.inotify.max_user_watches         = 1048576
@@ -78,10 +79,6 @@ var (
 		SystemMaxUse=5G
 		EOF
 		sudo systemctl force-reload systemd-journald
-		{{ end }}
-
-		{{ define "load-iptables-modules" }}
-		sudo modprobe ip_tables
 		{{ end }}
 	`)
 )
--- a/pkg/scripts/funcs.go
+++ b/pkg/scripts/funcs.go
@@ -79,6 +79,10 @@ var (
 		EOF
 		sudo systemctl force-reload systemd-journald
 		{{ end }}
+
+        {{ define "load-iptables-modules" }}
+        modprobe ip_tables
+        {{ end }}
 	`)
 )
 
--- a/pkg/scripts/funcs.go
+++ b/pkg/scripts/funcs.go
@@ -80,9 +80,9 @@ var (
 		sudo systemctl force-reload systemd-journald
 		{{ end }}
 
-        {{ define "load-iptables-modules" }}
-        modprobe ip_tables
-        {{ end }}
+		{{ define "load-iptables-modules" }}
+		sudo modprobe ip_tables
+		{{ end }}
 	`)
 )
 
--- a/pkg/scripts/funcs.go
+++ b/pkg/scripts/funcs.go
@@ -54,10 +54,11 @@ var (
 		cat <<EOF | sudo tee /etc/modules-load.d/containerd.conf
 		overlay
 		br_netfilter
+		ip_tables
 		EOF
 		sudo modprobe overlay
 		sudo modprobe br_netfilter
-        sudo modprobe ip_tables
+		sudo modprobe ip_tables
 		sudo mkdir -p /etc/sysctl.d
 		cat <<EOF | sudo tee /etc/sysctl.d/k8s.conf
 		fs.inotify.max_user_watches         = 1048576
--- a/pkg/scripts/os_centos.go
+++ b/pkg/scripts/os_centos.go
@@ -20,6 +20,9 @@ import "k8c.io/kubeone/pkg/apis/kubeone"
 
 const (
 	kubeadmCentOSTemplate = `
+{{ if .LOAD_IP_TABLES }}
+{{ template "load-iptables-modules" }}
+{{ end }}
 sudo swapoff -a
 sudo sed -i '/.*swap.*/d' /etc/fstab
 sudo setenforce 0 || true
@@ -102,6 +105,14 @@ sudo yum remove -y \
 	kubectl
 sudo yum remove -y kubernetes-cni || true
 `
+	disableNMCloudSetup = `
+if systemctl status 'nm-cloud-setup.timer' 2> /dev/null | grep -Fq "Active:"; then
+systemctl stop nm-cloud-setup.timer
+systemctl disable nm-cloud-setup.service
+systemctl disable nm-cloud-setup.timer
+reboot
+fi
+`
 )
 
 func KubeadmCentOS(cluster *kubeone.KubeOneCluster, force bool) (string, error) {
@@ -111,6 +122,7 @@ func KubeadmCentOS(cluster *kubeone.KubeOneCluster, force bool) (string, error) 
 	}
 
 	return Render(kubeadmCentOSTemplate, Data{
+		"LOAD_IP_TABLES":         isAWSCloudProvider(&cluster.CloudProvider),
 		"KUBELET":                true,
 		"KUBEADM":                true,
 		"KUBECTL":                true,
@@ -136,6 +148,7 @@ func UpgradeKubeadmAndCNICentOS(cluster *kubeone.KubeOneCluster) (string, error)
 	}
 
 	return Render(kubeadmCentOSTemplate, Data{
+		"LOAD_IP_TABLES":         isAWSCloudProvider(&cluster.CloudProvider),
 		"UPGRADE":                true,
 		"KUBEADM":                true,
 		"KUBERNETES_VERSION":     cluster.Versions.Kubernetes,
@@ -155,6 +168,7 @@ func UpgradeKubeletAndKubectlCentOS(cluster *kubeone.KubeOneCluster) (string, er
 	}
 
 	return Render(kubeadmCentOSTemplate, Data{
+		"LOAD_IP_TABLES":         isAWSCloudProvider(&cluster.CloudProvider),
 		"UPGRADE":                true,
 		"KUBELET":                true,
 		"KUBECTL":                true,
@@ -166,4 +180,16 @@ func UpgradeKubeletAndKubectlCentOS(cluster *kubeone.KubeOneCluster) (string, er
 		"INSTALL_DOCKER":         cluster.ContainerRuntime.Docker,
 		"INSTALL_CONTAINERD":     cluster.ContainerRuntime.Containerd,
 	})
+}
+
+func DisableNMCloudSetup() (string, error) {
+	return Render(disableNMCloudSetup, nil)
+}
+
+func isAWSCloudProvider(spec *kubeone.CloudProviderSpec) bool {
+	if spec.AWS != nil {
+		return true
+	}
+
+	return false
 }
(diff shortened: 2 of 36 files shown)


pkg/scripts/funcs.go
```diff
@@ -79,6 +79,10 @@ var (
 		EOF
 		sudo systemctl force-reload systemd-journald
 		{{ end }}
+
+        {{ define "load-iptables-modules" }}
+        modprobe ip_tables
+        {{ end }}
```
xmudrii, line 85: Indentation doesn't match other functions.

pkg/scripts/funcs.go
```diff
@@ -79,6 +79,10 @@ var (
 		EOF
 		sudo systemctl force-reload systemd-journald
 		{{ end }}
+
+        {{ define "load-iptables-modules" }}
+        modprobe ip_tables
```
xmudrii, line 84: Why not `sudo modprobe`? We usually SSH as a non-root user and then use `sudo` (see other commands).

pkg/scripts/os_centos.go
```diff
@@ -111,6 +122,7 @@ func KubeadmCentOS(cluster *kubeone.KubeOneCluster, force bool) (string, error)
 	}
 
 	return Render(kubeadmCentOSTemplate, Data{
+		"LOAD_IP_TABLES":         isAWSCloudProvider(&cluster.CloudProvider),
```
xmudrii, line 125: Why don't we do it regardless of the provider? I guess it shouldn't cause problems, but we can encounter other providers or images that don't have the `iptables` module enabled.

pkg/tasks/prerequisites.go
```diff
@@ -111,6 +111,28 @@ func createEnvironmentFile(s *state.State) error {
...
+				if err != nil {
+					return err
+				}
+
+				_, _, err = s.Runner.RunRaw(cmd)
+
+				return err
+			}
+		}
+	}
```
xmudrii, line 131: I'm pretty sure this is not what you want. This function is executing provided commands only on one specific node, which is provided via the `node` parameter.

You are going through all nodes, and for each node that's not initialized, run commands only on that one node. With this approach, if you have 2 initialized nodes and 1 uninitialized, you're going to run commands on initialized nodes as well.

What you need to do is to match the node you receive as a parameter to the node from `allHosts` array, and check if only that node is provisioned. If it isn't, run the commands and that's it.

Also, similar to the earlier comment, why don't we run this for all providers?

pkg/tasks/tasks.go
```diff
@@ -107,6 +107,13 @@ func WithFullInstall(t Tasks) Tasks {
 	return WithBinariesOnly(t).
 		append(kubernetesConfigFiles()...).
 		append(Tasks{
+			{
+				Fn: func(s *state.State) error {
+					s.Logger.Infoln("Disable nm-cloud-setup on aws rhel nodes...")
+					return s.RunTaskOnLeader(disableNMCloudSetup)
+				},
+				ErrMsg: "failed to disable nm-cloud-setup",
+			},
```
xmudrii, line 116: As discussed earlier, I'd run this at the very early beginning, even before installing binaries.

xmudrii (changes requested):

pkg/tasks/prerequisites.go
moadqassem: > check if only that node is provisioned 

Provisioned or Initialized ? 

> Also, similar to the earlier comment, why don't we run this for all providers? 
because this service runs only on aws. We can however enable it for cloud providers

pkg/tasks/tasks.go
xmudrii: @moadqassem Seems like this is a leftover that should be removed.

xmudrii (changes requested):
One more nit, otherwise LGTM.

pkg/tasks/tasks.go
```diff
@@ -104,7 +104,15 @@ func WithHostnameOSAndProbes(t Tasks) Tasks {
 // WithFullInstall with install binaries (using WithBinariesOnly) and
 // orchestrate complete cluster init
 func WithFullInstall(t Tasks) Tasks {
-	return WithBinariesOnly(t).
+	return Tasks{
+		{
+			Fn: func(s *state.State) error {
+				s.Logger.Infoln("Disable nm-cloud-setup on aws rhel nodes...")
```
kron4eg, line 110: this task will run unconditionally, which means this log message will appear even if there are no RHEL machines on AWS provider

moadqassem: So best is, making it generic. Will change it to Disable 'nm-cloud-setup tool...'

pkg/tasks/tasks.go
```diff
@@ -104,7 +104,15 @@ func WithHostnameOSAndProbes(t Tasks) Tasks {
 // WithFullInstall with install binaries (using WithBinariesOnly) and
 // orchestrate complete cluster init
 func WithFullInstall(t Tasks) Tasks {
-	return WithBinariesOnly(t).
+	return Tasks{
+		{
+			Fn: func(s *state.State) error {
+				s.Logger.Infoln("Disable nm-cloud-setup nodes...")
+				return s.RunTaskOnLeader(disableNMCloudSetup)
```
kron4eg, line 111: why only on a leader? aren't other nodes need this to be disabled?

kron4eg: Oh, I see why, because of the abuse of this and iterating over all nodes inside the OnLeader task....

Well, we have `s.RunTaskOnAllNodes()` for this

moadqassem: But the current node object that we receive on the task doesn't have a property which tells you this node has been initialized or not. And this is quite important.

kron4eg: initialized status can be found by ID in `(*state.State).LiveCluster`, just like you did in `disableNMCloudSetup`.

xmudrii: This should be s.RunTaskOnAllNodes(), otherwise the service will be disabled only on the leader (first) node.

pkg/tasks/tasks.go
```diff
@@ -104,7 +104,15 @@ func WithHostnameOSAndProbes(t Tasks) Tasks {
 // WithFullInstall with install binaries (using WithBinariesOnly) and
 // orchestrate complete cluster init
 func WithFullInstall(t Tasks) Tasks {
-	return WithBinariesOnly(t).
+	return Tasks{
+		{
+			Fn: func(s *state.State) error {
+				s.Logger.Infoln("Disable nm-cloud-setup nodes...")
```
xmudrii, line 110: Let’s move this logger to the executor function (disableNMCloudSetup) and only show it if we are about to execute the command. Also, I’d include some additional details such as: “Disabling nm-cloud-setup… The node will be rebooted…”.

xmudrii (changes requested):

pkg/scripts/funcs.go
```diff
@@ -54,9 +54,11 @@ var (
 		cat <<EOF | sudo tee /etc/modules-load.d/containerd.conf
 		overlay
 		br_netfilter
+		ip_tables
 		EOF
 		sudo modprobe overlay
 		sudo modprobe br_netfilter
+        sudo modprobe ip_tables
```
kron4eg, line 61:
```suggestion
	sudo modprobe ip_tables
```

pkg/tasks/tasks.go
```diff
@@ -104,7 +104,14 @@ func WithHostnameOSAndProbes(t Tasks) Tasks {
...
 func WithFullInstall(t Tasks) Tasks {
-	return WithBinariesOnly(t).
+	return Tasks{
+		{
+			Fn: func(s *state.State) error {
+				return s.RunTaskOnAllNodes(disableNMCloudSetup, state.RunParallel)
+			},
+			ErrMsg: "failed to disable nm-cloud-setup",
+		},
+	}.append(WithBinariesOnly(t)...).
```
kron4eg, line 114: sooo... this would work.

Inside disableNMCloudSetup you are checking the OS of the node and if it was initialized (in liveCluster), but non of those tasks has been run at this point since `WithBinariesOnly()` has inside it `WithHostnameOSAndProbes`, so basically WithHostnameOSAndProbes is a prerequisite to `disableNMCloudSetup`

kron4eg: The result will be, the RHEL specific script from `disableNMCloudSetup` will never run.

moadqassem: Regarding the `IsInitialized` that's ok, since the value is false and that's what I want. if the node is not initiaized then run the script. about the os, do you really need a task to know what OS is running ? don't you have this information already from tf?

pkg/tasks/prerequisites.go
```diff
@@ -111,6 +111,30 @@ func createEnvironmentFile(s *state.State) error {
...
+	var allHosts = s.LiveCluster.ControlPlane
+	allHosts = append(allHosts, s.LiveCluster.StaticWorkers...)
+	for _, host := range allHosts {
+		if node.ID == host.Config.ID && !host.Initialized() {
+			cmd, err := scripts.DisableNMCloudSetup()
+			if err != nil {
+				return err
+			}
+
+			s.Logger.Infoln("Disable nm-cloud-setup...the node will be rebooted…")
```
xmudrii, line 128: Please use three dots instead of three dots as a single character. We got some reports that it doesn't render well in the past.

```suggestion
			s.Logger.Infoln("Disable nm-cloud-setup... the node will be rebooted...")
```

pkg/tasks/tasks.go
xmudrii: @kron4eg is right. You check the operating system in `pkg/tasks/prerequisites.go` on L115, so your function is not going to execute if you don't determine the OS first. We have two options:

1) Run `WithBinaries` before running `disableNMCloudSetup`
2) Replace `WithBinaries` in a way that we first run `WithHostnameOSAndProbes`, then `disableNMCloudSetup`, and at the end `installPrerequisites`

If I'm not mistaken, you mentioned once that the first approach might not always work correctly, so if that's the case, the second approach would be better.

xmudrii (changes requested):

pkg/tasks/tasks.go
```diff
@@ -104,7 +104,15 @@ func WithHostnameOSAndProbes(t Tasks) Tasks {
...
-	return WithBinariesOnly(t).
+	return WithHostnameOSAndProbes(t).append(Tasks{
+		{
+			Fn: func(s *state.State) error {
+				return s.RunTaskOnAllNodes(disableNMCloudSetup, state.RunParallel)
+			},
+			ErrMsg: "failed to disable nm-cloud-setup",
+		},
+	}...).
+		append(WithBinariesOnly(t)...).
```
xmudrii, line 115: Super tiny nit: this already includes `WithHostnameOSAndProbes`, so the output will look a little bit weird. I propose that  we just do the remaining task, i.e. something like this:

```suggestion
		append(
			Task{Fn: installPrerequisites, ErrMsg: "failed to install prerequisites"},
		).
```

xmudrii (changes requested):
Just one more tiny nit, otherwise we're good to go IMO.

xmudrii (approved):

PR #1609, opened 2021-11-01
Add mirror registries

### LINKED ISSUE #1551: Support for mirror registries

For avoiding to get into bad situations due to download quotas having mirror registries can be used in Docker. I do not know about containerd.

In docker this registries can be configured like this in the `daemon.json` file:
```json
{
  "registry-mirrors": ["https://<my-docker-mirror-host>"]
}
```

It would be great to have a configuration possibility in the `kubeone.yaml` file

On having to install KubeOne multiple times
On having limited network bandwidth

kron4eg:
Fixes #1551 

```release-note
Add mirror registries
```


pkg/apis/kubeone/v1beta1/types.go
```diff
@@ -524,6 +524,8 @@ type RegistryConfiguration struct {
 	// in OverwriteRegistry as an insecure registry. This is also propagated
 	// to the worker nodes managed by machine-controller and/or KubeOne.
 	InsecureRegistry bool `json:"insecureRegistry,omitempty"`
+	// Mirrors sets registry-mirrors
+	Mirrors []string `json:"mirrors"`
```
xmudrii, line 528: What about the control plane nodes? I see that we provide it to machine-controller, but I guess it should affect control plane nodes as well.

kron4eg: Good point

pkg/apis/kubeone/types.go
```diff
@@ -524,6 +524,8 @@ type RegistryConfiguration struct {
 	// in OverwriteRegistry as an insecure registry. This is also propagated
 	// to the worker nodes managed by machine-controller and/or KubeOne.
 	InsecureRegistry bool `json:"insecureRegistry,omitempty"`
+	// Mirrors sets registry-mirrors
```
xmudrii, line 527: Please expand this comment a little bit to include some more details on what are registry mirrors.

pkg/containerruntime/containerd_config.go
```diff
@@ -0,0 +1,118 @@
...
+		criPlugin.Registry.Mirrors[insecureRegistry] = containerdMirror{
+			Endpoint: []string{fmt.Sprintf("http://%s", insecureRegistry)},
+		}
+	}
+
+	if reg := cluster.RegistryConfiguration; reg != nil {
+		if len(reg.Mirrors) > 0 {
+			criPlugin.Registry.Mirrors["docker.io"] = containerdMirror{
+				Endpoint: reg.Mirrors,
+			}
```
xmudrii, line 96: Why don't we support mirrors for registries other than `docker.io`? I assume that Docker doesn't support that, but if containerd does, it could be useful.

pkg/scripts/testdata/TestMigrateToContainerd-simple.golden
```diff
@@ -4,30 +4,6 @@ sudo systemctl stop kubelet
...
-[plugins."io.containerd.grpc.v1.cri".registry]
-[plugins."io.containerd.grpc.v1.cri".registry.mirrors]
-[plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]
-endpoint = ["https://registry-1.docker.io"]
-EOF
-
-cat <<EOF | sudo tee /etc/crictl.yaml
-runtime-endpoint: unix:///run/containerd/containerd.sock
-EOF
-
```
xmudrii, line 30: What happened with this config? Where is it generated now?

pkg/scripts/testdata/TestMigrateToContainerd-insecureRegistry.golden
```diff
@@ -4,32 +4,6 @@ sudo systemctl stop kubelet
...
-[plugins."io.containerd.grpc.v1.cri".registry.mirrors]
-[plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]
-endpoint = ["https://registry-1.docker.io"]
-[plugins."io.containerd.grpc.v1.cri".registry.mirrors."some.registry"]
-endpoint = ["http://some.registry"]
-EOF
-
-cat <<EOF | sudo tee /etc/crictl.yaml
-runtime-endpoint: unix:///run/containerd/containerd.sock
-EOF
```
xmudrii, line 31: Similar to the other comment, what happened with this config?

pkg/scripts/testdata/TestKubeadmFlatcar-simple.golden
```diff
@@ -59,23 +59,6 @@ curl -L https://github.com/kubernetes-sigs/cri-tools/releases/download/${CRI_TOO
...
-	"log-driver": "json-file",
-	"log-opts": {
-		"max-size": "100m"
-	}
-}
-EOF
-
-cat <<EOF | sudo tee /etc/crictl.yaml
-runtime-endpoint: unix:///var/run/dockershim.sock
-EOF
```
xmudrii, line 78: Why don't we provide Docker config for Flatcar any longer?

pkg/scripts/testdata/TestKubeadmFlatcar-with_containerd.golden
```diff
@@ -60,16 +60,7 @@ curl -L https://github.com/kubernetes-sigs/cri-tools/releases/download/${CRI_TOO
...
-cat <<EOF | sudo tee /etc/crictl.yaml
-runtime-endpoint: unix:///run/containerd/containerd.sock
-EOF
 
-sudo mkdir -p /etc/systemd/system/containerd.service.d
-cat <<EOF | sudo tee /etc/systemd/system/containerd.service.d/environment.conf
-[Service]
-Restart=always
-EnvironmentFile=-/etc/environment
-EOF
```
xmudrii, line 72: Why did you remove this?

xmudrii (changes requested):
Please also update the example config file.

pkg/scripts/testdata/TestMigrateToContainerd-simple.golden
kron4eg: All because of missing dot, to pass render context down to the template! :exploding_head:

pkg/containerruntime/containerd_config.go
kron4eg: Indeed, the docker is kinda anchor for such features that we can't supported in both. And to overcome this limitation I wanted to introduce different settings for different runtimes inside new v1beta2 API. So that' something to do in the future.

kron4eg:
@xmudrii OK, addressed all the comments

pkg/containerruntime/containerd_config.go
xmudrii: I think that we can easily overcome this limitation in the current API version as well. Let's say we have a struct such as:

```go
type RegistryMirror struct {
    Registry string
    Mirror string
}

...
  Mirrors []RegistryMirror
```

You can easily extend the validation to reject the config if the container runtime is Docker and there's a registry mirror that's not `docker.io`.

kron4eg:
@xmudrii ok, seems like NOW it's good

xmudrii:
@kron4eg Have you seen this comment: https://github.com/kubermatic/kubeone/pull/1609#discussion_r746432279?

pkg/containerruntime/containerd_config.go
kron4eg: Point is, I don’t want to change v1beta1 even more, given that we are about to have v1beta2

xmudrii: But we are already changing the v1beta1 API. I'm a little bit conflicted on providing two different versions of a feature in different API versions, but in the same release.

It would make total sense if we introduced this in some earlier release and then decided to improve it in the new API version. In that case, leaving v1beta1 API unchanged would probably be the best. However, in this concrete case, we're introducing a new feature in two APIs at the same time (v1beta1 and v1beta2), but we don't provide the same feature set. That doesn't make much sense to me and it can be confusing to users too. Such cases also make migration to v1beta2 more complicated.

I would prefer either leaving this only for v1beta2 or doing what I proposed above (I think I would prefer this approach).

kron4eg: Leaving this for v1beta2 is totally OK and I'd prefer that. In v1beta2 we should delete RegistryConfiguration and deconstruct to two parts inside `ContainerRuntimeContainerd` and `ContainerRuntimeDocker`, each with own fields and possibilities.

kron4eg: However I don't want to waste non-API related changes from this PR (refactoring of how whole container-runtime config generation works) and will repackage it in another PR.

xmudrii: I'm totally fine with leaving this for v1beta2. :+1: 

We discussed it already earlier -- you can just update the internal API version for now, and leave v1beta1 untouched. You don't even need to change the API much, we can always change the internal API.

kron4eg:
OK, after considering the required efforts to extract only refactoring parts of this PR I think we can wait for the new v1beta2 API.

/close

PR #1614, opened 2021-11-08
add clustername to openstack cloud provider

### LINKED ISSUE #1613: Cluster Name not Configured in Openstack OCCM
Wen creating multiple kubeone cluster in one openstack tenant. We could run i a load balancer issue, if we use the same namespace/service name combination in both.

Then both LB have the same name and will be grabbed by both cloud provider

There is a option int the openstack controller `--cluster-name` which would solve that problem, as cluster name is part of the lb name.

This is already implemented for azure, aws and the cinder provider.

So we would need this for the cloud provider too.

qeqar:
Closes #1613

### DIFF
--- a/addons/ccm-openstack/ccm-openstack.yaml
+++ b/addons/ccm-openstack/ccm-openstack.yaml
@@ -192,6 +192,7 @@ spec:
             - --cloud-provider=openstack
             - --use-service-account-credentials=true
             - --bind-address=127.0.0.1
+            - --cluster-name={{ .Config.Name }}
 {{ if .Config.CABundle }}
           env:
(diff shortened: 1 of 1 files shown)


qeqar:
Change is probably breaking as the it will create new LBs for existing cluster?

xmudrii:
/hold
This is a breaking change, but there’s a way to workaround it. @qeqar Would you mind me taking over this, I might have some idea how to fix it so it doesn’t break existing clusters?

qeqar:
As long as it not surpass my Go knowledge i can make it.

xmudrii:
Thank you for reporting the issue, @qeqar!
Suppressed by #1619
/close

PR #1619, opened 2021-11-12
Provide --cluster-name flag to the OpenStack external CCM

### LINKED ISSUE #1613: Cluster Name not Configured in Openstack OCCM

xmudrii:

The OpenStack external CCM takes the `--cluster-name` flag which is used as a prefix for cloud resources. Currently, we're not providing this flag, so the CCM uses the default value which is `kubernetes`. This can cause conflicts if there are multiple clusters in the same tenant, further causing some resources (e.g. Load Balancers) to be reused across clusters.

For example, let's look at Octavia Load Balancers. If you create a Service named `nginx` in the `default` namespace, the Octavia Load Balancer would be named as `kube_service_kubernetes_default_nginx`.

If you create the same service in a different cluster running in the same tenant, OpenStack will just update the Load Balancer created earlier, instead of creating a new LB or failing. The Load Balancer would now point to the new cluster until the Service in the first cluster is not reconciled again (reconciliations happen automatically periodically). Once the Service in the first/old cluster is reconciled, the Load Balancer would now point to the first/old cluster until the Service in the new cluster is not reconciled again, and so on.

This is can be quite hard to debug and detect. The solution is to provide a unique value using the `--cluster-name` flag. In this case, the cloud resources (e.g. Load Balancer) would be named properly.

However, you can't easily change the `--cluster-name` flag in a cluster that's already running. Doing that would cause OpenStack CCM to lose track of all Load Balancers, so all LBs would be recreated. That can be quite problematic and cause users' applications to go down. Therefore we need a backwards compatible approach.

The cluster name is determined in the following way:

* if the cluster is not provisioned, or if the cluster is not an OpenStack cluster, return the KubeOne cluster name
* if it's an existing OpenStack cluster:
    * if cluster is running in-tree cloud provider: return the KubeOne cluster name because the in-tree provider already has the `--cluster-name` flag set
    * if cluster is running external cloud provider: check if there is `--cluster-name` flag on the OpenStack CCM. If there is, read the value and return it, otherwise don't set the OpenStack cluster name, in which case it defaults to `kubernetes`
* if cluster is migrated to external CCM, return the KubeOne cluster name

If the operator wants to set the `--cluster-name` for clusters that don't have it set, they can change the OpenStack CCM DaemonSet to set the flag manually. It's recommended to set the value to the cluster name provided in the KubeOneCluster manifest, but it's not required. When running `kubeone apply`, KubeOne will read the value provided to the DaemonSet and will use it when reconciling the CCM deployment. Note that doing this will cause OpenStack CCM to effectively recreate all LBs. The old LBs need to be deleted manually.

Fixes #1613

```release-note
Provide --cluster-name flag to the OpenStack external CCM (read PR description for more details)
```

/assign @kron4eg 
cc: @qeqar

### DIFF
--- a/pkg/addons/applier.go
+++ b/pkg/addons/applier.go
@@ -58,6 +58,7 @@ type templateData struct {
 	Config                              *kubeoneapi.KubeOneCluster
 	Certificates                        map[string]string
 	Credentials                         map[string]string
+	CCMClusterName                      string
 	CSIMigration                        bool
 	CSIMigrationFeatureGates            string
 	MachineControllerCredentialsEnvVars string
@@ -155,6 +156,7 @@ func newAddonsApplier(s *state.State) (*applier, error) {
 			"KubernetesCA":                 mcCertsMap[resources.KubernetesCACertName],
 		},
 		Credentials:                         creds,
+		CCMClusterName:                      s.LiveCluster.CCMClusterName,
 		CSIMigration:                        csiMigration,
 		CSIMigrationFeatureGates:            csiMigrationFeatureGates,
 		MachineControllerCredentialsEnvVars: string(credsEnvVars),
--- a/pkg/tasks/probes.go
+++ b/pkg/tasks/probes.go
@@ -47,7 +47,8 @@ const (
 
 	kubeletInitializedCMD = `test -f /etc/kubernetes/kubelet.conf`
 
-	k8sAppLabel = "k8s-app"
+	k8sAppLabel               = "k8s-app"
+	openstackCCMAppLabelValue = "openstack-cloud-controller-manager"
 )
 
 var KubeProxyObjectKey = dynclient.ObjectKey{
@@ -157,6 +158,14 @@ func runProbes(s *state.State) error {
 			return err
 		}
 	}
+
+	clusterName, cnErr := detectClusterName(s)
+	if cnErr != nil {
+		return errors.Wrap(cnErr, "failed to detect the ccm --cluster-name flag value")
+	}
+	s.LiveCluster.Lock.Lock()
+	s.LiveCluster.CCMClusterName = clusterName
+	s.LiveCluster.Lock.Unlock()
 
 	switch {
 	case s.Cluster.ContainerRuntime.Containerd != nil:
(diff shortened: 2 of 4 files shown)


pkg/tasks/probes.go
```diff
@@ -636,3 +650,76 @@ func detectCCMMigrationStatus(s *state.State) (*state.CCMStatus, error) {
...
+// The function works in the following way:
+//   * do nothing if the cluster is not an OpenStack cluster
+//   * if it's a new cluster, return the KubeOne cluster name
+//   * if it's an existing cluster:
+//      * if cluster is running in-tree cloud provider: return the KubeOne cluster name
+//      * if cluster is running external cloud provider: check is there
+//        `--cluster-name flag on the OpenStack CCM. If there is, read the
+//        value and use it, otherwise return empty string (defaults to
+//        `kubernetes`)
+//   * if cluster is migrated to external CCM, use the KubeOne cluster name
```
embik, line 675: Nitpick: This documentation of `detectOpenStackClusterName` doesn't seem to be in line with the actual's function return type, which is `error`. It suggests - at least to me - that this returns a string.

embik (changes requested):

pkg/tasks/probes.go
xmudrii: @embik Good point, I actually thought about it as well, but I was too lazy to change it. PTAL now.

pkg/addons/applier.go
```diff
@@ -157,6 +165,7 @@ func newAddonsApplier(s *state.State) (*applier, error) {
 		Credentials:                         creds,
 		CSIMigration:                        csiMigration,
 		CSIMigrationFeatureGates:            csiMigrationFeatureGates,
+		OpenStackCCMClusterName:             openstackClusterName,
```
kron4eg, line 168: I really would like to see this function to stay provider-agnostic. Such functions probably should be part of LiveCluster, and we may even expose it to the templates and have some method like `LiveCluster.CCMStatus.ComputedClusterName()`

xmudrii:
/hold
this needs to be properly tested

kron4eg (approved):

kron4eg (approved):

kron4eg (approved):

PR #1627, opened 2021-11-17
Quality of Life (QOL) changes

xmudrii:

* Replace critical-pod annotation with priorityClassName
    * The critical-pod annotation is not-functional since Kubernetes 1.16
* Show warning about LBs on CCM migration for OpenStack clusters
    * Triggering migration might cause all LBs to be recreated if Neutron LBs are used
    * This also requires updating docs
* Force drain nodes to remove standalone pods
    * If there's a standalone pod (not managed by ReplicaSet) on a control plane or static worker node, draining is going to fail. Force draining node also removes the standalone pods
* Check for minor version when choosing kubeadm API version
     * The constraint-based check doesn't work properly if there's any additional info in the version (alpha/beta/rc or EKS-D versions)
* Create MachineDeployments only for newly-provisioned clusters
    * MachineDeployments should be created only when provisioning the cluster for the first time. Right now, if you add new static worker nodes after creating the cluster, MachineDeployments will be reconciled again (this shouldn't happen)

```release-note
Replace critical-pod annotation with priorityClassName
Show warning about LBs on CCM migration for OpenStack clusters
Force drain nodes to remove standalone pods
Check for minor version when choosing kubeadm API version
Create MachineDeployments only for newly-provisioned clusters
```

/assign @kron4eg

### DIFF
--- a/addons/ccm-digitalocean/ccm-digitalocean.yaml
+++ b/addons/ccm-digitalocean/ccm-digitalocean.yaml
@@ -14,9 +14,8 @@ spec:
     metadata:
       labels:
         app: digitalocean-cloud-controller-manager
-      annotations:
-        scheduler.alpha.kubernetes.io/critical-pod: ''
     spec:
+      priorityClassName: system-cluster-critical
       dnsPolicy: Default
       serviceAccountName: cloud-controller-manager
       tolerations:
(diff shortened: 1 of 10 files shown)


kron4eg (approved):
/approve

PR #1638, opened 2021-11-18
Fix cloud provider name for GCP in default-storage-class addon

dakraus:

The `default-storage-class` addon was ported from KKP and uses the value "gcp" to determine if the Kubernetes cluster uses the GCP cloud provider. However KubeOne uses "gce" as identifier for the GCP cloud provider and therefore the StorageClass was never rendered when provisioning a Kubernetes cluster with KubeOne on GCP.

```release-note
Deploy default StorageClass for GCP clusters if the `default-storage-class` addon is enabled
```


xmudrii (approved):
Great finding!

/approve

Issue #1611, opened 2021-11-04
Create KubeOneCluster v1beta2 API


xmudrii:
Related to #1539.

PR #1643, opened 2021-11-23
Ignore TLS verification for containerd insecure registries

### LINKED ISSUE #1547: Ignore the TLS verification for containerd insecure registries
Enabling the `insecureRegistry` option on the containerd cluster doesn't disable the TLS verification, which can cause errors when trying to pull images from the registry. We should disable the TLS verification for insecure registries by default.

ahmedwaleedmalik:

When using an insecure registry TLS verification should be ignored by default. This behaviour was missing when the user was using an insecure registry for containerd. This PR fixes that isue.

Fixes #1547 

```release-note
Skip TLS verification when using containerd for insecure registries
```

### DIFF
--- a/pkg/scripts/funcs.go
+++ b/pkg/scripts/funcs.go
@@ -151,11 +151,20 @@ type containerdCRIRuncOptions struct {
 }
 
 type containerdCRIRegistry struct {
-	Mirrors map[string]containerdMirror `toml:"mirrors"`
+	Mirrors map[string]containerdRegistryMirror `toml:"mirrors"`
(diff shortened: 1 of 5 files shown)


kron4eg:
Aren't we set `http` prefix to the endpoint, so there there is no TLS involved in insecureRegistry? How this helps then?

ahmedwaleedmalik:
Closing this pull request because it doesn't solve the actual issue. This requires an API change and will be resolved with https://github.com/kubermatic/kubeone/issues/1611

PR #1644, opened 2021-11-23
feat: add container logs max size to the kubeone API

### LINKED ISSUE #1587: Allow logging configuration for Kubelet
Right now it looks like we are not having any option to control the kubelet's configuration for logs size or rotation.

Right now, with containerd, it looks like 10Mi is used a default size. And for Docker it seems to be 100Mi (configured in /etc/docker/daemon.json).

See the related issues:
 * https://github.com/kubernetes/kubernetes/pull/59898
 * https://github.com/kubermatic/kubeone/pull/1180

It would be great if there are any configuration options which can be defined on the kubeone level and properly propagated at target cluster.

Being able to access more history for the long-running pods which has bunch of logs inside.

nerdeveloper:
This allow the user to set container logs maximum size on kubeone and defaults the container logs max size to 100Mi
Fixes #1587

YES

```release-note
Ability to change container log maximum size. Currently defaults to 100Mi
```


pkg/state/context.go
```diff
@@ -103,6 +103,7 @@ type State struct {
 	CredentialsFilePath       string
 	ManifestFilePath          string
 	PauseImage                string
+	ContainerLogsMaxSize      string
```
kron4eg, line 106: please default this in the constructor, otherwise this will be empty during kubeone install for exmaple.

kron4eg (changes requested):

WeirdMachine (changes requested):

pkg/cmd/config.go
```diff
@@ -501,7 +509,10 @@ kind: KubeOneCluster
 name: {{ .ClusterName }}
 
 versions:
-  kubernetes: "{{ .KubernetesVersion }}"
+ kubernetes: "{{ .KubernetesVersion }}"
+
+kubeletConfiguration:
+ containerLogMaxSize: "{{ .ContainerLogMaxSize }}"
```
kron4eg, line 515: since there should be no v1beta1 API change, whole diff in this file should also be gone

pkg/apis/kubeone/v1beta1/types.go
```diff
@@ -63,6 +63,14 @@ type KubeOneCluster struct {
 	AssetConfiguration AssetConfiguration `json:"assetConfiguration,omitempty"`
 	// RegistryConfiguration configures how Docker images are pulled from an image registry
 	RegistryConfiguration *RegistryConfiguration `json:"registryConfiguration,omitempty"`
+	// KubeletConfiguration configures the kubelet
+	KubeletConfiguration KubeletConfiguration `json:"kubeletConfiguration"`
```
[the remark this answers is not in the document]
nerdeveloper: I removed it but for some weird reason when I try to run update-codegen. it fails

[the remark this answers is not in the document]
nerdeveloper: Screenshot 2021-11-25 at 16 23 39
this is the issue I have when I remove it from v1beta1 API.

pkg/cmd/config.go
xmudrii: This is not relevant anymore since the example config uses the v1beta2 API already.

pkg/apis/kubeone/v1beta1/types.go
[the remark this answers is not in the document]
xmudrii: Try adding the types to the v1beta2 API and then re-run codegen.

pkg/apis/kubeone/types.go
```diff
@@ -182,6 +182,9 @@ type KubeletConfig struct {
 	// EvictionHard configure --eviction-hard command-line flag of the kubelet.
 	// See more at: https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/
 	EvictionHard map[string]string `json:"evictionHard,omitempty"`
+	//ContainerLogsMaxSize configure --container-logs-max-size command-line flag of the kubelet. Defaults to 100Mi.
```
xmudrii, line 185:
```suggestion
	// ContainerLogsMaxSize configure --container-log-max-size command-line flag of the kubelet. Defaults to 100Mi.
```

pkg/cmd/config.go
```diff
@@ -501,7 +501,7 @@ kind: KubeOneCluster
 name: {{ .ClusterName }}
 
 versions:
-  kubernetes: "{{ .KubernetesVersion }}"
+ kubernetes: "{{ .KubernetesVersion }}"
```
xmudrii, line 504: This blank space is expected (identation should be 2).

```suggestion
  kubernetes: "{{ .KubernetesVersion }}"
```

Also, please add this new field to the example config.

pkg/templates/kubeadm/v1beta3/kubeadm.go
```diff
@@ -429,6 +429,12 @@ func newNodeRegistration(s *state.State, host kubeoneapi.HostConfig) kubeadmv1be
 		kubeletCLIFlags["eviction-hard"] = kubeoneapi.MapStringStringToString(m, "<")
 	}
 
+	if m := host.Kubelet.ContainerLogMaxSize; m != "" {
+		kubeletCLIFlags["container-log-max-size"] = m
+	} else {
+		kubeletCLIFlags["container-log-max-size"] = "100Mi"
+	}
```
xmudrii, line 436: Can we use API defaulting instead?

xmudrii (changes requested):

pkg/apis/kubeone/types.go
```diff
@@ -63,6 +63,15 @@ type KubeOneCluster struct {
 	AssetConfiguration AssetConfiguration `json:"assetConfiguration,omitempty"`
 	// RegistryConfiguration configures how Docker images are pulled from an image registry
 	RegistryConfiguration *RegistryConfiguration `json:"registryConfiguration,omitempty"`
+	// KubeletConfiguration configures the kubelet
+	KubeletConfiguration KubeletConfiguration `json:"kubeletConfiguration"`
+}
+
+// KubeletConfiguration
```
xmudrii, line 70: Please write a proper comment describing what's this struct used for.

pkg/apis/kubeone/types.go
```diff
@@ -63,6 +63,15 @@ type KubeOneCluster struct {
...
+type KubeletConfiguration struct {
```
xmudrii, line 71: I don't like that this struct is called `KubeletConfiguration` because we already have the `KubeletConfig` struct used for something else. Maybe we can call it `LoggingConfiguration`? @kron4eg WDYT?

pkg/apis/kubeone/types.go
```diff
@@ -63,6 +63,15 @@ type KubeOneCluster struct {
...
+	// ContainerLogsMaxSize configure --container-logs-max-size command-line flag of the kubelet. Defaults to 100Mi.
```
xmudrii, line 72: This comment is not valid any longer because we use the (upstream) KubeletConfig and not command-line flags. You can say something along (copied from [the upstream docs](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/)):

```suggestion
	// ContainerLogMaxSize configures the maximum size of container log file before it is rotated
```

pkg/apis/kubeone/v1beta1/types.go
```diff
@@ -63,8 +63,13 @@ type KubeOneCluster struct {
 	AssetConfiguration AssetConfiguration `json:"assetConfiguration,omitempty"`
 	// RegistryConfiguration configures how Docker images are pulled from an image registry
 	RegistryConfiguration *RegistryConfiguration `json:"registryConfiguration,omitempty"`
+	// KubeletConfiguration configures the kubelet
+	KubeletConfiguration KubeletConfiguration `json:"kubeletConfiguration,omitempty"`
 }
 
+// Kubele
+type KubeletConfiguration struct{}
```
xmudrii, line 71: I don't think we need this at all. There's another way we can handle this.

pkg/apis/kubeone/v1beta2/defaults.go
```diff
@@ -133,7 +133,6 @@ func SetDefaults_ContainerRuntime(obj *KubeOneCluster) {
 		obj.ContainerRuntime.Containerd = &ContainerRuntimeContainerd{}
 	}
 }
-
```
xmudrii, line 136: Please leave this new line.

pkg/cmd/config.go
```diff
@@ -48,8 +48,9 @@ const (
 type printOpts struct {
 	FullConfig bool `longflag:"full" shortflag:"f"`
 
-	ClusterName       string `longflag:"cluster-name" shortflag:"n"`
-	KubernetesVersion string `longflag:"kubernetes-version" shortflag:"k"`
+	ClusterName         string `longflag:"cluster-name" shortflag:"n"`
+	KubernetesVersion   string `longflag:"kubernetes-version" shortflag:"k"`
+	ContainerLogMaxSize string `longflag:"container-log-max-size"`
```
xmudrii, line 53: Very tiny nit: I would put this at the end, so variables are grouped by priority/importance.

pkg/cmd/config.go
```diff
@@ -136,6 +137,13 @@ func configPrintCmd() *cobra.Command {
 		defaultKubernetesVersion,
 		"Kubernetes version")
 
+	cmd.Flags().StringVarP(
+		&opts.ContainerLogMaxSize,
+		longFlagName(opts, "ContainerLogMaxSize"),
+		shortFlagName(opts, "ContainerLogMaxSize"),
+		"100Mi",
+		"ContainerLogMaxSize")
```
xmudrii, line 145: Very tiny nit: I would put this at the end, so variables are grouped by priority/importance.

pkg/cmd/config.go
```diff
@@ -289,6 +297,8 @@ func createAndPrintManifest(printOptions *printOpts) error {
 	// Version
 	cfg.Set(yamled.Path{"versions", "kubernetes"}, printOptions.KubernetesVersion)
 
+	cfg.Set(yamled.Path{"kubeletConfiguration", "ContainerLogMaxSize"}, printOptions.ContainerLogMaxSize)
```
xmudrii, line 300: Very tiny nit: I would put this at the end, so variables are grouped by priority/importance.

pkg/cmd/config.go
```diff
@@ -502,6 +512,8 @@ name: {{ .ClusterName }}
 
 versions:
   kubernetes: "{{ .KubernetesVersion }}"
+kubeletConfiguration:
+ containerLogMaxSize: "{{ .ContainerLogMaxSize }}"
```
xmudrii, line 516: Very tiny nit: I would put this at the end, so variables are grouped by priority/importance.

pkg/templates/kubeadm/v1beta2/kubeadm.go
```diff
@@ -468,3 +470,11 @@ func kubeProxyConfiguration(s *state.State) *kubeproxyv1alpha1.KubeProxyConfigur
...
 	return kubeProxyConfig
 }
+
+func getContainerLogMaxSize(value string) string {
+	if value == "" {
+		return "100Mi"
+	}
+
+	return value
+}
```
xmudrii, line 480: Please use defaulting instead of this function.

pkg/templates/kubeadm/v1beta3/kubeadm.go
```diff
@@ -468,3 +470,11 @@ func kubeProxyConfiguration(s *state.State) *kubeproxyv1alpha1.KubeProxyConfigur
...
 	return kubeProxyConfig
 }
+
+func getContainerLogMaxSize(value string) string {
+	if value == "" {
+		return "100Mi"
+	}
+
+	return value
+}
```
xmudrii, line 480: Please use defaulting instead of this function.

pkg/apis/kubeone/types.go
kron4eg: Indeed, https://docs.kubermatic.com/kubeone/master/references/kubeone_cluster_v1beta2/#kubeletconfig
and we should _use that_ struct to configure log size

xmudrii: @kron4eg and I discussed this and we can't use this struct because it's used for per-host settings, and we don't want to configure logging that way. @nerdeveloper Can you just rename this struct to `LoggingConfig`?

nerdeveloper: Sure

pkg/templates/kubeadm/v1beta2/kubeadm.go
nerdeveloper: done

pkg/cmd/config.go
nerdeveloper: done

pkg/cmd/config.go
nerdeveloper: fixed

pkg/cmd/config.go
nerdeveloper: fixed

pkg/apis/kubeone/v1beta1/types.go
nerdeveloper: done

pkg/templates/kubeadm/v1beta3/kubeadm.go
nerdeveloper: done

pkg/apis/kubeone/types.go
```diff
@@ -63,6 +63,15 @@ type KubeOneCluster struct {
 	AssetConfiguration AssetConfiguration `json:"assetConfiguration,omitempty"`
 	// RegistryConfiguration configures how Docker images are pulled from an image registry
 	RegistryConfiguration *RegistryConfiguration `json:"registryConfiguration,omitempty"`
+	// LoggingConfig configures the Kubelet's log configuration
```
xmudrii, line 66: I think this is a little bit better:

```suggestion
	// LoggingConfig configures the Kubelet's log rotation
```

pkg/apis/kubeone/types.go
```diff
@@ -63,6 +63,15 @@ type KubeOneCluster struct {
...
+	LoggingConfig LoggingConfig `json:"loggingConfig,omitempty"`
+}
+
+// LoggingConfig
```
xmudrii, line 70: Please make this a valid Go comment:

```suggestion
// LoggingConfig configures the Kubelet's log rotation
```

xmudrii (changes requested):
We still need to decide what do we want to do with Docker.

ahmedwaleedmalik:
Should/will this configuration be propagated to the initial machine deployment for workers? @nerdeveloper @xmudrii

xmudrii:
@ahmedwaleedmalik Yes, we should. There's a PR in the machine-controller repo as well: https://github.com/kubermatic/machine-controller/pull/1173

pkg/apis/kubeone/config/config.go
```diff
@@ -237,6 +237,11 @@ func SetKubeOneClusterDynamicDefaults(cluster *kubeoneapi.KubeOneCluster, creden
 		}
 	}
 
+	// check if the container log max size is not set, if not set it
+	if cluster.LoggingConfig.ContainerLogMaxSize == "" {
+		cluster.LoggingConfig.ContainerLogMaxSize = "100Mi"
```
kron4eg, line 242: Defaulting is done in `pkg/apis/kubeone/v1beta2/defaults.go`

pkg/cmd/config.go
```diff
@@ -972,4 +984,7 @@ machineController:
 #     operatingSystem: 'ubuntu'
 #     operatingSystemSpec:
 #       distUpgradeOnBoot: true
+
+loggingConfig:
+ containerLogMaxSize: "{{ .ContainerLogMaxSize }}"
```
kron4eg, line 989:
```suggestion
  containerLogMaxSize: "{{ .ContainerLogMaxSize }}"
```

pkg/containerruntime/docker_config.go
```diff
@@ -32,12 +33,17 @@ type dockerConfig struct {
 }
 
 func marshalDockerConfig(cluster *kubeone.KubeOneCluster) (string, error) {
+	logSize := strings.ToLower(cluster.LoggingConfig.ContainerLogMaxSize)
+	logSize = strings.ReplaceAll(logSize, "ki", "k")
+	logSize = strings.ReplaceAll(logSize, "mi", "m")
```
kron4eg, line 38: ReplaceAll is case sensitive, it will not replace `Mi` with `m`, but instead would simply return `Mi` after which dockerd would freak out.

pkg/apis/kubeone/config/config.go
kron4eg: scratch that, that comment is a mistake.

pkg/apis/kubeone/v1beta2/types.go
```diff
@@ -61,6 +61,15 @@ type KubeOneCluster struct {
 	SystemPackages *SystemPackages `json:"systemPackages,omitempty"`
 	// RegistryConfiguration configures how Docker images are pulled from an image registry
 	RegistryConfiguration *RegistryConfiguration `json:"registryConfiguration,omitempty"`
+	// LoggingConfig configures the Kubelet's log configuration
+	LoggingConfig LoggingConfig `json:"loggingConfig,omitempty"`
+}
+
+// LoggingConfig
```
xmudrii, line 68: Please also extend this comment like in the internal API:

```suggestion
// LoggingConfig configures the Kubelet's log rotation
```

pkg/apis/kubeone/config/config.go
```diff
@@ -237,6 +237,11 @@ func SetKubeOneClusterDynamicDefaults(cluster *kubeoneapi.KubeOneCluster, creden
 		}
 	}
 
+	// check if the container log max size is not set, if not set it
```
xmudrii, line 240: Let's extend this comment.

```suggestion
	// Default ContainerLogMaxSize.
	// NB: We intentionally default here because LoggingConfig is not available in
	// the v1beta1 API. If we would default in the v1beta2 API instead, this value would
	// be empty when converting from v1beta1 to internal. This means that v1beta1 API
	// users would depend on default values provided by Docker/upstream, which are
	// different than our default values, so we want to avoid this.
```

pkg/cmd/config.go
```diff
@@ -370,6 +380,8 @@ func createAndPrintManifest(printOptions *printOpts) error {
 		cfg.Set(yamled.Path{"machineController", "deploy"}, printOptions.DeployMachineController)
 	}
 
+	cfg.Set(yamled.Path{"loggingConfig", "ContainerLogMaxSize"}, printOptions.ContainerLogMaxSize)
```
xmudrii, line 383:
```suggestion
	cfg.Set(yamled.Path{"loggingConfig", "containerLogMaxSize"}, printOptions.ContainerLogMaxSize)
```

pkg/containerruntime/docker_config_test.go
```diff
@@ -0,0 +1,57 @@
...
+				t.Errorf("marshalDockerConfig() error = %v,", err)
+				return
+			}
+			cfg := dockerConfig{}
+			err = json.Unmarshal([]byte(got), &cfg)
+			gotLogSize := cfg.LogOpts["max-size"]
+
+			if err != nil {
+				return
+			}
```
xmudrii, line 45: We shouldn't return error here, but instead do this:

```suggestion
			if err != nil {
				t.Errorf("failed to unmarshal docker config: %v", err)
			}
```

Also, we should do this before setting `gotLogSize`.

pkg/scripts/testdata/TestKubeadmAmazonLinux-force.golden
```diff
@@ -88,7 +88,7 @@ cat <<EOF | sudo tee /etc/docker/daemon.json
 	"storage-driver": "overlay2",
 	"log-driver": "json-file",
 	"log-opts": {
-		"max-size": "100m"
+		"max-size": ""
```
xmudrii, line 91: Can you add the LoggingConfig to the function below?

https://github.com/kubermatic/kubeone/blob/5f11decd23b06d9915af412be5f17fb0d3955467/pkg/scripts/os_test.go#L91-L111

xmudrii (commented):
Some more comments, but we are getting closer and closer.

kron4eg (approved):
/approve

PR #1641, opened 2021-11-22
Add validation for cluster_name

ahmedwaleedmalik:
**What this PR does / why we need it**:
Add validation for `cluster_name` variable

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1630 

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Add validation for `cluster_name` variable
```

PR #1646, opened 2021-11-23
ValidateName always prints empty string as value instead of actual value

cedi:
This commit fixes an rather annoying issue where `ValidateName` does not
print the correct value of the invalid cluster name which made the issue
hard to debug.

Background:
I recently upgraded from an earlier KubeOne version to master and my cluster-name contained dots (`.`).
However since https://github.com/kubermatic/kubeone/pull/1641 this is no longer valid.

The output of this validation looked like

```bash
$ kubeone apply --manifest kubeone.yaml --credentials credentials.yaml --tfjson output.json --auto-approve --verbose

Error: failed to initialize State: failed to build state: failed to load cluster: unable to load a given KubeOneCluster object: unable to validate the given KubeOneCluster object: name: Invalid value: "": .name should be lowercase and can only contain alphanumeric characters and hyphens(-)
```

With this commit in place the output looks like this:

```bash
$ kubeone install --manifest kubeone.yaml --credentials credentials.yaml --tfjson output.json --verbose
Error: failed to initialize State: failed to build state: failed to load cluster: unable to load a given KubeOneCluster object: unable to validate the given KubeOneCluster object: name: Invalid value: "k8s.cedi.dev": .name should be lowercase and can only contain alphanumeric characters and hyphens(-)
```

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/apis/kubeone/validation/validation.go
+++ b/pkg/apis/kubeone/validation/validation.go
@@ -69,7 +69,7 @@ func ValidateName(name string, fldPath *field.Path) field.ErrorList {
 	}
 	clusterNameRegex := regexp.MustCompile(`^[0-9a-z-]+$`)
 	if !clusterNameRegex.MatchString(name) {
-		allErrs = append(allErrs, field.Invalid(fldPath, "",
+		allErrs = append(allErrs, field.Invalid(fldPath, name,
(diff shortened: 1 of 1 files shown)


kron4eg:
Thanks @cedi!

/lgtm
/approve

PR #1648, opened 2021-11-29
Validate cluster name using RFC 1123 regexp

### LINKED ISSUE #1630: Avoid creating resources with uppercase names by validating the inputs
**What happened**:
We have seen the situation that when you setup the KubeOne cluster with terraform (which has the input variable of `cluster_name`, these are used for the various resources but also for the resources inside k8s.

This is not possible due to https://github.com/kubernetes/kubernetes/issues/71140.

Consider validation that uppercase naming is not allowed inside the `cluster_name` to avoid such issues.

Example issue at the end of kubeone apply logs:
```
Creating credentials secret...
...Task failed, error was: nodes "K8sMaster-cp-0" not found
```
While the nodes are:
```
$ kubectl get nodes
k8smaster-cp-0   Ready ...
```

Applied `cluster_name` inside the terraform vars was `K8sMaster`.

* do not allow to provision resources and cluster with this setup

Try to provision cluster with terraform + kubeone while setting `cluster_name` with some uppercase values (reproduced on Azure but I guess it will be the same on other providers).

N/A

KubeOne version (`kubeone version`): 1.3.0
Operating system: 

xmudrii:

#1641 introduced validation for the cluster name to fix #1630. However, the regexp used in that PR is too restrictive because it doesn't allow dots, which are allowed by RFC 1123. Also, the original implementation doesn't cover other requirements defined by RFC 1123:

* contain no more than 253 characters
* start with an alphanumeric character
* end with an alphanumeric character

This PR uses the regexp used by Kubernetes to [validate RFC 1123 subdomains](https://github.com/kubernetes/kubernetes/blob/9a75e7b0fd1b567f774a3373be640e19b33e7ef1/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L209-L218).

xref #1630

```release-note
Validate cluster name using RFC 1123 regexp
```

### DIFF
--- a/examples/terraform/aws/variables.tf
+++ b/examples/terraform/aws/variables.tf
@@ -19,7 +19,7 @@ variable "cluster_name" {
   type        = string
 
   validation {
-    condition     = can(regex("^[0-9a-z-]+$", var.cluster_name))
+    condition     = can(regex("^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$", var.cluster_name))
     error_message = "Value of cluster_name should be lowercase and can only contain alphanumeric characters and hyphens(-)."
   }
 }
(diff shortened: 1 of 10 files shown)


ahmedwaleedmalik (approved):
/lgtm
/approve

PR #1651, opened 2021-11-29
Use kubeone.k8c.io/v1beta2 for the example config

xmudrii:

I forgot to update the example config to use the new API, so this PR is a follow-up to fix this.

This PR also removes 2 minutes timeout from the E2E tests. That timeout is legacy from times before we retried reconnecting if the instances are not ready, so I don't think we need it anymore.

```release-note
NONE
```

/assign @kron4eg


kron4eg (approved):
yeah, nowadays we retry on our own

PR #1666, opened 2021-12-06
Add support for OpenStack application credentials and update validation

### LINKED ISSUE #1659: Support OpenStack Application Credentials for machine-controller

Currently, we only support username/password-based credentials for OpenStack. However, CCM and machine-controller also support Application Credentials, so we should add support for those credentials in KubeOne as well.

embik:
machine-controller supports OpenStack application credentials since https://github.com/kubermatic/machine-controller/pull/900. This PR adds support for picking that up and passing it as part of the credentials secret.

It also adjusts the validation function for OpenStack to make sure we're not passing both application and user credentials at the same time.

Fixes #1659 

```release-note
Add support for OpenStack application credentials
```

### DIFF
--- a/pkg/credentials/credentials.go
+++ b/pkg/credentials/credentials.go
@@ -31,27 +31,29 @@ import (
 // The environment variable names with credential in them
 const (
 	// Variables that KubeOne (and Terraform) expect to see
-	AWSAccessKeyID          = "AWS_ACCESS_KEY_ID"
-	AWSSecretAccessKey      = "AWS_SECRET_ACCESS_KEY" //nolint:gosec
-	AzureClientID           = "ARM_CLIENT_ID"
-	AzureClientSecret       = "ARM_CLIENT_SECRET" //nolint:gosec
-	AzureTenantID           = "ARM_TENANT_ID"
-	AzureSubscribtionID     = "ARM_SUBSCRIPTION_ID"
-	DigitalOceanTokenKey    = "DIGITALOCEAN_TOKEN"
-	GoogleServiceAccountKey = "GOOGLE_CREDENTIALS"
-	HetznerTokenKey         = "HCLOUD_TOKEN"
-	OpenStackAuthURL        = "OS_AUTH_URL"
-	OpenStackDomainName     = "OS_DOMAIN_NAME"
-	OpenStackPassword       = "OS_PASSWORD"
-	OpenStackRegionName     = "OS_REGION_NAME"
-	OpenStackTenantID       = "OS_TENANT_ID"
-	OpenStackTenantName     = "OS_TENANT_NAME"
-	OpenStackUserName       = "OS_USERNAME"
-	PacketAPIKey            = "PACKET_AUTH_TOKEN" //nolint:gosec
-	PacketProjectID         = "PACKET_PROJECT_ID"
-	VSphereAddress          = "VSPHERE_SERVER"
-	VSpherePassword         = "VSPHERE_PASSWORD"
-	VSphereUsername         = "VSPHERE_USER"
+	AWSAccessKeyID                       = "AWS_ACCESS_KEY_ID"
+	AWSSecretAccessKey                   = "AWS_SECRET_ACCESS_KEY" //nolint:gosec
+	AzureClientID                        = "ARM_CLIENT_ID"
+	AzureClientSecret                    = "ARM_CLIENT_SECRET" //nolint:gosec
+	AzureTenantID                        = "ARM_TENANT_ID"
+	AzureSubscribtionID                  = "ARM_SUBSCRIPTION_ID"
+	DigitalOceanTokenKey                 = "DIGITALOCEAN_TOKEN"
+	GoogleServiceAccountKey              = "GOOGLE_CREDENTIALS"
+	HetznerTokenKey                      = "HCLOUD_TOKEN"
+	OpenStackAuthURL                     = "OS_AUTH_URL"
+	OpenStackDomainName                  = "OS_DOMAIN_NAME"
+	OpenStackPassword                    = "OS_PASSWORD"
+	OpenStackRegionName                  = "OS_REGION_NAME"
+	OpenStackTenantID                    = "OS_TENANT_ID"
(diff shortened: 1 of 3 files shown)


pkg/credentials/credentials.go
```diff
@@ -158,6 +160,8 @@ func ProviderCredentials(cloudProvider kubeone.CloudProviderSpec, credentialsFil
 			{Name: OpenStackAuthURL},
 			{Name: OpenStackUserName, MachineControllerName: OpenStackUserNameMC},
 			{Name: OpenStackPassword},
+			{Name: OpenStackApplicationCredentialID, MachineControllerName: OpenStackUserNameMC},
```
xmudrii, line 163: `MachineControllerName` is not needed here:

```suggestion
			{Name: OpenStackApplicationCredentialID},
```

pkg/credentials/credentials.go
```diff
@@ -283,13 +287,45 @@ func defaultValidationFunc(creds map[string]string) error {
...
+			OpenStackUserName, OpenStackPassword,
+		)
+	}
+
+	if appCredsInUse && userCredsInUse {
+		return errors.Errorf("both app credentials (%s, %s) and user credentials (%s, %s) defined",
+			OpenStackApplicationCredentialID, OpenStackApplicationCredentialSecret,
+			OpenStackUserName, OpenStackPassword,
+		)
 	}
```
xmudrii, line 329: Should we care about this at all? The GopherCloud library used by Terraform and machine-controller knows how to handle this case -- [by preferring the Application Credentials](https://github.com/gophercloud/gophercloud/blob/master/openstack/auth_env.go).

Even if both are provided, we would just put them in the Secret, and the library would figure out what to use. However, a good thing in this approach is that we are explicit -- we are not going to put something in the Secret that we are not going to use, therefore, we're improving the security.

cc: @kron4eg (for opinion)

pkg/credentials/credentials.go
```diff
@@ -283,13 +287,45 @@ func defaultValidationFunc(creds map[string]string) error {
 }
 
 func openstackValidationFunc(creds map[string]string) error {
-	for k, v := range creds {
-		if k == OpenStackTenantID || k == OpenStackTenantName {
-			continue
+	alwaysRequired := []string{OpenStackAuthURL, OpenStackDomainName, OpenStackRegionName}
```
xmudrii, line 290: @kron4eg Do you have some idea if the Domain Name is really required? I'm trying to figure it out from the [GopherCloud library](https://github.com/gophercloud/gophercloud/blob/master/openstack/auth_env.go), and it seems that they only require it if `applicationCredentialID` is empty and `applicationCredentialSecret` is not empty, which can't happen in our case.

Instead, they have a validation that Tenant ID is required if the Tenant Name is provided, and domain ID and name are not provided.

pkg/credentials/credentials.go
kron4eg: Subjectively, I like this. It might be an error from the user that they have both in their ENV. We might accidentally leak non-application credentials, so having such error is good.

pkg/credentials/credentials.go
embik: The list of "always required" is basically just subtracting the variables that are checked separately from the complete list of variables, so this was required before, but I'm open to removing it.

kron4eg: I guess at this moment we don't have a choice

https://github.com/kubermatic/machine-controller/blob/07e3ca3aa199ca8ada73fada539ef565b45612a1/pkg/cloudprovider/provider/openstack/provider.go#L444-L446

pkg/credentials/credentials.go
xmudrii: Yeah, that's what I thought as well -- only one type of credentials can be used at the time anyways. Even if they have both, they can always temporarily `unset` credentials they don't want to use/deploy.

pkg/credentials/credentials.go
xmudrii: Good point, I forgot to check in machine-controller.

pkg/credentials/credentials.go
```diff
@@ -283,13 +287,45 @@ func defaultValidationFunc(creds map[string]string) error {
...
+		appCredsInUse  bool
+		userCredsInUse bool
+	)
+
+	if v, ok := creds[OpenStackApplicationCredentialID]; ok && len(v) != 0 {
+		if v, ok := creds[OpenStackApplicationCredentialSecret]; !ok || len(v) == 0 {
+			return errors.Errorf("key %v is required if %v is present", OpenStackApplicationCredentialSecret, OpenStackApplicationCredentialID)
 		}
+		appCredsInUse = true
+	}
```
xmudrii, line 308: What if we have the other way -- Credential Secret provided, but not the Credential ID? We'll pass the validation and deploy the Credential Secret, but it wouldn't be useful without the ID (unless I'm mistaken). I think we should catch this case as well.

pkg/credentials/credentials.go
```diff
@@ -283,13 +287,45 @@ func defaultValidationFunc(creds map[string]string) error {
...
+
+	if v, ok := creds[OpenStackUserName]; !appCredsInUse && ok && len(v) != 0 {
+		if v, ok := creds[OpenStackPassword]; !ok || len(v) == 0 {
+			return errors.Errorf("key %v is required but is not present", OpenStackPassword)
+		}
+		userCredsInUse = true
+	}
```
xmudrii, line 315: Similar to the other comment, what if the password is provided but not the username?

Also, I'm not sure about the `!appCredsInUse` check -- if the app credentials are used, `userCredsInUse` will always be false (because of `!appCredsInUse`), so the validation on L324 (`appCredsInUse && userCredsInUse`) doesn't have any effect. If this is correct, we should probably drop `!appCredsInUse`.

embik: Should be addressed now.

pkg/credentials/credentials.go
embik: Should be addressed now.

xmudrii (approved):
/lgtm
/approve

PR #1668, opened 2021-12-06
Addon directory path should be optional when using embedded addons

### LINKED ISSUE #1496: Addons directory path is required even if using only embedded addons
The users are still required to provide the addons directory path and have the directory created, even if using only addons embedded in the KubeOne binary.

For example, having a KubeOneCluster manifest such as this one will produce an error:

```yaml
kind: KubeOneCluster
versions:
  kubernetes: 1.21.4
cloudProvider:
  aws: {}
addons:
  enable: true
  addons:
    - name: default-storage-class
```

ahmedwaleedmalik:
Specifying `.addons.path` is considered a hard requirement even if a user is using embedded addons, that do not need any addon directory. This PR removes this unnecessary requirement.

Fixes #1496

```release-note
* Addon path is not required when using only embedded addons
* Defaulting of `.addons.path` to "./addons" removed
```

### DIFF
--- a/pkg/addons/helpers.go
+++ b/pkg/addons/helpers.go
@@ -47,8 +47,8 @@ func vSphereService() *corev1.Service {
 	}
 }
 
-// AreEmbeddedAddons checks if all specified addons are embedded addons
-func AreEmbeddedAddons(addons []kubeoneapi.Addon) (bool, error) {
+// EmbeddedAddonsOnly checks if all specified addons are embedded addons
+func EmbeddedAddonsOnly(addons []kubeoneapi.Addon) (bool, error) {
 	// Read the directory entries for embedded addons
 	embeddedAddons, err := fs.ReadDir(embeddedaddons.FS, ".")
 	if err != nil {
@@ -62,7 +62,7 @@ func AreEmbeddedAddons(addons []kubeoneapi.Addon) (bool, error) {
 		for _, embeddedAddon := range embeddedAddons {
 			if embeddedAddon.Name() == addon.Name {
 				embedded = true
-				continue
+				break
 			}
 		}
 		// At each iteration of the outer loop check if the "addon" was a customer/non-embedded addon
--- a/pkg/addons/helpers.go
+++ b/pkg/addons/helpers.go
@@ -17,6 +17,10 @@ limitations under the License.
 package addons
 
 import (
+	"io/fs"
+
+	embeddedaddons "k8c.io/kubeone/addons"
+	kubeoneapi "k8c.io/kubeone/pkg/apis/kubeone"
 	"k8c.io/kubeone/pkg/clientutil"
 	"k8c.io/kubeone/pkg/state"
 
@@ -41,4 +45,30 @@ func vSphereService() *corev1.Service {
 			Namespace: metav1.NamespaceSystem,
 		},
 	}
+}
+
+// AreEmbeddedAddons checks if all specified addons are embedded addons
+func AreEmbeddedAddons(addons []kubeoneapi.Addon) (bool, error) {
+	// Read the directory entries for embedded addons
+	embeddedAddons, err := fs.ReadDir(embeddedaddons.FS, ".")
+	if err != nil {
+		return false, err
+	}
+
+	// Iterate over addons specified in the KubeOneCluster object
+	for _, addon := range addons {
+		embedded := false
+		// Iterate over embedded addons directory to check if the addon exists
+		for _, embeddedAddon := range embeddedAddons {
+			if embeddedAddon.Name() == addon.Name {
+				embedded = true
+				continue
+			}
+		}
+		// At each iteration of the outer loop check if the "addon" was a customer/non-embedded addon
+		if !embedded {
+			return false, nil
+		}
+	}
+	return true, nil
 }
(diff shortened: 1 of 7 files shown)


pkg/addons/ensure.go
```diff
@@ -35,25 +35,26 @@ var (
...
+		resources.AddonCSIAwsEBS:           "",
+		resources.AddonCSIAzureDisk:        "",
+		resources.AddonCSIAzureFile:        "",
+		resources.AddonCSIHetzner:          "",
+		resources.AddonCSIOpenStackCinder:  "",
+		resources.AddonCSIVsphere:          "",
+		resources.AddonMachineController:   "",
+		resources.AddonMetricsServer:       "",
+		resources.AddonNodeLocalDNS:        "",
+		resources.AddonDefaultStorageClass: "",
```
xmudrii, line 57: 1. This map is used to list addons that are deployed by KubeOne automatically (e.g. CNIs, CCMs/CSIs, etc.). Addons listed here shouldn't appear in the configuration manifest unless the user wants to change some settings (however, no addon currently supports this).
2. `default-storage-class` is not the only embedded addon, we have a few others as well.

pkg/addons/helpers.go
```diff
@@ -42,3 +46,29 @@ func vSphereService() *corev1.Service {
...
+	}
+
+	// Iterate over addons specified in the KubeOneCluster object
+	for _, addon := range addons {
+		embedded := false
+		// Iterate over embedded addons directory to check if the addon exists
+		for _, embeddedAddon := range embeddedAddons {
+			if embeddedAddon.Name() == addon.Name {
+				embedded = true
+				continue
```
xmudrii, line 65: Shouldn't this be `break` instead?

```suggestion
				break
```

pkg/cmd/config.go
```diff
@@ -801,8 +801,8 @@ addons:
   enable: false
   # In case when the relative path is provided, the path is relative
   # to the KubeOne configuration file.
-  # This path must be always provided and the directory must exist, even if
-  # using only embedded addons.
+  # This path must be provided and the directory must exist, when using
+  # custom addons.
```
xmudrii, line 805: Nit:

```suggestion
  # This path is required only if you want to provide custom addons or override
  # embedded addons.
```

pkg/addons/helpers.go
```diff
@@ -42,3 +46,29 @@ func vSphereService() *corev1.Service {
 		},
 	}
 }
+
+// AreEmbeddedAddons checks if all specified addons are embedded addons
+func AreEmbeddedAddons(addons []kubeoneapi.Addon) (bool, error) {
```
xmudrii, line 51: Would it be better if we call this function `EmbeddedAddonsOnly` instead?

pkg/apis/kubeone/validation/validation.go
```diff
@@ -461,7 +462,13 @@ func ValidateAddons(o *kubeone.Addons, fldPath *field.Path) field.ErrorList {
...
 	}
 	if o.Enable && len(o.Path) == 0 {
-		allErrs = append(allErrs, field.Invalid(fldPath.Child("path"), "", ".addons.path must be specified"))
+		// Check if only embedded addons are being used; path is not required for embedded addons
+		areEmbeddedAddons, err := addons.AreEmbeddedAddons(o.Addons)
+		if err != nil {
+			allErrs = append(allErrs, field.Invalid(fldPath, "", "failed to read embedded addons directory"))
+		} else if !areEmbeddedAddons {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("path"), "", ".addons.path must be specified when using non-embedded addon(s)"))
+		}
```
xmudrii, line 471: What if someone provides something like this:

```yaml
addons:
  enable: true
```

Should we reject that as invalid or allow it? As far as I see, the current implementation will allow it, but does such configuration make sense at all?

pkg/addons/helpers.go
ahmedwaleedmalik: Sure thing let's call it that

pkg/apis/kubeone/validation/validation.go
ahmedwaleedmalik: TBH if the user specifies that is he doing something wrong? I don't think so but yes the configuration doesn't make sense either. It's something somewhat benign though.
WDYT @kron4eg? Let's weigh-in and make a quick decision.

kron4eg (approved):
/approve

pkg/apis/kubeone/validation/validation.go
xmudrii: I discussed this with @kron4eg and we concluded that the following config should be treated as invalid:

```yaml
...
addons:
  enable: true
```

In other words, if the user:

* doesn't enable any embedded addon
* doesn't provide a path (in which case no custom addon can be deployed)

we should reject such config, i.e. validation should fail. That's because such addons config is not useful or doing anything at all.

@ahmedwaleedmalik Please add a validation case for this.

ahmedwaleedmalik: Okay :+1:

ahmedwaleedmalik: Resolved, PTAL.

xmudrii (approved):
/lgtm
/approve

PR #1642, opened 2021-11-23
Addons list

kron4eg:
**What this PR does / why we need it**:
subcommand to list addons

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1622 
Fixes #1653

**Does this PR introduce a user-facing change?**:
```release-note
addons list subcommand
```

Issue #1612, opened 2021-11-04
registry override removes path from coredns image


qeqar:
Wen using the registryOverride option, the path part in the coredns image get lost.

k8s.gcr.io/coredns/coredns:v1.8.4 -> override.domain/coredns:v1.8.4
So the path coredns was striped and the image therefore not found.

**Information about the environment**:
KubeOne version 1.3.: 
Provider you're deploying cluster on: openstack
Operating system you're deploying on: flatcar

qeqar:
Missing Kubeone config:

-----
kind: KubeOneCluster
name: %CLUSTER_NAME%
versions:
  kubernetes: '%CLUSTER_VERSION%'
clusterNetwork:
  cni:
    canal:
      mtu: 1450
cloudProvider:
  openstack: {}
  external: true
  cloudConfig: |
    [Global]
    region=%OS_REGION_NAME%
    username="%OS_USERNAME%"
    password="%OS_PASSWORD%"
    auth-url=%OS_AUTH_URL%
    tenant-id=%OS_PROJECT_ID%
    domain-id=%OS_PROJECT_DOMAIN_ID%

    [LoadBalancer]
    subnet-id=%SUBNET_ID%
    floating-network-id=ID
    create-monitor=true
    monitor-delay=10s
    monitor-max-retries=3
    monitor-timeout=3s
    use-octavia=true

    [Metadata]
    search-order=metadataService,configDrive
containerRuntime:
  containerd: {}
features:
  openidConnect:
    enable: true
    config:
      issuerUrl: xxx
      clientId: xxx
      usernamePrefix: xxx
  metricsServer:
    enable: true
registryConfiguration:
  overwriteRegistry: registry.domain
-----

qeqar:
using the assetConfiguration helps as a workaround.

```
assetConfiguration:                                                             
       coreDNS:                                                                      
         imageRepository: registry.domain/coredns     
```

But this is deprecated stuff and will removed in kubeone 1.4

embik:
I looked into this and it seems to be something that it is intentional in `kubeadm`. See https://github.com/kubernetes/kubeadm/issues/2513 and https://github.com/kubernetes/kubernetes/pull/102502.

Since a repository that mirrors/caches upstream is likely placing the image in `coredns/coredns`, we probably need to adjust the passed kubeadm config to include it. Let me check on this.

PR #1671, opened 2021-12-08
Fix the overwrite registry functionality

xmudrii:

This is the first phase of fixing the overwrite registry functionality and the image loader script, which includes:

* The image loader script now supports KubeOne 1.3+ and Kubernetes 1.22+
  * The image list from KubeOne is now properly parsed
  * Images are now filtered by the Kubernetes version as well (using the newly-added `--kubernetes-version` flag)
  * The CoreDNS image regression is properly handled by checking if the Kubernetes version is 1.21 or not
      * For 1.21 clusters => tag CoreDNS as `<registry>/coredns/coredns`
      * For other versions (including 1.22+) => tag CoreDNS as `<registry>/coredns`
  * The Kubelet image is not anymore preloaded because machine-controller doesn't use it any longer for Flatcar clusters
* Add the `--kubernetes-version` flag to the `kubeone config images` command
  * Many optional images depend on the Kubernetes version. When preloading the images, it's critical to have images that you'll be using for the concrete Kubernetes version
* The `kubeone config images` command now returns images for the latest Kubernetes version instead of for the oldest
  * There are more chances that users will run the latest or latest-1 Kubernetes than 1.19 or 1.20

The problem with the current approach is that the `kubeone config images` command includes all images, even those that are not applicable for the desired cloud provider. For example, the command will include images for Azure CCM and CSI even if the provider is AWS.

This problem will be fixed in the second phase. The reason for doing this in two phases is to make it easier to cherry-pick this fix to the `release/v1.3` branch. My current plan is to use the addons list implemented in #1642 to get the images that will be deployed on the concrete cluster, but cherry-picking that would be too complicated. A follow-up issue will be created for this.

The accomplishing docs PR is available here: https://github.com/kubermatic/docs/pull/868

Related to #1612

```release-note
* Fix the image loader script to support KubeOne 1.3+ and Kubernetes 1.22+
* The `kubeone config images` command now shows images for the latest Kubernetes version (instead of for the oldest)
* Add a new `--kubernetes-version` flag to the `kubeone config images` command. This flag is used to filter images for a particular Kubernetes version. The flag cannot be used along with the KubeOneCluster manifest (`--manifest` flag)
```

/assign @kron4eg @embik

### DIFF
--- a/hack/image-loader.sh
+++ b/hack/image-loader.sh
@@ -55,6 +55,7 @@ function fail() {
 
 function retag() {
   local image="$1"
+  local custom_image_name="${2:-}"
 
   # Trim registry
   local local_image
@@ -65,6 +66,9 @@ function retag() {
   # Split into name and tag
   name="$(echo "${local_image}" | cut -d: -f1)"
   tag="$(echo "${local_image}" | cut -d: -f2)"
+  if [ -n "$custom_image_name" ]; then
+    name="$custom_image_name"
+  fi
 
   # Build target image name
   local target_image="${TARGET_REGISTRY}/${name}:${tag}"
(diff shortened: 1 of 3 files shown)


embik (approved):
/lgtm

kron4eg (approved):

PR #1691, opened 2021-12-16
Resolve environment variables in addon parameters

### LINKED ISSUE #1498: Make it possible to reference to environment variables for addons params
It would be very useful if addons params could refer to the environment variables. That would allow users to put secret values in the environment instead of the KubeOneCluster manifest, which might be committed to the version control systems.

We already have that for `hosts.ssh_agent_socket` where users can use `env:SSH_AUTH_SOCK` to take the value of `SSH_AUTH_SOCK` environment variable.

embik:
Parameters for addons can potentially include sensitive information / secrets. Since the kubeone config file could be checked into version control, this PR adds the capability to prefix a parameter string with `env:` to make kubeone load the environment variable. It supports both addon-specific and global parameters.

The config file could then look like this:

```yaml
addons:
  enable: true
  path: "./addons"
  globalParams:
    globalTest: "env:SUPER_SECRET_VALUE2"
  addons:
    - name: "default-storage-class"
    - name: "testaddon"
      params:
        test: "env:SUPER_SECRET_VALUE"
```

`SUPER_SECRET_VALUE2` and `SUPER_SECRET_VALUE` need to be set as environment variables then.

Fixes #1498

```release-note
Addon parameters can be resolved into environment variable contents if the "env:" prefix is set in the parameter value
```

### DIFF
--- a/pkg/addons/manifest.go
+++ b/pkg/addons/manifest.go
@@ -20,6 +20,7 @@ import (
 	"bufio"
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"io"
 	"io/fs"
 	"os"
@@ -43,7 +44,7 @@ import (
 )
 
 const (
-	paramsEnvPrefix = "env:"
+	ParamsEnvPrefix = "env:"
 )
 
 func (a *applier) getManifestsFromDirectory(s *state.State, fsys fs.FS, addonName string) (string, error) {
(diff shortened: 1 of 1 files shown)


pkg/addons/manifest.go
```diff
@@ -128,6 +133,17 @@ func (a *applier) loadAddonsManifests(
 			tplDataParams[k] = v
 		}
 
+		// Resolve environment variables in Params
+		for k, v := range tplDataParams {
+			if strings.HasPrefix(v, paramsEnvPrefix) {
+				envName := strings.TrimPrefix(v, paramsEnvPrefix)
+				if env := os.Getenv(envName); len(env) > 0 {
```
xmudrii, line 140: I think an empty env is okay and we should respect it. Therefore, I'd replace this with [`os.LookupEnv`](https://pkg.go.dev/os#LookupEnv):

```suggestion
				if env, ok := os.Getenv(envName); ok {
```

Also, we should consider returning an error if the provided environment variable doesn't exist. I think that would provide a better UX instead of silently failing.

embik: Fair point, I'll change that!

xmudrii (approved):
/lgtm
/approve

xmudrii:
/hold
The test is broken, but it's not related to this PR.

PR #1692, opened 2021-12-17
Fix flatcar AMI filter

kron4eg:
Seems like recently flatcar got new arm64 images, and THEY starter to return from search query, which leads to incompatible architecture errors between AMI and instance.

This PR fixes that problem

```release-note
Fix flatcar AMI filter
```


xmudrii (approved):

kron4eg:
cherry pick: #1693

PR #1698, opened 2021-12-20
New KubeletConfig API

### LINKED ISSUE #1216: Add ability to set Eviction and resource reservation kubelet policies
Hi,
Is it possible to set specific options mentioned here: https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/

I've found a kubelet-18.0 configmap in the kube-system namespace, but I have been unsure if I can just adjust it. It would be nice to set such arguments via kubeone.

kron4eg:
Allows configuring some, strict set of kubelet options

Fixes #1216

```release-note
New KubeletConfig API
```


pkg/apis/kubeone/types.go
```diff
@@ -63,6 +63,8 @@ type KubeOneCluster struct {
 	AssetConfiguration AssetConfiguration `json:"assetConfiguration,omitempty"`
 	// RegistryConfiguration configures how Docker images are pulled from an image registry
 	RegistryConfiguration *RegistryConfiguration `json:"registryConfiguration,omitempty"`
+	// Kubelet
+	Kubelet KubeletConfig `json:"kubelet,omitempty"`
```
xmudrii, line 67: Does it make sense to have kubelet settings per node instead? Or at least per node type (control plane and worker)? Control plane and worker nodes run different components, and therefore might have different settings. For example, you might reserve fewer resources on worker nodes because you don't run all the control plane components there. You might also use smaller instances for worker nodes as an example.

kron4eg: The problem here is kubeadm and the way it works with kubelet params.

The `KubeletConfiguration` that we provide is a shared resource (uploaded as configMap) and will be used across all the kubeadm-managed nodes. 

There is one way to have node specific kubelet customizations, and it's done by providing `NodeRegistrationOptions.KubeletExtra` argument to the Kubeadm API. KubeletExtra is a map[string]string that will be transformed into the command line flags, and flags that are responsible for resource reservation/eviction are deprecated and is being asked to be configured in the `KubeletConfiguration`.

kron4eg: All this means that we technically can provide a possibility to specify different resource-reservations/evicion for different nodes (literally, per each node) but we will have to use deprecated flags.

In fact, I've started from doing exactly this, having new field KubeletConfig in HostConfig structure, but later realized that it will lead to the deprecated flags usage.

xmudrii (approved):
/lgtm
/approve

PR #1701, opened 2021-12-23
Let kubeadm manage cri-tools on Amazon Linux 2

xmudrii:

Before this PR, we were manually installing cri-tools 1.13.0 on Amazon Linux 2. Newer Kubernetes (kubeadm) versions are requiring cri-tools 1.19.0 which is conflicting with cri-tools 1.13.0. This is causing the cluster reconciliation/provisioning to fail.

With this PR, cri-tools is now installed automatically as a dependency of kubeadm. This way the package manager (`yum`) will choose the proper version.

If users are using the AssetConfiguration API (i.e. when they're not using packages to install kubeadm), we install the latest available cri-tools package using `yum`.

```release-note
cri-tools is now installed automatically as a dependency of kubeadm on Amazon Linux 2. This fixes provisioning issues on Amazon Linux 2 with newer Kubernetes versions.
```

/assign @moadqassem

### DIFF
--- a/docs/api_reference/v1beta1.en.md
+++ b/docs/api_reference/v1beta1.en.md
@@ -1,6 +1,6 @@
 +++
 title = "v1beta1 API Reference"
-date = 2021-12-21T13:30:12+01:00
+date = 2021-12-23T12:22:34+01:00
 weight = 11
 +++
 ## v1beta1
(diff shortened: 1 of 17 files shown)


moadqassem (approved):
/approve
/lgtm

Issue #1669, opened 2021-12-07
Generate Software Bill of Materials (SBOM) for every KubeOne release


xmudrii:
We should generate SBOM files for each KubeOne release, so the release can be properly verified. To accomplish this, we can use the [BOM tool](https://github.com/kubernetes-sigs/bom) created by the Kubernetes Release Engineering team.

This concrete tool has been already tested with KubeOne in the past, and it seems to work very well.

**Acceptance criteria:**

* Generate SBOM file for each KubeOne release when cutting the release

PR #1708, opened 2021-12-27
Make goreleaser generate SBOM using kubernetes-sigs/bom

xmudrii:

This PR updates the GoReleaser script to generate SBOM on release time using [`kubernetes-sigs/bom` tool](https://github.com/kubernetes-sigs/bom). This PR:

* is an RFC to get the discussion started on how should we structure our SBOM files
* adds a way to generate initial SBOMs, so we can test and exercise the tooling

For now, there are some key questions:

* Which tool do we want to use for generating SBOMs?
    * There are many tools and formats out there, but I'm in favor of [`kubernetes-sigs/bom`](https://github.com/kubernetes-sigs/bom) because it's a Go tool created by Kubernetes, so it's probably the closest to us
    * Possible alternative is to use [Syft](https://github.com/anchore/syft) which also integrates with GoReleaser
* What do we want to include in our SBOM files?
  * Right now, this would only include source files and information about dependencies (including their LICENSEs). Ideally, we should include a few more things.
  * We should include information about the generated release archive, but we can't use GoReleaser for that. We would have to do generate SBOMs on our own in this case.
  * Should we include images that we use and deploy in clusters? Are those images our dependencies? Should we have them analyzed? Is it just enough to have YAML manifests listed in the SBOM?

Relevants to #1669

```release-note
KubeOne releases now include Software Bill of Materials (SBOM)
```

/assign @scheeles 
/hold
for discussion

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,4 +1,7 @@
 project_name: kubeone
+before:
+  hooks:
+    - "bom generate -n https://kubeone.io/kubeone-v{{.Version}}.spdx -o ./kubeone-v{{.Version}}.spdx --dirs=."
 builds:
 - ldflags:
   - "-s -w"
@@ -26,3 +29,6 @@ archives:
       - examples/**/*
       - hack/image-loader.sh
       - addons/**/*
+release:
+  extra_files:
+    - glob: kubeone-v{{.Version}}.spdx


xmudrii:
We'll revisit this topic soon, until then I'll close this PR.
/close

PR #1713, opened 2022-01-03
update cluster-autoscaler addon to match upstream example

c4tz:
Most importantly: Adds permissions to watch `csistoragecapacities`, `csidrivers` and `namespaces` objects, reducing errors such as:

```
Failed to watch *v1.CSIDriver: failed to list *v1.CSIDriver: csidrivers.storage.k8s.io is forbidden: User "system:serviceaccount:kube-system:cluster-autoscaler" cannot list resource "csidrivers" in API group "storage.k8s.io" at the cluster scope
```

Also, it adds labels to the ClusterRole, ClusterRoleBinding, etc. for better filtering as done in [the offiicial cluster-autoscaler example for Hetzner](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/hetzner/examples/cluster-autoscaler-run-on-master.yaml).

None, created this PR directly instead of opening an issue first.

I'm not quite sure whether it makes sense to also add the `affinity` bit to the deployment aswell:

```
      # Node affinity is used to force cluster-autoscaler to stick
      # to the master node. This allows the cluster to reliably downscale
      # to zero worker nodes when needed.
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
              - matchExpressions:
                  - key: node-role.kubernetes.io/master
                    operator: Exists
```

No.

```release-note
NONE
```

### DIFF
--- a/addons/cluster-autoscaler/cluster-autoscaler.yaml
+++ b/addons/cluster-autoscaler/cluster-autoscaler.yaml
@@ -2,6 +2,9 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
+  labels:
+    k8s-addon: cluster-autoscaler.addons.k8s.io
+    k8s-app: cluster-autoscaler
   name: cluster-autoscaler
   namespace: kube-system
 ---
@@ -9,6 +12,9 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: cluster-autoscaler
+  labels:
+    k8s-addon: cluster-autoscaler.addons.k8s.io
+    k8s-app: cluster-autoscaler
 rules:
   - apiGroups: [""]
     resources: ["events", "endpoints"]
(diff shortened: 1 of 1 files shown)


kron4eg:
@c4tz thank you!

/lgtm
/approve

c4tz:
Thank you for the fast responses and merging @kron4eg ! :)

What about the `affinity` bit? Should I make a new PR for it (maybe with a release note)? As far as I currently understand [the documentation](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/), the **toleration** *allows* the autoscaler to also run on master nodes, while the **affinity** *binds* the pod to them, so that it doesn't schedule on worker nodes at all.

Is this correct? If yes, wouldn't that be behaviour we want `cluster-autoscaler` to have?

kron4eg:
@c4tz given that CA has tolaration, it will be rescheduled to the control-plane node automatically in the event if it will kill the Machine where it is running and that Machine is the last worker. Should be OK. But having affinity in place wouldn't harm anyway.

PR #1717, opened 2022-01-06
Add support for different credentials for machine-controller and CCM

### LINKED ISSUE #1262: Allow using different credentials for machine-controller and external CCM

Currently, we deploy credentials that are used both by machine-controller and external CCMs (for supported providers). Those two components have a different set of required permissions, so it would be useful to allow users to specify different credentials for those two components.

xmudrii:

This PR adds support for using different credentials for machine-controller and external CCMs. This is implemented in the following way:

* We now create `machine-controller-credentials` and `ccm-credentials` secrets instead of `cloud-provider-credentials` secret
* Typed credentials are prefixed with `MC_` for machine-controller credentials and `CCM_` from CCM credentials
* KubeOne will first try to fetch prefixed credentials and then fallback to non-prefixed (default) credentials

Fixes #1262

```release-note
Add support for different credentials for machine-controller and CCM. Environment variables can be prefixed with `MC_` for machine-controller credentials and `CCM_` for CCM credentials.
[BREAKING] `cloud-provider-credentials` secret is not created and maintained by KubeOne any longer
```

/assign @kron4eg

### DIFF
--- a/pkg/credentials/credentials.go
+++ b/pkg/credentials/credentials.go
@@ -28,6 +28,15 @@ import (
 	"k8c.io/kubeone/pkg/apis/kubeone"
 )
 
+// CredentialsType is a type of credentials that should be fetched
+type CredentialsType string //nolint:revive
+
+const (
+	CredentialsTypeUniversal CredentialsType = ""
+	CredentialsTypeCCM       CredentialsType = "CCM"
+	CredentialsTypeMC        CredentialsType = "MC"
+)
+
 // The environment variable names with credential in them
 const (
 	// Variables that KubeOne (and Terraform) expect to see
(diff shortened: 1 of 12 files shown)


pkg/credentials/credentials.go
```diff
@@ -28,6 +28,15 @@ import (
 	"k8c.io/kubeone/pkg/apis/kubeone"
 )
 
+// CredentialsType is a type of credentials that should be fetched
+type CredentialsType string //nolint:revive
```
kron4eg, line 32: Removing this nolint shows:
```
pkg/credentials/credentials.go:32:6: exported: type name will be used as credentials.CredentialsType by other packages, and that stutters; consider calling this Type (revive)
```

Sounds reasonalbe. The name of the package is also a part of the type identity, so why neglect it?

xmudrii: @kron4eg Because just `Type` sounds very weird to me. Just `Type` doesn't really describe what it refers to. But when I think a bit more about it, it might make sense. Let me rename it.

kron4eg (approved):

PR #1723, opened 2022-01-10
Add Terraform configs for Nutanix

### LINKED ISSUE #1720: Add example Terraform configs for Nutanix
Related to #1697

**Acceptance criteria:**

* Add example Terraform configs for Nutanix

xmudrii:

This PR adds example Terraform configs for Nutanix.

The example configs have the following assumptions:

* Project is supposed to be created and provided by the user because Projects are sort-of-like accounts, and you also need to figure out various settings and permissions, which is complicated to do on our own
* Subnet is supposed to be created and provided by the user as well because creating subnets might require additional network configuration
* We don't manage Network Security Rules because they require an additional license from Nutanix. Users can extend their Terraform configs on their own if they want to use Network Security Rules
* We don't manage Access Control Policies as we also don't manage Projects

The user needs to provide the following information via Terraform variables:

* Kubernetes Cluster name (`cluster_name`) - name to be used as a prefix for Nutanix resources and Kubernetes cluster
* Nutanix Cluster name (`nutanix_cluster_name`) - can be obtained via Prism Central
* Project Name (`project_name`) - can be obtained via Prism Central
* Subnet Name (`subnet_name`) - can be obtained via Prism Central
* Image Name (`image_name`) - the image must be uploaded by the user manually

The following resources are created by Terraform configs:

* KubeOne-Cluster category - the value is the cluster name
* 3 virtual machines for the Control Plane (by default each has 2 vCPUs, 1 CPU socket, 4 GB RAM, 100 GB disk)
* 1 virtual machine as a bastion and for GoBetween Load Balancer (by default machine has 1 vCPUs, 1 CPU socket, 4 GB RAM, 100 GB disk)
  * GoBetween is used to provide Load Balancing for the Kubernetes API servers
  * It uses the same GoBetween config as other providers

Notes:

* There's no way to provide an SSH key to the instance, and the default hostname is `ubuntu`. To fix this, a simple cloud-config is generated to set the hostname and the SSH public key

TODO:

* Terraform integration for Nutanix is not yet implemented, therefore there's no template for MachineDeployments, and cloud provider is set to `none`. This is going to be fixed in a follow up.

Fixes #1720

```release-note
Add example Terraform configs for Nutanix
```

/assign @embik @kron4eg

### DIFF
--- a/examples/terraform/nutanix/README.md
+++ b/examples/terraform/nutanix/README.md
@@ -1,0 +1,77 @@
+# Nutanix Quickstart Terraform configs
+
+The Nutanix Quickstart Terraform configs can be used to create the needed
+infrastructure for a Kubernetes HA cluster. Check out the following
+[Creating Infrastructure guide][docs-infrastructure] to learn more about how to
+use the configs and how to provision a Kubernetes cluster using KubeOne.
+
+## Kubernetes API Server Load Balancing
+
+See the [Terraform loadbalancers in examples document][docs-tf-loadbalancer].
+
+[docs-infrastructure]: https://docs.kubermatic.com/kubeone/master/guides/using_terraform_configs/
+[docs-tf-loadbalancer]: https://docs.kubermatic.com/kubeone/master/examples/ha_load_balancing/
+
+## Requirements
+
+| Name | Version |
+|------|---------|
+| <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 1.0.0 |
+| <a name="requirement_nutanix"></a> [nutanix](#requirement\_nutanix) | 1.2.2 |
+
+## Providers
+
+| Name | Version |
+|------|---------|
+| <a name="provider_null"></a> [null](#provider\_null) | 3.1.0 |
+| <a name="provider_nutanix"></a> [nutanix](#provider\_nutanix) | 1.2.2 |
+
+## Resources
+
+| Name | Type |
+|------|------|
+| [null_resource.lb_config](https://registry.terraform.io/providers/hashicorp/null/latest/docs/resources/resource) | resource |
+| [nutanix_category_key.category_key](https://registry.terraform.io/providers/nutanix/nutanix/1.2.2/docs/resources/category_key) | resource |
+| [nutanix_category_value.category_value](https://registry.terraform.io/providers/nutanix/nutanix/1.2.2/docs/resources/category_value) | resource |
+| [nutanix_virtual_machine.control_plane](https://registry.terraform.io/providers/nutanix/nutanix/1.2.2/docs/resources/virtual_machine) | resource |
+| [nutanix_virtual_machine.lb](https://registry.terraform.io/providers/nutanix/nutanix/1.2.2/docs/resources/virtual_machine) | resource |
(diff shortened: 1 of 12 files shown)


examples/terraform/nutanix/main.tf
```diff
@@ -0,0 +1,155 @@
...
+data "nutanix_subnet" "subnet" {
+  subnet_name = var.subnet_name
+}
+
+data "nutanix_image" "image" {
+  image_name = var.image_name
+}
+
+resource "nutanix_category_key" "category_key" {
+  name        = "KubeOne-Cluster"
```
embik, line 38: Should be `KubeOneCluster` as far as I can say. The built-in categories are all camel case and I was planning to add `KKPCluster` as category key (but I'm very much open to discussing this).

Btw, will this work with multiple cluster?

examples/terraform/nutanix/README.md
```diff
@@ -0,0 +1,77 @@
...
+infrastructure for a Kubernetes HA cluster. Check out the following
+[Creating Infrastructure guide][docs-infrastructure] to learn more about how to
+use the configs and how to provision a Kubernetes cluster using KubeOne.
+
+## Kubernetes API Server Load Balancing
+
+See the [Terraform loadbalancers in examples document][docs-tf-loadbalancer].
+
+[docs-infrastructure]: https://docs.kubermatic.com/kubeone/master/guides/using_terraform_configs/
+[docs-tf-loadbalancer]: https://docs.kubermatic.com/kubeone/master/advanced/example_loadbalancer/
```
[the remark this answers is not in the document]
xmudrii: Yes, that's the document I wanted to link to. Thanks!

examples/terraform/nutanix/variables.tf
```diff
@@ -0,0 +1,150 @@
...
+
+variable "nutanix_cluster_name" {
+  description = "Name of the Nutanix Cluster which will be used for this Kubernetes cluster"
+  type        = string
+}
+
+variable "project_id" {
+  description = "ID of the Nutanix Project"
+  type        = string
+}
```
embik, line 85: Why give a project ID when we ask for names for clusters and subnets?

xmudrii: The Project ID is needed because we use it for VMs (pushed a change to make it work). We can't get Project ID from Clusters or Subnets.

examples/terraform/nutanix/main.tf
```diff
@@ -0,0 +1,163 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+provider "nutanix" {
+  insecure = true
```
embik, line 18: Can we maybe make this configurable and default to `false`? There's hopefully Nutanix setups out there with valid TLS certs.

xmudrii: Added `allow_insecure` field which is `false` by default.

kron4eg (approved):

PR #1724, opened 2022-01-10
Registry credentials configuration

### LINKED ISSUE #1574: add credentials to RegistryConfiguration

Make it possible to add credentials to RegistryConfiguration.

During deployments we're hitting _rate limits_ of Docker Hub causing our Kubeone deployments to fail. We've tried patching containerd config.json with the [credentials of our registry](https://github.com/containerd/containerd/blob/main/docs/cri/registry.md#configure-registry-credentials) (harbor) instances, but those are [overwritten by KubeOne](https://github.com/kubermatic/kubeone/blob/master/pkg/scripts/render.go#L31).
As our registries over the internet internet, we're not allowed by the security department to make the Proxy Cache type _Public_.

This issue is related to #1551, #1470 and #1547 

Use a _Private_ (not _Public_) registry Proxy Cache for caching Docker Hub artifacts. This allows for successful deployments that are not failing due to Docker Hub rate limits.

kron4eg:

Fixes #1574

```release-note
Registry credentials configuration
```


pkg/cmd/config.go
```diff
@@ -610,6 +610,9 @@ containerRuntime:
   #     docker.io:
   #       mirrors:
   #       - http://plain-text2.tld
+  #       auth:
+  #         username: "uzer"
+  #         password: "myc00lp455w0rd"
```
xmudrii, line 615: Can we also document Auth and IdentityToken fields?

xmudrii (changes requested):
Just a nit, but otherwise LGTM.

pkg/cmd/config.go
kron4eg: The docs are in the generated API, this is just an example. But nether the less, done.

xmudrii (approved):
/lgtm
/approve

PR #1745, opened 2022-01-19
Fetch containerd auth config from the credentials file

### LINKED ISSUE #1740: Support sourcing registry credentials from the environment

We introduced support for providing the registry credentials in #1724. The issue is that the credentials can be only provided as plain text in the KubeOneCluster manifest. This wouldn't work well for users that are committing their manifest to Git (e.g. because of the GitOps approach).

We should add a way for credentials to be sourced from somewhere else, e.g. from the environment or a credentials file.

* The registry credentials can be sourced from the environment or the credentials file

cc @kron4eg

kron4eg:
Instead of storing registry authentication info directly in the kubeone config, it's now possible to have this info fetched from the credentials file.

The API reference [ContainerRuntimeContainerd](https://docs.kubermatic.com/kubeone/master/references/kubeone_cluster_v1beta2/#containerruntimecontainerd) and it's used together with apiVersion and kind fields for type metadata.

Limitation: ONLY `ContainerdRegistry.Auth` will be extracted from credentials file.

Example:
```yaml
registriesAuth: |
  kind: ContainerRuntimeContainerd
  registries:
    auth:
      docker.io:
        username: "u5er"
        password: "myc00lp455w0rd"
        auth: "base64(user:password)"
        identityToken: ""  
```

Fixes #1740

```release-note
Fetch containerd auth config from the credentials file
```

### DIFF
--- a/pkg/apis/kubeone/config/config.go
+++ b/pkg/apis/kubeone/config/config.go
@@ -17,13 +17,14 @@ limitations under the License.
 package config
 
 import (
+	"fmt"
 	"io"
 	"os"
 	"os/exec"
+	"reflect"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-	"gopkg.in/yaml.v2"
 
 	kubeoneapi "k8c.io/kubeone/pkg/apis/kubeone"
 	kubeonescheme "k8c.io/kubeone/pkg/apis/kubeone/scheme"
@@ -34,6 +35,7 @@ import (
 	terraformv1beta2 "k8c.io/kubeone/pkg/terraform/v1beta2"
 
 	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/yaml"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/apis/kubeone/config/config.go
```diff
@@ -222,15 +224,61 @@ func SetKubeOneClusterDynamicDefaults(cfg *kubeoneapi.KubeOneCluster, credential
...
+		return fmt.Errorf(".ContainerRuntime.Containerd should be set")
+	}
+
+	if cluster.ContainerRuntime.Containerd.Registries == nil {
+		cluster.ContainerRuntime.Containerd.Registries = map[string]kubeoneapi.ContainerdRegistry{}
+	}
+
+	for registryName, registryInfo := range containerdConfig.Registries {
+		internalRegistry := cluster.ContainerRuntime.Containerd.Registries[registryName]
+		internalRegistry.Auth = (*kubeoneapi.ContainerdRegistryAuthConfig)(registryInfo.Auth)
```
xmudrii, line 279: If I understand this correctly, this is going to map the v1beta2 representation to the internal representation. While this works fine now, is this going to work if we change the internal API for any reason? Would it be better to have some conversion function that's going to map each field from the versioned to the internal API?

xmudrii (commented):
Just one question, otherwise LGTM.

pkg/apis/kubeone/config/config.go
kron4eg: It will blow up in compile time if something will be changed in internal cluster API, and that's deliberately! This compile will let us know for sure that something has been changed and we need to take a look and be aware.

xmudrii (approved):
/lgtm
/approve

PR #1746, opened 2022-01-20
Force static manifests render on upgrade

### LINKED ISSUE #1749: Fix propagation of proxy settings from kubeone.yaml to apiserver env
**What happened**:
if you are having a definition of proxy and no_proxy settings in kubeone, these are propagated on first run to the apiserver manifest on control plane hosts.

Anyway if you update these proxy values, these are not refreshed after running `kubeone apply` or `kubeone apply --force-upgrade`. Only option to trigger that is to change for example the kubernetes version.

Example files:
```
# file: kubeone.yaml
kind: KubeOneCluster
name: test
versions:
  kubernetes: '1.21.9'
proxy:
  http: 'http://xxx:8080'
  https: 'http://xxx:8080'
  noProxy: 'xxx,yyy,zzz'
```

```
# file: kube-apiserver.yaml
env:                                                                                                                                                                                                                                                                    
    - name: HTTP_PROXY                                                                                                                                                                                                                                                      
      value: http://xxx:8080                                                                                                                                                                                                                       
    - name: http_proxy                                                                                                                                                                                                                                                      
      value: http://xxx:8080                                                                                                                                                                                                                       
    - name: HTTPS_PROXY                                                                                                                                                                                                                                                     
      value: http://xxx:8080                                                                                                                                                                                                                       
    - name: https_proxy                                                                                                                                                                                                                                                     
      value: http://xxx:8080                                                                                                                                                                                                                       
    - name: NO_PROXY                                                                                                                                                                                                                                                        
      value: xxx,yyy,zzz                                                                                                                                                                  
    - name: no_proxy                                                                                                                                                                                                                                                        
      value: xxx,yyy,zzz
```

Login is happening here - https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/util/env.go#L27

There is also a related cause from k8s - https://github.com/kubernetes/kubernetes/commit/1eeb3f155fe76d8376dbcf7ee582bda1042e65f0

Proxy setting is updated at least when `--force-upgrade` is used.

Follow the kubeone deployment with proxy settings, then add some new value and run kubeone apply (--force-upgrade) and check that values were not updated on control-plane.

KubeOne version (`kubeone version`): 1.3.3
Operating system: CentOS
Provider you're deploying cluster on: bare-metal (static kubeone setup)

kron4eg:
Fix proxy settings rendering when running kubeone upgrade --force (or apply --force-upgrade).

Fixes #1749 

```release-note
Force static manifests render on upgrade
```

### DIFF
--- a/pkg/tasks/prerequisites.go
+++ b/pkg/tasks/prerequisites.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 
 	kubeoneapi "k8c.io/kubeone/pkg/apis/kubeone"
 	"k8c.io/kubeone/pkg/scripts"
@@ -85,18 +86,27 @@ func generateConfigurationFiles(s *state.State) error {
 func installPrerequisitesOnNode(s *state.State, node *kubeoneapi.HostConfig, _ ssh.Connection) error {
 	logger := s.Logger.WithField("os", node.OperatingSystem)
 
+	err := setupProxy(logger, s)
+	if err != nil {
+		return err
+	}
+
+	logger.Infoln("Installing kubeadm...")
+	return errors.Wrap(installKubeadm(s, *node), "failed to install kubeadm")
+}
+
+func setupProxy(logger *logrus.Entry, s *state.State) error {
 	logger.Infoln("Creating environment file...")
 	if err := createEnvironmentFile(s); err != nil {
 		return errors.Wrap(err, "failed to create environment file")
 	}
 
 	logger.Infoln("Configuring proxy...")
-	if err := configureProxy(s); err != nil {
+	if err := containerRuntimeEnvironment(s); err != nil {
 		return errors.Wrap(err, "failed to configure proxy for docker daemon")
 	}
 
-	logger.Infoln("Installing kubeadm...")
-	return errors.Wrap(installKubeadm(s, *node), "failed to install kubeadm")
+	return nil
 }
 
 func createEnvironmentFile(s *state.State) error {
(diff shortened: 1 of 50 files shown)


pkg/tasks/upgrade_follower.go
```diff
@@ -51,13 +51,17 @@ func upgradeFollowerExecutor(s *state.State, node *kubeoneapi.HostConfig, conn s
 		return errors.Wrap(err, "failed to drain follower control plane node")
 	}
 
+	if err := setupProxy(logger, s); err != nil {
+		return err
+	}
```
xmudrii, line 56: The problem with this approach is that it's going to restart the container runtime. While it's okay-ish for containerd, it might not be okay for Docker because it's going to restart all containers. Do we have some other way around? Can we only update the environment file here?

kron4eg: OK, I've changed a bit whole concept

kron4eg: On the other hand, by the time when docker is restarted the node **IS DRAINED AND SCHEDULING IS DISABLED**, so any remaining pods are safe to restart.

pkg/tasks/prerequisites.go
```diff
@@ -250,13 +260,13 @@ func uploadConfigurationFilesToNode(s *state.State, _ *kubeoneapi.HostConfig, co
...
-func configureProxy(s *state.State) error {
+func containerRuntimeEnvironment(s *state.State) error {
 	if s.Cluster.Proxy.HTTP == "" && s.Cluster.Proxy.HTTPS == "" && s.Cluster.Proxy.NoProxy == "" {
 		return nil
 	}
 
-	s.Logger.Infoln("Configuring docker/kubelet proxy...")
-	cmd, err := scripts.DaemonsProxy()
+	s.Logger.Infoln("Configuring docker/containerd/kubelet environment...")
+	cmd, err := scripts.DaemonsEnvironmentDropIn("docker", "containerd", "kubelet")
```
xmudrii, line 269: Shouldn't we add drop-in for Docker only if it's used?

kron4eg: It's OK to have them all despite if docker present or absent. And turns our both, containerd and docker wants to have proxies defined.

xmudrii (approved):
/lgtm
/approve

PR #1752, opened 2022-01-24
replace hubble static certs with CronJob generation

WeirdMachine:
According to https://github.com/kubermatic/kubermatic/pull/8548 

```release-note
Replace hubble static ca with cron job generation.
Update Cilium to v11.1.
```


addons/cni-cilium/cilium.yaml
```diff
@@ -1208,4 +1235,85 @@ spec:
...
+    k8s-app: hubble-generate-certs
+spec:
+  template:
+    metadata:
+      labels:
+        k8s-app: hubble-generate-certs
+    spec:
+      containers:
+        - name: certgen
+          image: "quay.io/cilium/certgen:v0.1.5"
```
xmudrii, line 1255: This image should be parametrized like other images.

addons/cni-cilium/cilium.yaml
```diff
@@ -1208,4 +1235,85 @@ spec:
...
+  jobTemplate:
+    spec:
+      template:
+        metadata:
+          labels:
+            k8s-app: hubble-generate-certs
+        spec:
+          containers:
+            - name: certgen
+              image: "quay.io/cilium/certgen:v0.1.5"
```
xmudrii, line 1297: This image should be parametrized like other images.

addons/cni-cilium/cilium.yaml
```diff
@@ -1208,4 +1235,85 @@ spec:
...
+---
+# Source: cilium/templates/hubble/tls-cronjob/cronjob.yaml
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: hubble-generate-certs
+  namespace: kube-system
+  labels:
+    k8s-app: hubble-generate-certs
+spec:
```
xmudrii, line 1285: Should we allow this job to run on control plane nodes?

xmudrii (changes requested):

addons/cni-cilium/cilium.yaml
WeirdMachine: No, they are only needed for hubble and hubble resources are not allowed to run on controle plane nodes.

xmudrii:
Can you please extend the release note to mention that Cilium is updated to a newer version?

pkg/templates/images/images.go
```diff
@@ -57,6 +57,7 @@ const (
 	HubbleUI
 	HubbleUIBackend
 	HubbleProxy
+	CertGen
```
xmudrii, line 60: Nit: can you please rename this to CiliumCertGen, so that we know it's a Cilium-related image?

xmudrii (approved):
/lgtm
/approve

PR #1757, opened 2022-01-26
add proposal for k1-for-devs

stroebitzer:

Having a discussion base for the project `kubeone for devs`

This is a rather quick draft missing a lot of technical details. It should be seen as a document to gather the requirements.


docs/proposals/20220126_k1-for-devs.md
```diff
@@ -0,0 +1,76 @@
...
+* Having a proper Git setup
+* Having GitOps tooling up and running
+* Having kubectl and other useful tools (kubectx, bash completion, ) installed
+* Having a tool in place for handling sensitive data
+
+## Implementation
+
+### IDE
+* Using [Eclipse Che](https://github.com/eclipse/che) as Cloud IDE. Eclipse Che is a workspace server integrating Theia as a default. 
+* Theia is a nodejs application which provides the IDE. Workspace management has to be provided by Eclipse Che.
```
scheeles, line 31: For the future, I think we should consider VS Code as the default but still need some development https://github.com/eclipse/che/issues/20500

stroebitzer: Yes we should definitely consider something different than Theia. I had bad experiences with it due to nodejs dependency hell.

docs/proposals/20220126_k1-for-devs.md
```diff
@@ -0,0 +1,76 @@
...
+* Having a tool in place for handling sensitive data
+
+## Implementation
+
+### IDE
+* Using [Eclipse Che](https://github.com/eclipse/che) as Cloud IDE. Eclipse Che is a workspace server integrating Theia as a default. 
+* Theia is a nodejs application which provides the IDE. Workspace management has to be provided by Eclipse Che. 
+* Eclipse Che is holding all the workspace data and all user-specific file-based configurations, eg the .gitconfig
+
+### Terminal integrations
```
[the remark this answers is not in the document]
stroebitzer: @archups @nerdeveloper can you please take a look at it and give some info about your experiences

docs/proposals/20220126_k1-for-devs.md
```diff
@@ -0,0 +1,76 @@
+# KubeOne for Developers
```
xmudrii, line 1: This proposal is missing some key sections:

* Values and alternatives -- the key question I'd like to see answered is why would someone want to use this over GitPod and GitHub Codespaces? Those two products are also cheaper than a Kubernetes cluster created by KubeOne on any cloud provider, so there's that too.
* Ownership -- who is going to own and maintain the project?
* Location -- where is this addon going to live? This depends on the ownership question as well. If this is to be maintained by a dedicated team then it should live in another repo.

In my opinion, everyone should be able to use KubeOne as a building block. In that aspect, I'd like this proposal to focus on what we can do to make implementing this possible. I see two key aspects:

* Improving usability and the infrastructure management experience
* Improving the addons mechanism

If that's the case, I'd like to see what we can do to improve those two aspects.

docs/proposals/20220126_k1-for-devs.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Goals
+* "Developer Batteries included"
+* Developers should get enabled to start working on the Kubernetes Cluster without having to deal with infrastructure setup problems
```
xmudrii, line 5: This is unclear and should be clarified.

It's just impossible to get a cluster on any cloud provider without infrastructure setup problems. You at least have to create an account, and in many cases, set up your account, configure billing, configure initial subnets and network gateways (especially on AWS), etc...

Therefore, what are the problems that we should solve?

docs/proposals/20220126_k1-for-devs.md
```diff
@@ -0,0 +1,76 @@
...
+### Visualization of the Kubernetes Cluster
+* Either via enabling the Kubernetes Dashboard or via the [VSCode Kubernetes AddOn](https://marketplace.visualstudio.com/items?itemName=ms-kubernetes-tools.vscode-kubernetes-tools) (if that is doable in Theia)
+
+### Container Registry
+* Open Source Container Registry should become its own KubeOne AddOn. Harbor should be supported, may other Open Source Container Registries can be considered.
+
+### GitOps
+* FluxCD and/or ArgoCD should get supported. They should become its own KubeOne AddOn. Their controllers should get installed into the KubeOne cluster. We should take care that the Git Repo holding the yaml files for the Kubernetes Cluster can be configured easily.
+
+### Vault
```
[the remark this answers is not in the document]
xmudrii: TBH, I wouldn't consider Vault at all. Vault is in fact a very complex component if you want to set it up in a proper way. You mostly like want to integrate with your cloud provider KMS solution, secure your keys, etc... Abstracting that away from the user can create a huge security hole.

xmudrii (changes requested):

docs/proposals/20220126_k1-for-devs.md
stroebitzer: **Values and alternatives** Yes, I think we should spent some time looking for alternatives on the market before starting implementation. Concerning GitPod and CodeSpaces: I am not sure if those are usable on private clusters. Those tools are running on some GitHub infra and not in a dedicated cluster. IMO the idea is to add functionality to clusters created via KubeOne. Which you may can do with GitPod and CodeSpaces via connecting them to your Kubernetes Cluster but this will not be doable all the time (private clusters) and also has some security implications.

**Ownership & Location** Good question, we should discuss this, maybe @scheeles has some input here

docs/proposals/20220126_k1-for-devs.md
stroebitzer: I think this is a different problem you are talking. Yes, you are right you need that stuff to get your cluster via KubeOne up and running. 

The problem which we want to solve with this proposal is more about: OK, now I have my running cluster via KubeOne, what do I need to do to have a cool developer experience on it. By developer I mean the "enduser-developer", the one writing apps on the already existing Kubernetes Cluster.

docs/proposals/20220126_k1-for-devs.md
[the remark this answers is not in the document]
nerdeveloper: I have heard about it, I am not entirely sure if tilt and entando do the same thing. I could take tilt for a spin and check the developer experience.

xmudrii:
Given that there's no activity on this PR for a long time, I'll go ahead and close it. Please feel free to reopen it or create a new one if you want to proceed with this proposal.
/close

PR #1750, opened 2022-01-23
Generate and approve CSRs for conntrol-plane and static workers nodes

kron4eg:
**What this PR does / why we need it**:
Fixes the issue that kubelets on control-plane and static workers nodes are service requests under self-signed certificates.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1093

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Generate and approve CSRs for conntrol-plane and static workers nodes
```

PR #1758, opened 2022-01-27
Small CSR tweaks

kron4eg:
Addressing some review comments from https://github.com/kubermatic/kubeone/pull/1750#pullrequestreview-864477584

```release-note
NONE
```

### DIFF
--- a/pkg/tasks/certs.go
+++ b/pkg/tasks/certs.go
@@ -235,26 +235,26 @@ func approvePendingCSR(s *state.State, node *kubeoneapi.HostConfig, conn ssh.Con
 
 func validateCSR(spec certificatesv1.CertificateSigningRequestSpec, node *kubeoneapi.HostConfig) error {
 	if fmt.Sprintf("%s:%s", nodeUser, node.Hostname) != spec.Username {
-		return errors.New("")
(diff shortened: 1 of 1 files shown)


ahmedwaleedmalik (approved):
/lgtm 
/approve

pkg/tasks/certs.go
```diff
@@ -235,26 +234,26 @@ func approvePendingCSR(s *state.State, node *kubeoneapi.HostConfig, conn ssh.Con
 
 func validateCSR(spec certificatesv1.CertificateSigningRequestSpec, node *kubeoneapi.HostConfig) error {
 	if fmt.Sprintf("%s:%s", nodeUser, node.Hostname) != spec.Username {
-		return errors.New("")
+		return fmt.Errorf("CSR username %q and node hostname %q do not match", spec.Username, node.Hostname)
```
xmudrii, line 237: I guess this should be `errors.New` instead.

kron4eg: I like Errorf better since it manifests that "format" is used

xmudrii: There’s also Errorf in the errors package, so we get to use same package everywhere. Just a nit though.

xmudrii (approved):

PR #1760, opened 2022-01-28
Additional linters

kron4eg:
To enforce better code style / practices

Fixes #

```release-note
NONE
```


.golangci.yml
```diff
@@ -25,24 +35,46 @@ linters:
   - gosec
   - gosimple
   - govet
+  - importas
   - ineffassign
+  - makezero
   - misspell
   - nakedret
+  - nilnil
+  - nlreturn
```
xmudrii, line 44: What's the point of this linter? To be honest, it seems a bit annoying and I don't see any advantages.

.golangci.yml
```diff
@@ -13,10 +13,20 @@ run:
 linters:
   disable-all: true
```
xmudrii, line 14: Why don't we reverse -- enable all linters, but disable ones that we don't want?

pkg/containerruntime/docker_config.go
```diff
@@ -33,8 +33,8 @@ type dockerConfig struct {
 	RegistryMirrors    []string          `json:"registry-mirrors,omitempty"`
 }
 
-func marshalDockerConfig(cluster *kubeone.KubeOneCluster) (string, error) {
-	// Parse log max size to ensure that it has the correct units
+// Parse log max size to ensure that it has the correct units
```
xmudrii, line 36: This comment is used to describe the `logSize` logic, not the function.

xmudrii (changes requested):
I strongly believe we should get rid of that new line linter.

.golangci.yml
kron4eg: to avoid surprises, golangci-lint changes its set of linters from version to version.

.golangci.yml
kron4eg: Separated "return" from the rest of the function body increases readability. And since we read the code much more often compared to write, it's crucial to make it as readable as possible. nlreturn just another microstep in that direction.

pkg/containerruntime/docker_config.go
kron4eg: yeah, you're right, incorrect rebase

xmudrii (approved):
/lgtm
/approve

PR #1767, opened 2022-02-01
Change baseurl to vault.centos.org for CentOS 8

xmudrii:

**[ATTENTION NEEDED]** CentOS 8 has reached End-Of-Life (EOL) on January 31st, 2022. It will no longer receive any updates (including security updates). Support for CentOS 8 in KubeOne is deprecated and will be removed in a future release. We strongly recommend migrating to another operating system or CentOS distribution as soon as possible.

CentOS 8 has reached EOL yesterday, and therefore, the base URL for repositories has been changed to `vault.centos.org`.

Without this change, CentOS 8 nodes cannot be provisioned due to the following error:

```
sudo yum install -y yum-plugin-versionlock device-mapper-persistent-data lvm2 conntrack-tools ebtables socat iproute-tc rsync
Error: Failed to download metadata for repo 'AppStream': Cannot prepare internal mirrorlist: No URLs in mirrorlist
```

References:

* https://www.centos.org/centos-linux-eol/
* https://stackoverflow.com/questions/70926799/centos-through-vm-no-urls-in-mirrorlist

```release-note
[ATTENTION NEEDED] CentOS 8 has reached End-Of-Life (EOL) on January 31st, 2022. It will no longer receive any updates (including security updates). Support for CentOS 8 in KubeOne is deprecated and will be removed in a future release. We strongly recommend migrating to another operating system or CentOS distribution as soon as possible. 
Change baseurl to vault.centos.org for CentOS 8
```

/assign @kron4eg

### DIFF
--- a/pkg/scripts/os_centos.go
+++ b/pkg/scripts/os_centos.go
@@ -52,6 +52,12 @@ gpgcheck=1
 repo_gpgcheck=0
 gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg
 EOF
+
+source /etc/os-release
+if [ "$ID" == "centos" ] && [ "$VERSION_ID" == "8" ]; then
+	sudo sed -i 's/mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-*
+	sudo sed -i 's|#baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g' /etc/yum.repos.d/CentOS-*
+fi
 {{ end }}
 
 sudo yum install -y \
--- a/examples/terraform/aws/output.tf
+++ b/examples/terraform/aws/output.tf
@@ -74,6 +74,9 @@ output "kubeone_workers" {
     "${var.cluster_name}-${local.zoneA}" = {
       replicas = var.initial_machinedeployment_replicas
       providerSpec = {
+        annotations = {
+          "k8c.io/operating-system-profile" = var.ami_filters[var.os].osp_name
+        }
         sshPublicKeys   = local.worker_deploy_ssh_key
         operatingSystem = local.worker_os
         operatingSystemSpec = {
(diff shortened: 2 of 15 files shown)


pkg/scripts/os_centos.go
```diff
@@ -52,6 +52,13 @@ gpgcheck=1
 repo_gpgcheck=0
 gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg
 EOF
+
+os_name=$(cat /etc/os-release | head -n1 | cut -d \" -f2)
+os_version=$(cat /etc/os-release | head -n2 | tail -1 | cut -d \" -f2)
```
kron4eg, line 57: it's more robust to "source" that file since line number can be different you never know for sure what would you get.

pkg/scripts/os_centos.go
```diff
@@ -52,6 +52,13 @@ gpgcheck=1
...
+if [ "$os_name" == "CentOS Linux" ] && [ "$os_version" == "8 (Core)" ]; then
+	sudo sed -i 's/mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-*
+	sudo sed -i 's|#baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g' /etc/yum.repos.d/CentOS-*
+fi
```
kron4eg, line 61: I have mixed fillings about this:
* it has to be ported to MC and OSM
* this is band-aid
* it's prologues the agony of already dead centos8 :zombie: 

What would help is if kubeone learns to recognizing other RHEL rebuilds and act on them as they were "just centos". There are two non-contradicting ways to do this:
* open `HostConfig.OperatingSystem` to be set from the config/terraform by the user (like we did with Hostname)
* if `HostConfig.OperatingSystem` only then use automatic detection, and at this step we need to teach K1 to recognize all those RockyLinux/OracleLinux/AlmaLinux and other RHEL rebuilds.

kron4eg: Since we already set OS anyway in the machineDeployments, opening `HostConfig.OperatingSystem` to be set from the config seems like the easiest way to help those in transition.

kron4eg (approved):

kron4eg (approved):

PR #1808, opened 2022-02-04
Use semver constraints on Kubernetes version for upper and lower bounds

### LINKED ISSUE #1807: Fail validation for unsupported newer Kubernetes versions

KubeOne has an official support matrix for Kubernetes versions available in the docs: https://docs.kubermatic.com/kubeone/v1.3/architecture/compatibility/#supported-kubernetes-versions (for 1.3) and https://docs.kubermatic.com/kubeone/master/architecture/compatibility/#supported-kubernetes-versions (for the upcoming 1.4 release).

We should "codify" that and validate the Kubernetes version in the config file with an "upper limit" version as well. Right now, we only validate that the version isn't too old.

Prevent deployment of unsupported Kubernetes versions that can break clusters in all sorts of (non-obvious) ways.

embik:

This adds validation against two semver constraints that can easily be bumped from the top of the file when adding new Kubernetes versions.

Fixes #1807

```release-note
Validate Kubernetes version against supported versions constraints
```

### DIFF
--- a/pkg/apis/kubeone/validation/validation.go
+++ b/pkg/apis/kubeone/validation/validation.go
@@ -40,6 +40,11 @@ const (
 	upperVersionConstraint = "<= 1.23"
 )
 
+var (
+	lowerConstraint = mustParseConstraint(lowerVersionConstraint)
+	upperConstraint = mustParseConstraint(upperVersionConstraint)
+)
+
 // ValidateKubeOneCluster validates the KubeOneCluster object
 func ValidateKubeOneCluster(c kubeoneapi.KubeOneCluster) field.ErrorList {
 	allErrs := field.ErrorList{}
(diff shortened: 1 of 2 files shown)


pkg/apis/kubeone/validation/validation.go
```diff
@@ -253,13 +260,36 @@ func ValidateVersionConfig(version kubeoneapi.VersionConfig, fldPath *field.Path
...
 	}
-	if v.Major() != 1 || v.Minor() < 19 {
-		allErrs = append(allErrs, field.Invalid(fldPath.Child("kubernetes"), version, "kubernetes versions lower than 1.19 are not supported. You need to use an older KubeOne version to upgrade your cluster to v1.19. Please refer to the Compatibility section of docs for more details."))
-	}
+
 	if strings.HasPrefix(version.Kubernetes, "v") {
 		allErrs = append(allErrs, field.Invalid(fldPath.Child("kubernetes"), version, ".versions.kubernetes can't start with a leading 'v'"))
 	}
 
+	lowerConstraint, err := semver.NewConstraint(lowerVersionConstraint)
```
kron4eg, line 268: I'd replace this with semver.MustParse or something and saved the result in the package variable. So that later you wouldn't need to return parsing error and check for len(errs) == 0

embik: Good suggestion, thanks!

embik: Although I don't think there is a `MustParse` equivalent for constraints, so maybe this doesn't work.

kron4eg: microscopic wrapper would do it easily

```
func mustParseConstraint(constraint string) *semver.Constraint {
  result, err := semver.ParseConstraint(constraint)
  if err != nil {
    panic(err)
  }
  return result
}
```

embik: You're of course right, thanks. Added that.

kron4eg (approved):
/approve

admitting this to 1.4 release

xmudrii:
@kron4eg @embik Do we also want to cherry-pick this to the release/v1.3 branch?

kron4eg:
I'm not sure, most likely no since that would be an unexpected novelty in 1.3.

embik:
I agree with @kron4eg, it's unexpected in a patch release for 1.3. IMHO it should not be cherry-picked. It's "just" codifying the contract we already have in place through the support documentation.

PR #1809, opened 2022-02-04
feat: Provide operating system from config

kron4eg:
Many RHEL rebuilds are floating around and they all can be treated equal as "centos" for example.

```release-note
Provide operating system from config
```


examples/terraform/aws/variables.tf
```diff
@@ -157,9 +157,9 @@ variable "ami_filters" {
     }
 
     centos = {
-      owners     = ["125523088429"] # CentOS
-      image_name = ["CentOS 8.* x86_64"]
-      osp_name   = "osp-centos8"
+      owners     = ["792107900819"] # RockyLinux
+      image_name = ["Rocky-8-ec2-*.x86_64"]
```
kron4eg, line 161: Live example of RHEL-rebuild is being used as a substitution for CentOS8

kron4eg:
@xmudrii rebased the PR against current master

xmudrii (approved):

PR #1826, opened 2022-02-09
Allow docker as container-runtime up to k8s v1.24

ahmedwaleedmalik:

Allow docker as container-runtime up to k8s v1.24

Fixes #

```release-note
Allow docker as container-runtime up to k8s v1.24
```

### DIFF
--- a/pkg/tasks/probes.go
+++ b/pkg/tasks/probes.go
@@ -90,7 +90,9 @@ func safeguard(s *state.State) error {
 
 		if nodesContainerRuntime != configuredClusterContainerRuntime {
 			errMsg := "Migration is not supported yet"
-			if cr.Containerd != nil {
+			if nodesContainerRuntime == "docker" {
+				errMsg = "Support for docker will be removed with Kubernetes 1.24 release. It is recommended to switch to containerd as container runtime using `kubeone migrate to-containerd`. To continue using Docker please specify ContainerRuntime explicitly in KubeOneCluster manifest"
+			} else if cr.Containerd != nil {
 				errMsg = "Use `kubeone migrate to-containerd`"
 			}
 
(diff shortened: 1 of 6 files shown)


pkg/tasks/probes.go
```diff
@@ -90,7 +90,9 @@ func safeguard(s *state.State) error {
 
 		if nodesContainerRuntime != configuredClusterContainerRuntime {
 			errMsg := "Migration is not supported yet"
-			if cr.Containerd != nil {
+			if nodesContainerRuntime == "docker" {
+				errMsg = "Support for docker will be removed with Kubernetes 1.24 release. It is recommended to switch to containerd as container runtime using `kubeone migrate to-containerd`. To continue using Docker please specify ContainerRuntime explicitly in KubeOneCluster manifest"
```
ahmedwaleedmalik, line 94: Let me know what you think of this @xmudrii @kron4eg, I'm divided on this one. This is the scenario where a user running docker is trying to upgrade to k8s v.1.22.x. If he hasn't explicitly specified the `containerRuntime` in cluster config then it's a little hard, still do-able maybe, to determine if `containerd` was a result of defaulting or was explicitly set here.

kron4eg: sounds good

kron4eg (approved):
/approve

/hold
/assign @xmudrii to review

pkg/tasks/probes.go
xmudrii: Sounds good.

pkg/tasks/probes.go
```diff
@@ -174,10 +176,10 @@ func runProbes(s *state.State) error {
 		return nil
 	}
 
-	gteKube121Condition, _ := semver.NewConstraint(">= 1.21")
+	gteKube124Condition, _ := semver.NewConstraint(">= 1.24")
```
xmudrii, line 179: If I'm not mistaken, this is doing defaulting. As we discussed, we want to containerd by default for new 1.21+ clusters, so this shouldn't be changed.

I'll double check and get back to you.

xmudrii (changes requested):

PR #1831, opened 2022-02-10
Install CSI drivers for k8s 1.23 un-conditionally

ahmedwaleedmalik:
Signed-off-by: Waleed Malik <ahmedwaleedmalik@gmail.com>

**What this PR does / why we need it**:
Install available CSI driver for k8s 1.23 un-conditionally

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Install available CSI driver for k8s 1.23 un-conditionally
```

PR #1853, opened 2022-02-15
Fix the default-storage-class addon

xmudrii:

We started unconditionally deploying the CSI driver in some cases with #1831. This was needed because Kubernetes 1.23 enabled CSIMigration for some providers by default, requiring the CSI driver to be deployed. Namely:

* Deploy CSI driver on Azure, AWS, and vSphere clusters if the external CCM is enabled or if the Kubernetes version is >= 1.23
* Deploy CSI driver on DigitalOcean, Hetzner, Nutanix, and OpenStack unconditionally
  * OpenStack has the CSIMigration enabled by default since Kubernetes 1.18
  * Other providers don't have the in-tree provider, so the CSI driver is required to use volumes

However, we haven't adjusted the `default-storage-class` addon used to deploy StorageClasses, so it might happen that the StorageClass is not deployed even if it should be deployed.

On top of that, if a cluster is migrated from the in-tree provider to the external CCM, there will be two StorageClasses, and both SCs will be default. This is breaking some workflows, as Kubernetes will fail to create a PV.

The first issue is solved by adjusting the criteria when we deploy the StorageClass.

The second issue (two default StorageClasses) is a more complicated one. We don't want to remove a StorageClass because we don't know is it going to break existing PVCs/PVs. We need to somehow reconcile the old StorageClass (the one that uses the in-tree provider) to remove the default annotation. Right now, we are reconciling only one StorageClass, depending on if the CSI driver is deployed or not.

To solve the issue, we now deploy both StorageClasses unconditionally if the CSI driver is deployed. We are checking if the CSI driver is deployed, and based on that set the default annotation. With this approach, we will have one StorageClass that's not needed, but both StorageClasses are effectively doing the same thing since the CSIMigration will translate all calls from the in-tree provider to the CSI driver.

```release-note
Deploy the StorageClass based on the CSI driver if the CSI driver is deployed
Always deploy the StorageClass based on the in-tree provider. This StorageClass will use the CSI driver if the CSI migration is enabled (default for all providers if the Kubernetes version is >= 1.23 or if the external CCM is used)
```

### DIFF
--- a/addons/default-storage-class/storage-class.yaml
+++ b/addons/default-storage-class/storage-class.yaml
@@ -1,5 +1,8 @@
+{{ $version := semver .Config.Versions.Kubernetes }}
+{{ $csi := or .Config.CloudProvider.External (ge $version.Minor 23) }}
+
 {{ if eq .Config.CloudProvider.CloudProviderName "azure" }}
-{{ if .Config.CloudProvider.External }}
+{{ if $csi }}
 apiVersion: storage.k8s.io/v1
 kind: StorageClass
 metadata:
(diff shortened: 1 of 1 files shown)


moadqassem (approved):
/approve
/lgtm

xmudrii:
/hold
for manual testing

PR #1855, opened 2022-02-15
Fix Terraform configs for AWS

xmudrii:

Fix Terraform configs for AWS:

* Don't default the SSH username because it's supposed to be determined automatically
* Properly default the `worker_os` variable if Amazon Linux 2 is used
  * Terraform configs expect `amzn` to be provided as the operating system, while machine-controller expects `amzn2`

```release-note
NONE
```


xrstf (approved):
/approve

xrstf (approved):
/approve

PR #1856, opened 2022-02-15
e2e: allow amzn2 for worker nodes

xmudrii:

Terraform configs for AWS use `amzn`, while machine-controller uses `amzn2`. We need to make sure both are considered valid by our E2E tests.

```release-note
NONE
```


xrstf (approved):
/approve

PR #1871, opened 2022-02-24
Fix missing CRDs for cinder csi

WeirdMachine:
Adds snapshot CRDs for cinder snapshot controller that is deployed as part of the csi.

```release-note
Fix missing snapshot CRDs for Openstack CSI
```


kron4eg (approved):
/approve

PR #1874, opened 2022-02-26
Add the kubeone config merge command

### LINKED ISSUE #1867: Create a command to print the KubeOneCluster manifest merged with the Terraform output

We should add a command to print the complete KubeOneCluster manifest, i.e. merge the provided manifest with the Terraform output.

https://github.com/kubermatic/kubermatic/issues/8950

xmudrii:
**What type of PR is this?**

/kind feature

This PR implements the `kubeone config merge` command used to merge the versioned KubeOneCluster manifest with the Terraform output. The merged config is printed to stdout.

The workflow is the following:

* Load the KubeOneCluster manifest
* Determine the original API version
* Merge with Terraform output, run defaulting and validation (this returns the internal representation)
* Convert the internal representation to the versioned representation (uses the same API version as the original manifest)
* Print the merged manifest to stdout

NB: Validation works only with the internal representation, so we can't avoid conversion from versioned to internal and then again to versioned unless we drop validation. However, I find validation useful, and I don't think this is something problematic (maybe just a little inconvenient).

Fixes #1867

```release-note
Implement the `kubeone config merge` command used to merge the KubeOneCluster manifest with the Terraform output. The resulting (merged) manifest is printed to stdout.
```


kron4eg (approved):
👍

PR #1880, opened 2022-03-04
Update secret name for backup-restic addon to backup-restic-s3-credentials

ahmedwaleedmalik:

**What type of PR is this?**

/kind cleanup

Currently, there is a conflict between the `s3-credentials` secret created by KKP and KubeOne. Both use the same secret name and deploy it in the same namespace i.e. kube-system. But the keys within the secret are different for both of them.

In a bid to solve this issue, we are renaming the secret in KubeOne.

Fixes #

```release-note
[ACTION REQUIRED] Update secret name for backup-restic addon to kubeone-backups-credentials. Manual migration steps are needed for users running KKP on top of a KubeOne installation and using both backup-restic addon from KubeOne and s3-exporter from KKP. Ensure that the `s3-credentials` Secret with keys `ACCESS_KEY_ID` and `SECRET_ACCESS_KEY` exists in `kube-system` namespace and doesn't have the label `kubeone.io/addon:`. Remove the label if it exists. Else s3-exporter won't be functional.
```


addons/backups-restic/backups-restic.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: Secret
 metadata:
-  name: s3-credentials
+  name: backup-restic-s3-credentials
```
ahmedwaleedmalik, line 4: @kron4eg @xmudrii This or more specific `kubeone-backup-restic-credentials`?

xmudrii: I was planning to go with `kubeone-backups-credentials`.

ahmedwaleedmalik: That sounds fine as well, my intention behind not adding the `kubeone-` prefix but we already use that prefix for other credentials. So let's do the same here.

ahmedwaleedmalik: Updated PTAL

xmudrii:
@ahmedwaleedmalik Can you please add `[ACTION REQUIRED]` release note and mention that users using both this addon and s3-exporter should take additional steps?

ahmedwaleedmalik:
@xmudrii Do they really have to? As in when they'll update kubeone, the new dedicated secret will be used for this addon. And the de-coupling would be removed.

Same will happen if they update KKP first.

ahmedwaleedmalik:
Are you talking about the case where a K1 update would prune/remove the existing `s3-credentials`?

ahmedwaleedmalik:
Are you talking about the case where a K1 update would prune/remove the existing `s3-credentials`?

xmudrii:
@ahmedwaleedmalik Yes, I was talking about pruning the s3-credentials Secret.

pkg/kubeconfig/kubeconfig.go
```diff
@@ -76,18 +76,10 @@ func BuildKubernetesClientset(s *state.State) error {
 
 	s.RESTConfig.Dial = tunn.TunnelTo
 
-	return errors.WithStack(HackIssue321InitDynamicClient(s))
-}
-
-// HackIssue321InitDynamicClient initialize controller-runtime/client
-// name comes from: https://github.com/kubernetes-sigs/controller-runtime/issues/321
-func HackIssue321InitDynamicClient(s *state.State) error {
```
kron4eg, line 84: This function is a reminder and explanation of "why it's done this way", please don't remove it

ahmedwaleedmalik: It's not required anymore and actually results in a nil pointer exception if k8s was not properly installed.

kron4eg (approved):

PR #1882, opened 2022-03-08
Error handling refactoring

### LINKED ISSUE #1869: Refactor the error handling

The errors returned by KubeOne don't tell where's the issue exactly and how to fix it. We need to ensure that errors are descriptive enough, so users know what failed and how to fix it. Also, we should ensure that external tools integrating with KubeOne can more easily parse the errors.

* https://github.com/kubermatic/kubermatic/issues/8950

kron4eg:
**What type of PR is this?**
/kind feature

This PR is only first in series. XRef #1869

```release-note
Error reporting in CLI now exists with different codes for different error reasons
```

### DIFF
--- a/pkg/cmd/kubeconfig.go
+++ b/pkg/cmd/kubeconfig.go
@@ -38,7 +38,8 @@ func kubeconfigCmd(rootFlags *pflag.FlagSet) *cobra.Command {
 			This command takes KubeOne manifest which contains information about hosts. It's possible to source information about
 			hosts from Terraform output, using the '--tfjson' flag.
 		`),
-		Example: `kubeone kubeconfig -m mycluster.yaml -t terraformoutput.json`,
+		Example:       `kubeone kubeconfig -m mycluster.yaml -t terraformoutput.json`,
+		SilenceErrors: true,
 		RunE: func(_ *cobra.Command, args []string) error {
 			gopts, err := persistentGlobalOptions(rootFlags)
 			if err != nil {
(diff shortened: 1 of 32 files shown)


pkg/cmd/kubeconfig.go
```diff
@@ -38,7 +38,8 @@ func kubeconfigCmd(rootFlags *pflag.FlagSet) *cobra.Command {
 			This command takes KubeOne manifest which contains information about hosts. It's possible to source information about
 			hosts from Terraform output, using the '--tfjson' flag.
 		`),
-		Example: `kubeone kubeconfig -m mycluster.yaml -t terraformoutput.json`,
+		Example:       `kubeone kubeconfig -m mycluster.yaml -t terraformoutput.json`,
+		SilenceErrors: true,
```
xmudrii, line 42: What is this option used for?

kron4eg: When it's set to default (by default), cobra will automatically report errors. And then we report error, it's the same error printed out two times.

```
        // inside the cobra
	if err != nil {
		// If found parse to a subcommand and then failed, talk about the subcommand
		if cmd != nil {
			c = cmd
		}
		if !c.SilenceErrors {
			c.PrintErrln("Error:", err.Error())
			c.PrintErrf("Run '%v --help' for usage.\n", c.CommandPath())
		}
		return c, err
	}
```

kron4eg: However I want to control the way to output so I had to `SilenceErrors: true`, to skip automatic print.

xmudrii (approved):

PR #1887, opened 2022-03-09
Approve pending CSRs when upgrading nodes

xmudrii:
**What type of PR is this?**

/kind bug

Approve pending CSRs when upgrading control plane and static worker nodes. It seems that upgrading the cluster causes new CSRs to be created. If we don't approve those CSRs, the certificates will not work. This will cause errors such as:

```
Error from server: Get "https://10.0.42.12:10250/containerLogs/kube-system/machine-controller-7b647fdc45-tfskm/machine-controller?follow=true": remote error: tls: internal error
```

```release-note
Approve pending CSRs when upgrading control plane and static worker nodes
```

/assign @kron4eg

### DIFF
--- a/pkg/tasks/upgrade_follower.go
+++ b/pkg/tasks/upgrade_follower.go
@@ -83,5 +83,5 @@ func upgradeFollowerExecutor(s *state.State, node *kubeoneapi.HostConfig, conn s
 		return errors.Wrap(err, "failed to unlabel follower control plane node")
 	}
 
-	return nil
+	return approvePendingCSR(s, node, conn)
 }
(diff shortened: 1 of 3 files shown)


kron4eg (approved):
/approve

fwiedmann:
@xmudrii  @kron4eg  Hey Hey, is this fix already present in the current 1.4 release? We're currently facing the error described in the PR description? How can we fix this in our current running cluster? :)

kron4eg:
@fwiedmann this is in the release branch but hasn't been released yet, You could try to build it from `release/v1.4`.

fwiedmann:
Hey @xmudrii, @xmudrii we built kubeone from source on release branch `release/v1.4.`. No we get the error with the image pull backoff for the core DNS image. We upgraded yesterday from kubeone 1.3 to version 1.4 to mitigate this issue. Which it also did.

Followed by your suggestion to use the `release/v1.4` we could fix the CSR error, but again facing the problem with the coreDNS image.

PR #1906, opened 2022-03-30
PDB Objects for all addons Deployments

### LINKED ISSUE #1891: Cluster-Autoscaler doesn't scale down properly
**What happened**:

[As discussed in Slack](https://kubernetes.slack.com/archives/CNEV2UMT7/p1647361805511139), I'm opening an issue here now.

I am using KubeOne with Hetzner Cloud and the `cluster-autoscaler` is not able to scale down (scaling up works fine) my cluster. The reason for this seems to be several pods (`metrics-server`, `hcloud-cloud-controller-manager`, `hcloud-csi-controller`, `coredns/kube-dns` and `cluster-autoscaler` itself) from `kube-system`, which are non DaemonSet-Pods or do not have a PodDisruptionBudget and usually only have 1 replica.

Now, I could solve this using two different solutions:
- Scale the deployments up and create a PDB
- Run them as a DaemonSet, maybe only on the Master Nodes

Both of these options require me to override several embedded KubeOne Addons and to me it seems that this wouldn't be a feasible solution. Or is it?

I could also write a PR for this, but without autoscaling, there wouldn't be a need to always run more than one instance of those tools. :thinking:

`cluster-autoscaler` should scale down the cluster

Make `cluster-autoscaler` scale up, then reschedule one of the aforementioned components to the new node and remove all other stuff running on it so that it should usually scale down.

`cluster-autoscaler` will log a message like this:

```
Fast evaluation: node pool-1-c76dd79df-fnlz6 cannot be removed: non-daemonset, non-mirrored, non-pdb-assigned kube-system pod present: hcloud-csi-controller-0
```

(As [written in Slack](https://kubernetes.slack.com/archives/CNEV2UMT7/p1647420324566079?thread_ts=1647361805.511139&cid=CNEV2UMT7)): At least for `metrics-server`, the cluster-autoscaler docs state that it should ["be left alone"](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-to-set-pdbs-to-enable-ca-to-move-kube-system-pods), so downtime due to a unrestrictive PDB shouldn't be considered.

KubeOne version (`kubeone version`): 1.4.0
Operating system: Arch Linux (local machine) / Gitlab CI (`FROM hashicorp/terraform:1.1.6`)

kron4eg:
Introducing PodDisruptionBudget object for all addons Deployments, in order to let cluster-autoscaler to drain worker nodes with pods running in kube-system namespace.

Fixes #1891

/kind bug

```release-note
PDB Objects for all addons Deployments 
```

### DIFF
--- a/addons/cni-cilium/poddisruptionbudgets.yaml
+++ b/addons/cni-cilium/poddisruptionbudgets.yaml
@@ -1,0 +1,45 @@
+{{ $version := semver .Config.Versions.Kubernetes }}
+{{- if ge $version.Minor 21 }}
+apiVersion: policy/v1
+{{- else }}
+apiVersion: policy/v1beta1
+{{- end }}
+kind: PodDisruptionBudget
+metadata:
+  name: cilium-operator
+  namespace: kube-system
+spec:
+  minAvailable: 1
+  selector:
+    matchLabels:
+      name: cilium-operator
+---
+{{- if ge $version.Minor 21 }}
+apiVersion: policy/v1
+{{- else }}
+apiVersion: policy/v1beta1
+{{- end }}
+kind: PodDisruptionBudget
+metadata:
+  name: hubble-relay
+  namespace: kube-system
+spec:
+  minAvailable: 1
+  selector:
+    matchLabels:
+      k8s-app: hubble-relay
+---
+{{- if ge $version.Minor 21 }}
+apiVersion: policy/v1
+{{- else }}
+apiVersion: policy/v1beta1
+{{- end }}
+kind: PodDisruptionBudget
+metadata:
+  name: hubble-ui
+  namespace: kube-system
+spec:
+  minAvailable: 1
+  selector:
+    matchLabels:
+      k8s-app: hubble-ui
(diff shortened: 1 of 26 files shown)


addons/cni-cilium/poddisruptionbudgets.yaml
```diff
@@ -0,0 +1,45 @@
...
+{{- end }}
+kind: PodDisruptionBudget
+metadata:
+  name: hubble-relay
+  namespace: kube-system
+spec:
+  minAvailable: 1
+  selector:
+    matchLabels:
+      k8s-app: hubble-relay
```
xmudrii, line 30: Hubble is optional, so we should probably create PDB only if it's deployed.

kron4eg: Not that such "dangling" PDB objects would harm anyone, but anyway, surrounded them with the condition.

xmudrii (approved):
/lgtm
/approve

xmudrii:
The tests are legitimately broken.
/lgtm cancel

kron4eg:
@xmudrii I've changed PDBs, now they should work better.

xmudrii (approved):

PR #1910, opened 2022-03-31
Add (re-)rendering container runtime config to flatcar upgrades

### LINKED ISSUE #1901: containerd-config is not being updated on flatcar control plane nodes
**What happened**:
When changing the containerd config in the kubeone manifest the config is not being updated on flatcar control plane nodes.

The containerd config should be updated according to the changes in the manifest.

1. Run kubeone with following manifest to install kubernetes on flatcar nodes:
```
kind: KubeOneCluster
versions:
  kubernetes: v1.23.3
cloudProvider:
  aws: {}
containerRuntime:
  containerd: {}
    registries:
      "*":
        mirrors:
          - https://docker-hub-proxy.gec.io
addons:
  enable: false
```
2. Change the manifest and run `kubeone apply --force-upgrade`:
```
kind: KubeOneCluster
versions:
  kubernetes: v1.23.3
cloudProvider:
  aws: {}
containerRuntime:
  containerd: {}
    registries:
      "docker.io":
        mirrors:
          - https://my.dockerhub.proxy
addons:
  enable: false
```
3. Check `/etc/containerd/config.toml` on the control plane nodes.

1. On initial install the containerd config is written correctly. This bug only affects updates.
2. This bug is flatcar specific, it works as expected on ubuntu control plane nodes.
3. The machine-controller is correctly updated. After forcing recreations of the worker nodes the new ones will have the correct containerd configs.

KubeOne version (`kubeone version`): v1.4.0
Operating system: Linux
Provider you're deploying cluster on: aws
Operating system you're deploying on: flatcar

dermorz:
**What type of PR is this?**

/kind bug

This PR fixes a bug where it's only possible to render the containerd-config on control-plane nodes during the initial kubeone installation, but not re-render it during (forced) cluster upgrades using kubeone.

Fixes #1901 

```release-note
Regenerate container runtime configurations based on kubeone.yaml during control-plane upgrades on Flatcar Linux nodes, not only on the initial installation.
```

### DIFF
--- a/pkg/scripts/os_test.go
+++ b/pkg/scripts/os_test.go
@@ -457,6 +457,15 @@ func TestKubeadmFlatcar(t *testing.T) {
 				cluster: genCluster(withContainerd),
 			},
 		},
+		{
+			name: "with containerd with insecure registry",
+			args: args{
+				cluster: genCluster(
+					withContainerd,
+					withInsecureRegistry("127.0.0.1:5000"),
+				),
+			},
+		},
 	}
 
 	for _, tt := range tests {
@@ -572,7 +581,12 @@ func TestUpgradeKubeadmAndCNIAmazonLinux(t *testing.T) {
 func TestUpgradeKubeadmAndCNIFlatcar(t *testing.T) {
 	t.Parallel()
 
-	got, err := UpgradeKubeadmAndCNIFlatcar("v1.17.4")
+	c := genCluster(
+		withKubeVersion("v1.17.4"),
+		withContainerd,
+		withInsecureRegistry("127.0.0.1:5000"),
+	)
+	got, err := UpgradeKubeadmAndCNIFlatcar(&c)
 	if err != nil {
 		t.Errorf("UpgradeKubeadmAndCNIFlatcar() error = %v", err)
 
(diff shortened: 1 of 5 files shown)


pkg/scripts/os_test.go
```diff
@@ -572,7 +581,12 @@ func TestUpgradeKubeadmAndCNIAmazonLinux(t *testing.T) {
 func TestUpgradeKubeadmAndCNIFlatcar(t *testing.T) {
```
dermorz, line 581: Is it okay to extend these 2 test cases or should I create extra ones?

ahmedwaleedmalik: I think it's fine

ahmedwaleedmalik (approved):
/approve

pkg/scripts/testdata/TestKubeadmFlatcar-with_containerd_with_insecure_registry.golden
```diff
@@ -0,0 +1,153 @@
+set -xeu pipefail
```
wozniakjan, line 1: shouldn't this be?
```bash
set -xeuo pipefail
```
without `-o` at the end, pipefail gets ignored I think

PR #1923, opened 2022-04-04
Use Infof instead of Info for correct logging format

tboerger:
The logging was using `Info` instead of `Infof`within a single place which resulted in a bad formatted loginline like this:

```
time="10:06:56 UTC" level=info msg="No CSR found for node %q, assuming it was garbage-collectedworker-01" node=XXX.XXX.XXX.XXX
```

instead of the expected result:

```
time="10:06:56 UTC" level=info msg="No CSR found for node \"worker-01\", assuming it was garbage-collected" node=XXX.XXX.XXX.XXX
```

**What type of PR is this?**

/kind bug

It just fixes the logging format

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/tasks/certs.go
+++ b/pkg/tasks/certs.go
@@ -245,7 +245,7 @@ func approvePendingCSR(s *state.State, node *kubeoneapi.HostConfig, conn ssh.Con
 	}
 
 	if !csrFound {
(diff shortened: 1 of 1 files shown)


ahmedwaleedmalik (approved):
Thank you! 👍 

/approve

kron4eg:
Hm... I had impression we have an linter that should have warned us about such problem. :shrug:

PR #1941, opened 2022-04-11
Escape docker/containerd versions to avoid wildcard matching

### LINKED ISSUE #1908: Fail to install Kubernetes if sshUser uses zsh
The sshUser which gets used by KubeOne got zsh as shell, if this is the case KubeOne is not able to install dependencies like containerd.io because the scripts are lacking quotes. If you quote the version like https://github.com/kubermatic/kubeone/blob/8ca3b42189104dbc9935402e17f9088a6efddb5e/pkg/scripts/render.go#L151 which currently gets a value of `1.4.*` everything should work fine. 

**What happened**:

Kubernetes doesn't get installed

Kubernetes gets installed

Try to install Kubernetes with an sshUser that uses zsh as a shell

KubeOne version (`kubeone version`):  1.4.0
Operating system: Ubuntu 20.04

kron4eg:
**What type of PR is this?**
/kind bug

Fixes #1908

```release-note
Escape docker/containerd versions to avoid wildcard matching
```

### DIFF
--- a/pkg/scripts/funcs.go
+++ b/pkg/scripts/funcs.go
@@ -73,8 +73,8 @@ var (
 )
 
 const (
-	defaultDockerVersion           = "19.03.*"
-	latestDockerVersion            = "20.10.*"
-	defaultContainerdVersion       = "1.4.*"
-	defaultAmazonContainerdVersion = "1.4.*"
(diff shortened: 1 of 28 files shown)


ahmedwaleedmalik (approved):
/approve

/hold if @xmudrii wants to review this

PR #1942, opened 2022-04-11
Allow having no OIDC GroupsPrefix

### LINKED ISSUE #1940: OIDC groupsPrefix can't be set to an empty string
**What happened**:
If I set `features.openidConnect.config.groupsPrefix` to an empty string or delete the key. The default value (`oidc:`) will be used. 

I would expect that an empty string in the KubeOne config would result in an empty string in the manifest. But this is not the case. Maybe you could offer a dash (`-`) for an empty value like as for `usernamePrefix`. 

Also, I would interpret the doc string to mean that if I omit the key, it should not set a prefix.
```
# **If provided**, all groups will be prefixed with this value to prevent
# conflicts with other authentication strategies.
```

Look at >**What happened**<

`null` does not work as well -> default used.

KubeOne version (`kubeone version`): 1.4.1
Operating system: Rocky Linux 8.5
Provider you're deploying cluster on: on prem
Operating system you're deploying on: Rocky Linux 8.5

kron4eg:
/kind feature

Fixes #1940

API change is only in Go types, from YAML point of view nothing has been changed.

```release-note
 Allow having no OIDC GroupsPrefix
```

### DIFF
--- a/pkg/features/oidc.go
+++ b/pkg/features/oidc.go
@@ -31,7 +31,17 @@ func activateKubeadmOIDC(feature *kubeoneapi.OpenIDConnect, args *kubeadmargs.Ar
 	optionalMapSet(args.APIServer.ExtraArgs, "oidc-username-claim", feature.Config.UsernameClaim)
 	optionalMapSet(args.APIServer.ExtraArgs, "oidc-username-prefix", feature.Config.UsernamePrefix)
 	optionalMapSet(args.APIServer.ExtraArgs, "oidc-groups-claim", feature.Config.GroupsClaim)
-	optionalMapSet(args.APIServer.ExtraArgs, "oidc-groups-prefix", feature.Config.GroupsPrefix)
+
+	// While we have to handle the "-" value for GroupsPrefix, the same is done by the kube-apiserver for the
+	// UsernamePrefix. This is why there is no `if feature.Config.UsernamePrefix != "-" { ... }`
+	//
+	// Docs at the https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/ says:
+	// `--oidc-username-prefix string` If provided, all usernames will be prefixed with this value. If not provided,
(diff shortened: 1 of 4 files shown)


pkg/features/oidc.go
```diff
@@ -31,7 +31,9 @@ func activateKubeadmOIDC(feature *kubeoneapi.OpenIDConnect, args *kubeadmargs.Ar
 	optionalMapSet(args.APIServer.ExtraArgs, "oidc-username-claim", feature.Config.UsernameClaim)
 	optionalMapSet(args.APIServer.ExtraArgs, "oidc-username-prefix", feature.Config.UsernamePrefix)
 	optionalMapSet(args.APIServer.ExtraArgs, "oidc-groups-claim", feature.Config.GroupsClaim)
-	optionalMapSet(args.APIServer.ExtraArgs, "oidc-groups-prefix", feature.Config.GroupsPrefix)
+	if feature.Config.GroupsPrefix != "-" {
```
xmudrii, line 34: Shouldn't we also do this for `oidc-username-prefix`?

xmudrii (changes requested):
Please, also:

* Retitle the PR
* Update the PR description and the release note

pkg/features/oidc.go
kron4eg: No, it's handled by the kubeapi-server

kron4eg: The docs at https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/ says:

`--oidc-username-prefix string`
> If  provided, all usernames will be prefixed with this value. If not  provided, username claims other than 'email' are prefixed by the issuer  URL to avoid clashes. To skip any prefixing, provide the value '-'.

xmudrii: Can you please add a comment clarifying this?

kron4eg: done

xmudrii (approved):

PR #1944, opened 2022-04-12
Restart kubelet after containerd

### LINKED ISSUE #1847: Upgrading HA cluster without API server LB is failing because Docker is restarted
**What happened**:

Upgrading an HA cluster that doesn't use LB as an API endpoint is failing because we restart Docker which causes all containers to get terminated. In case the first node is used as an API endpoint, the cluster will be inaccessible until Kubelet detects that containers are terminated. However, that can take several minutes, which is quite long, and will also cause the upgrade process to fail/timeout.

The upgrade process works on HA clusters without an API server LB.

* Create a cluster on some cloud provider without creating an LB
* Try to upgrade the cluster

KubeOne version (`kubeone version`): 1.4.0

kron4eg:
**What type of PR is this?**
/kind bug

Fixes #1847

```release-note
Restart kubelet after containerd
```

### DIFF
--- a/pkg/scripts/render.go
+++ b/pkg/scripts/render.go
@@ -83,6 +83,10 @@ var (
 			{{ template "container-runtime-daemon-config" . }}
 			{{ template "containerd-systemd-setup" . -}}
 			sudo systemctl enable --now docker
+			if systemctl status kubelet 2>&1 > /dev/null; then
+				sudo systemctl restart kubelet
+				sleep 10
+			fi
 			`,
 			defaultDockerVersion,
 			latestDockerVersion,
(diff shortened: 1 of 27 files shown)


pkg/scripts/render.go
```diff
@@ -83,6 +83,7 @@ var (
 			{{ template "container-runtime-daemon-config" . }}
 			{{ template "containerd-systemd-setup" . -}}
 			sudo systemctl enable --now docker
+			systemctl status kubelet && sudo systemctl restart kubelet
```
xmudrii, line 86: Why do we need `systemctl status kubelet` here? To check does the service exists?

kron4eg: Yep, at the time if the initial install kubelet service doesn't exist yet.

xmudrii (approved):

PR #1948, opened 2022-04-13
Don't cordon and drain single node control-plane cluster

kron4eg:
**What type of PR is this?**
/kind bug

Single node cluster will fail to upgrade due to PDB resisting draining.

```release-note
Don't cordon and drain single node control-plane cluster
```


kron4eg:
Well, it turns our, after recent changes it works even with draining so I'm closing the PR without merge.

PR #1949, opened 2022-04-14
install.sh: Get latest version sorted by semver

### LINKED ISSUE #1936: install.sh does not always download the latest version
**What happened**:

The script `install.sh` downloaded an older version and not the newest version (1.3.4 instead of currently 1.4.1) because the releases page lists 1.3.4 as the "latest". The version cannot be changed using the script (e.g. pass a certain version directly via parameter). This leads to non working deployments due to incompatibilities of the used kubeone.yaml with older versions.

The script `install.sh` should download the latest version, or even better accept a parameter to download a specific version.

Execute `curl -sfL get.kubeone.io | sh` when the latest release on the github page is not the most current version.

There is a recommendation to use the same code as in the install.sh script to get the latest release in the documentation [in the documentation](https://docs.kubermatic.com/kubeone/v1.4/getting_kubeone/) inside the code snippet under "For a quick way to install the latest version of KubeOne, use the following commands:".

This might also be missleading in the (probably rare) occasions that the latest release on github is actually not the most current version.

KubeOne version (`kubeone version`): Desired: 1.4.1 Actual: 1.3.4
Operating system: Ubuntu

dermorz:
**What type of PR is this?**
/kind bug

Before this fix the top entry of the github release page was used as the "latest" version. This was not always correct, because on top of the list is always the release that happened last, not necessarily the latest version.

Fixes #1936 

```release-note
Fixed "latest version" in install.sh.
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -21,7 +21,7 @@ set -eu
 # What OS is used
 OS="$(uname | tr '[:upper:]' '[:lower:]')"
 # Find out what's the latest version
-VERSION="$(curl -s -H "Accept: application/vnd.github.v3+json" https://api.github.com/repos/kubermatic/kubeone/releases | grep -oP '"tag_name": "v\K([^"]*)' | sort -V | tail -1)"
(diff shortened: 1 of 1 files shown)


ahmedwaleedmalik:
/hold

Unfortunately, this doesn't work on Mac

dermorz:
> /hold
> 
> Unfortunately, this doesn't work on Mac

I will change it to a GNU/BSD compatible variant.

Issue #2010, opened 2022-04-29
Support Kubernetes 1.24


xmudrii:
This is a collector issue for Kubernetes 1.24 support in KubeOne. Kubernetes 1.24 is scheduled for May 3rd. The following tasks should be taken care of:

* [x] Check the Kubernetes changelog to ensure there are no breaking changes and removals
* [x] Update [the `kubeone-e2e` image](https://github.com/kubermatic/kubeone/tree/master/hack/images/kubeone-e2e) to add the needed Kubernetes test binaries (#2016)
* [x] Update [default admission controllers](https://github.com/kubermatic/kubeone/blob/master/pkg/kubeflags/data.go) if needed (not needed)
* [x] Add E2E tests (#2045)
* [x] Update daily periodics to use the latest Kubernetes release
* [x] Update [the Compatibility Matrix](https://docs.kubermatic.com/kubeone/master/architecture/compatibility/) (https://github.com/kubermatic/docs/pull/1066)
* [x] Create an issue to track updating [images](https://github.com/kubermatic/kubeone/blob/master/pkg/templates/images/images.go) (#2011)
* [x] Run the full conformance tests suite using [Sonobuoy](https://github.com/vmware-tanzu/sonobuoy)

PR #2024, opened 2022-05-09
Remove --network-plugin Kubelet flag for Kubernetes 1.24 clusters

xmudrii:
**What type of PR is this?**

/kind bug

* Remove `--network-plugin` Kubelet flag when migrating from Docker to containerd
  * New containerd clusters don't have this flag out of the box
* Remove `--network-plugin` Kubelet flag when upgrading from Kubernetes 1.23.x to 1.24.x
  * This is required because in earlier KubeOne versions we didn't remove the `--network-plugin` flag when migrating to containerd. Therefore, the flag is still present in the Kubelet config. Upgrading such a cluster to 1.24 causes Kubelet to crash.

This is required since `--network-plugin` and some other flags got removed in Kubernetes 1.24 (https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#feature).

xref #2010 

This is still not tested.

```release-note
Remove the `--network-plugin` Kubelet flag when migrating from Docker to containerd
Remove the `--network-plugin` Kubelet flag when upgrading from Kubernetes 1.23.x to 1.24.x
```

/assign @kron4eg

### DIFF
--- a/pkg/tasks/fix_network_plugin.go
+++ b/pkg/tasks/fix_network_plugin.go
@@ -17,10 +17,14 @@ limitations under the License.
 package tasks
 
 import (
-	"github.com/Masterminds/semver/v3"
-
 	kubeoneapi "k8c.io/kubeone/pkg/apis/kubeone"
+	"k8c.io/kubeone/pkg/semverutil"
 	"k8c.io/kubeone/pkg/state"
+)
+
+var (
+	version124 = semverutil.MustParseConstraint("1.24.*")
+	version123 = semverutil.MustParseConstraint("1.23.*")
 )
 
 // removeNetworkPluginFlagKubelet removes --network-plugin flag from Kubelet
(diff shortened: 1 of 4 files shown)


pkg/tasks/fix_network_plugin.go
```diff
@@ -0,0 +1,66 @@
...
+	"k8c.io/kubeone/pkg/state"
+)
+
+// removeNetworkPluginFlagKubelet removes --network-plugin flag from Kubelet
+// config because it has been removed in Kubernetes 1.24.
+// This function is executed only when upgrading cluster from 1.23 to 1.24.
+// TODO: Remove this function after dropping support for Kubernetes 1.23.
+func removeNetworkPluginFlagKubelet(s *state.State, node kubeoneapi.HostConfig) error {
+	version124, _ := semver.NewConstraint("1.24.*")
+	version123, _ := semver.NewConstraint("1.23.*")
```
kron4eg, line 32: I'd move this to package level `var ()` block, with wrapping it in a `must(string) *semver.Constraints` that would panic on error. This way we'd always parse those strings and freakout on kubeone start (good indicator on CI).

kron4eg (commented):
looks good, but have a small nit.

pkg/tasks/fix_network_plugin.go
xmudrii: @kron4eg Fixed, PTAL.

PR #2029, opened 2022-05-10
Fix pre-pull images

kron4eg:
**What type of PR is this?**
/kind bug

Because we tried to pre-pull images before upgrading kubeadm it failed
during the upgrade process for kubernetes v1.21 (prior v1beta3 kubeadm).

```release-note
Fix pre-pull images
```


xmudrii (approved):
/approve

PR #2043, opened 2022-05-13
Switch to quay.io from docker.io for calico images

ahmedwaleedmalik:

**What type of PR is this?**

/kind feature

Switching to `quay.io` from `docker.io` for the following images:
- calico/cni
- calico/kube-controllers
- calico/node

Fixes #

```release-note
Switching to quay.io from docker.io for calico cni, kube-controllers, and node images.
```


kron4eg:
please also move to quay.io addons/calico-vxlan images, example:

https://github.com/kubermatic/kubeone/blob/f062a8cb68196a639bb5042f14b38b9fb64539df/addons/calico-vxlan/calico-vxlan.yaml#L4204

ahmedwaleedmalik:
> please also move to quay.io addons/calico-vxlan images, example:

Done but wouldn't it be better to declare these images https://github.com/kubermatic/kubeone/blob/master/pkg/templates/images/images.go#L168 here as well? Or just assume that the ones already declared should be consumed for calico-vxlan addon.

xmudrii:
@ahmedwaleedmalik We can do that, but if we would do that, I wouldn't use the same images/keys. Instead, I would add new keys for Calico VXLAN. There are two reasons for that:

* Someone might forget to update the Calico VXLAN manifest (e.g. not knowing about it), which could break something for Calico VXLAN users
* We might want to intentionally use different Calico versions for Canal and Calico VXLAN

ahmedwaleedmalik:
> @ahmedwaleedmalik We can do that, but if we would do that, I wouldn't use the same images/keys. Instead, I would add new keys for Calico VXLAN. There are two reasons for that:
> 
> * Someone might forget to update the Calico VXLAN manifest (e.g. not knowing about it), which could break something for Calico VXLAN users
> * We might want to intentionally use different Calico versions for Canal and Calico VXLAN

Very valid point, that makes sense. I have updated this accordingly PTAL.

xmudrii:
/lgtm
/assign @kron4eg 
for approval

Issue #2011, opened 2022-04-29
Update images to support Kubernetes 1.24


xmudrii:
Action items:

- [x] Update the issue template to add/remove [images](https://github.com/kubermatic/kubeone/blob/master/pkg/templates/images/images.go) as appropriate

The following components/images should be updated:

- [x] [AWS CCM](https://github.com/kubernetes/cloud-provider-aws) (#2055, #2171)
- [x] [AWS CSI](https://github.com/kubernetes-sigs/aws-ebs-csi-driver) (#2055, #2171)
- [x] [Azure CCM](https://github.com/kubernetes-sigs/cloud-provider-azure) (#2058, #2172)
- [x] [AzureDisk CSI](https://github.com/kubernetes-sigs/azuredisk-csi-driver) (#2058, #2172)
- [x] [AzureFile CSI](https://github.com/kubernetes-sigs/azurefile-csi-driver) (#2058, #2172)
- [x] [DigitalOcean CCM](https://github.com/digitalocean/digitalocean-cloud-controller-manager) (#2053)
- [x] [DigitalOcean CSI](https://github.com/digitalocean/csi-digitalocean) (#2173)
- [x] [EquinixMetal CCM](https://github.com/equinix/cloud-provider-equinix-metal) (#2054, #2174)
- [x] [GCP CSI](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver) (#2176)
- [x] [Hetzner CCM](https://github.com/hetznercloud/hcloud-cloud-controller-manager) (already the latest)
- [x] [Hetzner CSI](https://github.com/hetznercloud/csi-driver) (already the latest)
- [x] [Nutanix CSI](https://github.com/nutanix/helm) (#2182)
- [x] [OpenStack CCM](https://github.com/kubernetes/cloud-provider-openstack) (#2061, #2151)
- [x] [OpenStack CSI](https://github.com/kubernetes/cloud-provider-openstack) (#2061, #2151)
- [x] [vSphere CCM](https://github.com/kubernetes/cloud-provider-vsphere) (#2169)
- [x] [vSphere CSI](https://github.com/kubernetes-sigs/vsphere-csi-driver) (#2050, #2169)
- [x] [Cluster Autoscaler](https://github.com/kubernetes/autoscaler) (#2052, #2175)
- [x] [NodeLocalDNS Cache](https://kubernetes.io/docs/tasks/administer-cluster/nodelocaldns/) (#2079)
[...]

PR #2058, opened 2022-05-26
Update Azure CCM, AzureDisk CSI and AzureFile CSI

xmudrii:
**What type of PR is this?**

/kind feature

* Update Azure CCM to v1.24.0, v1.23.11, v1.1.14 (for Kubernetes 1.22), v1.0.18 (for Kubernetes 1.21), v0.7.21 (for Kubernetes 1.20)
* Update AzureDisk CSI driver to v1.18.0
* Update AzureFile CSI driver to v1.18.0

xref #2011 

```release-note
Update Azure CCM to v1.24.0, v1.23.11, v1.1.14 (for Kubernetes 1.22), v1.0.18 (for Kubernetes 1.21), v0.7.21 (for Kubernetes 1.20)
Update AzureDisk CSI driver to v1.18.0
Update AzureFile CSI driver to v1.18.0
```

/assign @kron4eg


addons/ccm-azure/ccm-azure.yaml
```diff
@@ -155,23 +154,20 @@ spec:
...
       hostNetwork: true
       serviceAccountName: cloud-controller-manager
+      nodeSelector:
+        node-role.kubernetes.io/control-plane: ""
       tolerations:
-        - key: node-role.kubernetes.io/control-plane
-          effect: NoSchedule
         - key: node-role.kubernetes.io/master
           effect: NoSchedule
-        - key: "node.cloudprovider.kubernetes.io/uninitialized"
```
kron4eg, line 163: I think we need this toleration, how else the CCM would run?

addons/ccm-azure/ccm-azure.yaml
```diff
@@ -281,41 +277,41 @@ spec:
     spec:
       priorityClassName: system-node-critical
       serviceAccountName: cloud-node-manager
-      hostNetwork: true # required to fetch correct hostname
+      hostNetwork: true   # required to fetch correct hostname
```
kron4eg, line 280: we probably should experiment with disabling this, as the node-manager get the $(NODE_NAME) as a flag and dien't need to fetch anything

addons/csi-azuredisk/csi-azuredisk-node.yaml
```diff
@@ -90,6 +90,8 @@ spec:
             - "--nodeid=$(KUBE_NODE_NAME)"
             - "--metrics-address=0.0.0.0:29605"
             - "--enable-perf-optimization=true"
+            - "--allow-empty-cloud-config=true"
```
kron4eg, line 93: in azure-disk-controller this flag is false, is this correct here (or there)?

kron4eg (commented):
some questions

addons/ccm-azure/ccm-azure.yaml
xmudrii: The upstream manifest doesn't have this annotation any longer and it works without it (I tested it manually), so I don't think it's needed at all.

addons/csi-azuredisk/csi-azuredisk-node.yaml
xmudrii: This is coming from the upstream manifest. I believe it's intentional. It might be very valid that you don't need cloud-config for the node-controller, but that you need it for the disk-controller.

addons/ccm-azure/ccm-azure.yaml
xmudrii: TBH, I would leave it as it is. Pretty much all other CCMs have hostNetwork enabled, and who knows is it only required for the hostname or something else as well. Even if we would try to disable it, I would leave it for a follow-up and merge this PR as it is (since this is not new and we were doing this since the beginning).

addons/ccm-azure/ccm-azure.yaml
kron4eg: Super strange... I suppose some other component initializes the node (node-agent maybe?).

PR #2130, opened 2022-07-04
Add new HostConfig.Labels API to manage custom labels on the static Nodes

### LINKED ISSUE #1822: Allow removal of the ServiceNodeExclusion label  from Master node
Currently this can be manually handle - if this label is removed manually from the node prior to the Nginx ingress controller installation, then only the control plane node is getting added to the load balancer. 

However, to support the KKP setup on single node with minimal manual interventions, it would be good to have this as part of kubeone. 

Possibility to exclude/remove label from the control plane node using Kubeone

Use case - In case of single master/worker node, the control plane is not getting added to the backend pool of Nginx Ingress Controller External Load balancer since node contains label  `node.kubernetes.io/exclude-from-external-load-balancers`

kron4eg:
**What type of PR is this?**
/kind feature
/kind api-change

To allow better management of the workloads distribution across different Nodes we need ability to add/remove custom labels on the control plane and static worker Nodes.

Fixes #1822

```release-note
Add new HostConfig.Labels API to manage custom labels on the static Nodes
```

### DIFF
--- a/pkg/apis/kubeone/types.go
+++ b/pkg/apis/kubeone/types.go
@@ -212,7 +212,7 @@ type HostConfig struct {
 	Taints []corev1.Taint `json:"taints,omitempty"`
 
 	// Lables to be used to apply (or remove, with minus sign suffix, see more kubectl help label) lables to/from node
-	Labels map[string]string `json:"labels"`
+	Labels map[string]string `json:"labels,omitempty"`
 
 	// Kubelet
 	Kubelet KubeletConfig `json:"kubelet,omitempty"`
--- a/pkg/apis/kubeone/types.go
+++ b/pkg/apis/kubeone/types.go
@@ -31,38 +31,55 @@ type KubeOneCluster struct {
 
 	// Name is the name of the cluster.
 	Name string `json:"name"`
+
 	// ControlPlane describes the control plane nodes and how to access them.
 	ControlPlane ControlPlaneConfig `json:"controlPlane"`
+
 	// APIEndpoint are pairs of address and port used to communicate with the Kubernetes API.
 	APIEndpoint APIEndpoint `json:"apiEndpoint"`
+
 	// CloudProvider configures the cloud provider specific features.
 	CloudProvider CloudProviderSpec `json:"cloudProvider"`
+
 	// Versions defines which Kubernetes version will be installed.
 	Versions VersionConfig `json:"versions"`
+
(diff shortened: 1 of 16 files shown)


pkg/apis/kubeone/types.go
```diff
@@ -142,46 +160,63 @@ const (
...
 	// Taints are taints applied to nodes. Those taints are only applied when the node is being provisioned.
 	// If not provided (i.e. nil) for control plane nodes, it defaults to:
 	//   * For Kubernetes 1.23 and older: TaintEffectNoSchedule with key node-role.kubernetes.io/master
 	//   * For Kubernetes 1.24 and newer: TaintEffectNoSchedule with keys
 	//     node-role.kubernetes.io/control-plane and node-role.kubernetes.io/master
 	// Explicitly empty (i.e. []corev1.Taint{}) means no taints will be applied (this is default for worker nodes).
 	Taints []corev1.Taint `json:"taints,omitempty"`
+
+	// Lables to be used to apply (or remove, with minus sign suffix, see more kubectl help label) lables to/from node
+	Labels map[string]string `json:"labels"`
```
xmudrii, line 215: Why don't we have omitempty here? We have it for other fields.

pkg/tasks/nodes.go
```diff
@@ -131,6 +132,21 @@ func labelNodeOSes(s *state.State) error {
 
 			node.Labels["v1.kubeone.io/operating-system"] = string(host.OperatingSystem)
 
+			for labKey, labVal := range host.Labels {
+				if strings.HasSuffix(labKey, "-") {
+					// additional `if` instead of `if strings.HasSuffix(labKey, "-") && labVal == ""` is used to avoid
+					// entering `else` section when the labVal is not empty.
+					if labVal == "" {
```
xmudrii, line 139: I see that if a label has `-` suffix and a value, we ignore it. I think we should have some API validation to reject such cases.

kron4eg: Good point, fixed

pkg/apis/kubeone/types.go
kron4eg: Hm, probably forgotten :shrug:

xmudrii (changes requested):
Can you also please update the example config?

xmudrii (approved):
/approve

PR #2163, opened 2022-07-19
testv2/upgrade: Open ZIP as RDWR and provide only one version to install

xmudrii:
**What type of PR is this?**

/kind failing-test

New upgrade tests are failing with:

1. Bad descriptor error because we open KubeOne ZIP archive as `WRONLY`, but we also try to read it
    ```
    helpers.go:168: opening zip file for reading: read /tmp/TestHetznerDefaultUpgradeContainerdFromV1_21_14_ToV1_22_111377836571/001/kubeone-1.4.4.zip: bad file descriptor
    ```
2. We provide two versions to the install scenario which causes a validation error
    ```
    scenario_install.go:66: only 1 version is expected to be set, got [v1.21.14 v1.22.11]
    ```

This PR fixes both issues.

```release-note
NONE
```

### DIFF
--- a/testv2/e2e/helpers.go
+++ b/testv2/e2e/helpers.go
@@ -147,7 +147,7 @@ func downloadKubeone(t *testing.T, version string) string {
 	}
 	defer resp.Body.Close()
 
-	zipBin, err := os.OpenFile(zipPath, os.O_CREATE|os.O_WRONLY, 0600)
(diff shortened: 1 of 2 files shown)


xmudrii:
/override pull-kubeone-e2e-hetzner-default-upgrade-containerd-from-v1.21.14-to-v1.22.11
We'll investigate later what's going on with this job.

ahmedwaleedmalik (approved):
/approve 🚀

PR #2165, opened 2022-07-20
Make CoreDNS deployment configurable, add Pod Anti Affinity and PodDisruptionBudget

xmudrii:
**What type of PR is this?**

/kind feature

This PR extends the KubeOneCluster API to make CoreDNS configurable. Two configuration options are added:

* Replicas -- number of CoreDNS replicas (default: 2)
* DeployPodDisruptionBudget -- whether should KubeOne create a PDB for CoreDNS (default: true)

Example KubeOneCluster manifest:

```yaml
kind: KubeOneCluster
versions:
  kubernetes: 1.23.9
cloudProvider:
  aws: {}
  external: false
features:
  coreDNS:
    replicas: 3
    deployPodDisruptionBudget: true
```

Additionally, this PR adds Pod Anti Affinity to the CoreDNS deployment to avoid having multiple CoreDNS pods on the same node.

### Implementation details

Kubeadm doesn't have any option to configure the CoreDNS deployment. Some upstream issues are requesting this as a feature, but all those issues are from a few years ago with no progress so far. Since we can't do this via Kubeadm, we let Kubeadm deploy CoreDNS as it is, then we manually patch it. This means that configured options will not be respected between running kubeadm and patching CoreDNS. This shouldn't be an issue because this happens just when creating or upgrading a cluster.

PodDisruptionBudget is deployed as a KubeOne Addon because it allows users to override the addon if they want to customize it.

```release-note
* Extend KubeOneCluster API with the `CoreDNS` feature allowing users to configure the number of CoreDNS replicas and whether should KubeOne create a PodDistruptionBudget for CoreDNS. Default values are 2 replicas and create PDB. Run `kubeone config print --full` for more details
* Add Pod Anti Affinity to the CoreDNS deployment to avoid having multiple CoreDNS pods on the same node
```


ahmedwaleedmalik (approved):
/approve ⚡

PR #2166, opened 2022-07-20
Update golangci-lint image to v1.47.1

xmudrii:
**What type of PR is this?**

/kind cleanup

This PR updates the `golangci-lint` image to v1.47.1. This release re-enables some of the linters disabled because of Go 1.18 in the previous `golangci-lint` release.

There are three new lint errors. Two errors are related to unused function arguments. The third error is related to the `kubeone proxy` Server, which is fixed by adding `ReadHeaderTimeout` of 1 minute to the `kubeone proxy` Server:

```
pkg/cmd/proxy.go:82:13: G112: Potential Slowloris Attack because ReadHeaderTimeout is not configured in the http.Server (gosec)
```

```release-note
NONE
```

### DIFF
--- a/pkg/cmd/proxy.go
+++ b/pkg/cmd/proxy.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"time"
 
 	"github.com/MakeNowJust/heredoc/v2"
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


pkg/cmd/proxy.go
```diff
@@ -97,6 +98,7 @@ func setupProxyTunnel(opts *proxyOpts) error {
 				http.Error(w, err.Error(), code)
 			}
 		}),
+		ReadHeaderTimeout: 1 * time.Minute,
```
xmudrii, line 101: While this fixes the linter failure, I'm not sure what's the "correct" value that we should use here. Different servers are using different values. For example, Nginx uses 60 seconds and that was the inspiration for choosing 1 minute.

ahmedwaleedmalik: 60 seconds sounds fine to me.

ahmedwaleedmalik (approved):
/approve

PR #2167, opened 2022-07-21
default-storage-class: Fix logic for determining if the CSI driver is deployed

xmudrii:
**What type of PR is this?**

/kind bug

We are determining if the CSI driver-based StorageClass should be deployed based on if the external CCM is enabled or if the Kubernetes version is >= 1.23. However, this is not always correct. In the case of vSphere, we require the external CCM to be enabled and we ignore the Kubernetes version.

This PR fixes this issue by using the `ensureCSIAddons` function to determine if the CSI driver will be deployed or not.

```release-note
Fix the logic for determining if the CSI driver is deployed in the default-storage-class addon. This fixes an issue with deploying the default-storage-class addon on vSphere clusters using the in-tree cloud provider
```

### DIFF
--- a/addons/default-storage-class/storage-class.yaml
+++ b/addons/default-storage-class/storage-class.yaml
@@ -1,8 +1,5 @@
-{{ $version := semver .Config.Versions.Kubernetes }}
-{{ $csi := or .Config.CloudProvider.External (ge $version.Minor 23) }}
-
 {{ if eq .Config.CloudProvider.CloudProviderName "azure" }}
-{{ if $csi }}
+{{ if .DeployCSIAddon }}
 apiVersion: storage.k8s.io/v1
 kind: StorageClass
 metadata:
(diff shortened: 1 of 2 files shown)


ahmedwaleedmalik (commented):
The PR in general is fine. But this statement `In the case of vSphere, we require both Kubernetes 1.23+ and the external CCM to be enabled.` is a bit contradictory. Based on [our code](https://github.com/kubermatic/kubeone/blob/62e2080041b88a32e8ae5be68613229d7c59bf3d/pkg/addons/ensure.go#L412), for vSphere, we only check if the external cloud provider is used. We don't check for k8s version. 

Are you sure no additional changes are required for vSphere?

xmudrii:
@ahmedwaleedmalik Good point, updated the PR description.

ahmedwaleedmalik (approved):
/approve ⚡

PR #2177, opened 2022-07-22
testv2: Port cloud provider tests

xmudrii:
**What type of PR is this?**

/kind feature

This PR ports [the cloud provider tests from the old tests](https://github.com/kubermatic/kubeone/blob/4fd7276f5623ef4ad42e7371f106eea478268d02/test/e2e/cloudprovider/cloudprovider.go) to the new tests.

Those tests are needed to properly test CCM and CSI drivers.

```release-note
NONE
```

### DIFF
--- a/testv2/e2e/cloudprovider.go
+++ b/testv2/e2e/cloudprovider.go
@@ -1,0 +1,381 @@
+/*
+Copyright 2022 The KubeOne Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import (
+	"context"
+	"net/http"
+	"strings"
+	"testing"
+	"time"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func supportsCloudProviderTests(provider string) bool {
+	switch provider {
+	case "aws", "digitalocean", "hetzner", "gce", "openstack":
+		return true
+	default:
+		return false
+	}
+}
+
+const (
+	cpTestNamespaceName   = "test-cloud-provider"
+	cpTestStatefulSetName = "test-sts"
+	cpTestServiceName     = "test-svc"
+
+	cpTestPollPeriod = 10 * time.Second
+	cpTestTimeout    = 10 * time.Minute
+)
+
+var (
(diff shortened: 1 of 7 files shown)


testv2/e2e/cloudprovider.go
```diff
@@ -0,0 +1,342 @@
...
+
+func (c *cloudProviderTests) run(t *testing.T) error {
+	if err := c.createPVCWithStorage(t); err != nil {
+		return err
+	}
+	if err := c.exposePVC(t); err != nil {
+		return err
+	}
+
+	return c.cleanUp(t)
```
kron4eg, line 88: cleanUp should be a defer function, or even better a [testing.T.Cleanup](https://pkg.go.dev/testing#T.Cleanup), and it should be registered before any cloud-resource-producing functions are run.

xmudrii: This should be fixed now.

testv2/e2e/scenario_migrate_csi_ccm.go
```diff
@@ -73,33 +73,27 @@ func (scenario *scenarioMigrateCSIAndCCM) Run(t *testing.T) {
 	k1Old := install.kubeone(t)
 	waitKubeOneNodesReady(t, k1Old)
 
+	client := dynamicClientRetriable(t, k1Old)
+	cpTests := newCloudProviderTests(client, scenario.infra.Provider())
+	defer cpTests.cleanUp(t)
```
kron4eg, line 78: cpTests.runWithCleanup()

xmudrii: This is intentional. I want to ensure that once the CCM/CSI migration is done, the pod with PVC is Running and LB is accessible. For that, I intentionally defer cleanup until the end.

testv2/e2e/cloudprovider.go
```diff
@@ -0,0 +1,381 @@
...
+
+		return
+	}
+
+	c.createStatefulSetWithStorage(t)
+	c.exposeStatefulSet(t)
+}
+
+func (c *cloudProviderTests) runWithCleanup(t *testing.T) {
+	defer c.cleanUp(t)
```
kron4eg, line 84: This can be t.Cleanup(), it defer the cleanup cross the calls.

testv2/e2e/cloudprovider.go
```diff
@@ -0,0 +1,381 @@
...
+		t.Logf("Skipping cloud provider tests because cloud provider %q is not supported.", c.provider)
+
+		return
+	}
+
+	c.createStatefulSetWithStorage(t)
+	c.exposeStatefulSet(t)
+}
+
+func (c *cloudProviderTests) runWithCleanup(t *testing.T) {
```
kron4eg, line 83: Why not cleanUp always inside run()? If we didn't created anything, cleanUp() would simply return after few 404 GETs.

xmudrii: Please see https://github.com/kubermatic/kubeone/pull/2177#discussion_r940343385

testv2/e2e/scenario_migrate_csi_ccm.go
kron4eg: You could `t.Cleanup()` in that case. And `cpTests.run()` can do it for you ;)

kron4eg: `t.Cleanup()` defers until the very end of the whole test!

xmudrii: I want it to work in the following way:

* non-CCM/CSI migration tests: cleanup as soon as possible, i.e. before starting Sonobuoy tests (that's why `defer` is useful here)
* CCM/CSI migration tests: cleanup at the end, after making sure everything is running after the migration (note that we also don't have Sonobuoy tests for CCM/CSI migration, which is completely okay)

The first case is to save resources/money and leave space for other tests to run. Sonobuoy tests can take 10-20 minutes (or even longer if you run a bigger suite), so by cleaning up earlier, we let more tests run in parallel.

PR #2182, opened 2022-07-22
Update the Nutanix CSI driver to v2.5.1

xmudrii:
**What type of PR is this?**

/kind documentation

Update the Nutanix CSI driver to v2.5.1.

The [upstream chart](https://github.com/nutanix/helm) structure has been changed, so the addon structure has been changed as a result of that. Some files are renamed, some components are using Deployments instead of StatefulSets, there are new ServiceAccounts, and more.

I already manually tested this PR and it seems to work as expected. This includes running this PR on a cluster created with KubeOne v1.4.5.

xref #2011 

```release-note
Update the Nutanix CSI driver to v2.5.1
```


ahmedwaleedmalik (approved):
/approve ⚡

PR #2188, opened 2022-07-25
Update Canal and Calico VXLAN to v3.23.3

### LINKED ISSUE #2118: Update Calico/Canal to v3.23

We should update Calico/Canal to the latest minor release (v3.22) in order to get the latest features and bug fixes.

### Solution details

- Update Calico in the Canal addon to v3.23
- Update Calico VXLAN addon to v3.23

xmudrii:
**What type of PR is this?**

/kind feature

This PR updates Canal and Calico VXLAN to the latest available version (v3.23.3).

Fixes #2118 

```release-note
Update Canal and Calico VXLAN to v3.23.3
```

/assign @ahmedwaleedmalik


addons/cni-canal/canal.yaml
```diff
@@ -4341,7 +4483,7 @@ spec:
             privileged: true
           resources:
             requests:
-              cpu: {{ default .Params.RequestsCPU "250m" }}
+              cpu: 250m
```
ahmedwaleedmalik, line 4486: Please revert this change. 
```suggestion
             cpu: {{ default .Params.RequestsCPU "250m" }}
```

addons/calico-vxlan/calico-vxlan.yaml
```diff
@@ -1263,12 +1364,22 @@ spec:
...
               vxlanEnabled:
+                description: 'VXLANEnabled overrides whether Felix should create the
+                  VXLAN tunnel device for VXLAN networking. Optional as Felix determines
+                  this based on the existing IP pools. [Default: nil (unset)]'
                 type: boolean
               vxlanMTU:
-                description: 'VXLANMTU is the MTU to set on the tunnel device. See
-                  Configuring MTU [Default: 1440]'
+                description: 'VXLANMTU is the MTU to set on the IPv4 VXLAN tunnel
+                  device. See Configuring MTU [Default: 1410]'
```
xmudrii, line 1378: It appears that the default MTU has been changed in Calico 3.23. Do we need to take any action on this?

addons/calico-vxlan/calico-vxlan.yaml
```diff
@@ -4316,6 +4446,9 @@ spec:
             # Enable or Disable VXLAN on the default IP pool.
             - name: CALICO_IPV4POOL_VXLAN
               value: "CrossSubnet"
+            # Enable or Disable VXLAN on the default IPv6 IP pool.
+            - name: CALICO_IPV6POOL_VXLAN
+              value: "CrossSubnet"
```
xmudrii, line 4451: Since we don't support dual-stack or IPv6, should we disable this option?

addons/cni-canal/canal.yaml
```diff
@@ -1280,12 +1381,22 @@ spec:
...
               vxlanEnabled:
+                description: 'VXLANEnabled overrides whether Felix should create the
+                  VXLAN tunnel device for VXLAN networking. Optional as Felix determines
+                  this based on the existing IP pools. [Default: nil (unset)]'
                 type: boolean
               vxlanMTU:
-                description: 'VXLANMTU is the MTU to set on the tunnel device. See
-                  Configuring MTU [Default: 1440]'
+                description: 'VXLANMTU is the MTU to set on the IPv4 VXLAN tunnel
+                  device. See Configuring MTU [Default: 1410]'
```
xmudrii, line 1395: Same here -- do we need to take any action on this MTU change?

addons/cni-canal/canal.yaml
xmudrii: Fixed.

addons/calico-vxlan/calico-vxlan.yaml
ahmedwaleedmalik: I don't think so.

kron4eg: I wonder why they did it

kron4eg: It's about ipv6 support: https://github.com/projectcalico/calico/commit/009498285c7f6634ab0fb3db7b38c4567f4cbe34

So probably we can do the same

addons/cni-canal/canal.yaml
rastislavs: I think for Canal this effectively isn't important at all - not used, as Calico is used only to provide network policies support, VXLANs are done by Flannel

addons/calico-vxlan/calico-vxlan.yaml
rastislavs: I think that in case that IPv6 is not enabled (there are no IPv6 pools) this should not be important (should have no effect). Since `CALICO_IPV4POOL_VXLAN` is set to `CrossSubnet` as well, I think it is dafe to just keep it like it is.

addons/calico-vxlan/calico-vxlan.yaml
rastislavs: To me this actually isn't a change for us - it is a change in the `FelixConfiguration` CRD, that we are not using, as we do not deploy Calico via operator. In this manifest-based deployment of Calico, there is no use of `FelixConfiguration` at all, and the MTU is passed from the `veth_mtu` config in `calico-config` configmap. By default is set to `0` - which means auto-configuration - and that should work for most of the deployments.

xmudrii:
/override pull-kubeone-e2e-aws-defaults-calico-containerd-v1.22.11
It seems like this is not caused by this update, so I created #2192 to track and fix this issue.

PR #2214, opened 2022-08-01
Rollout pods if credentials are changed

### LINKED ISSUE #2180: machine-controller pods should restart on rotation of credentials
Currently, for few cloud providers, like Openstack and Azure, the credentials for service account expire after certain number of days. It is not possible to get un-expiring passwords for service accounts for some of these cloud providers. So we must rotate the credentials by running `kubeone apply --force-upgrade`

But currently, post rotation, machine-controller pods (controller as well as webhook) do not get restarted and they continue to use old credentials.

So.. we can do one of the below options:

1. Make changes in kubeone logic to restart machine-controller in such cases.
2. Add annotations for [stakater reloader](https://github.com/stakater/Reloader) for on machine-controller and webhook pods so we can add stakater reloader as kubeone addon and it will take care of restarting machine-controller pods when the secrets get updated with new credentials.

xmudrii:
**What type of PR is this?**

/kind feature

This PR implements rolling out pods that are using `kubeone-*-credentials` Secrets if credentials are changed.

The implementation is based on applying annotation to all pods that are using `kubeone-*-credentials` Secrets. The annotation contains salt and credentials hashed using the SHA256 algorithm.

Fixes #2180 

```release-note
Rollout pods that are using `kubeone-*-credentials` Secrets if credentials are changed
```

### DIFF
--- a/addons/ccm-digitalocean/ccm-digitalocean.yaml
+++ b/addons/ccm-digitalocean/ccm-digitalocean.yaml
@@ -14,6 +14,8 @@ spec:
     metadata:
       labels:
         app: digitalocean-cloud-controller-manager
+      annotations:
+        "credentials-hash": "{{ .CredentialsCCMHash }}"
     spec:
       priorityClassName: system-cluster-critical
       dnsPolicy: Default
--- a/addons/ccm-digitalocean/ccm-digitalocean.yaml
+++ b/addons/ccm-digitalocean/ccm-digitalocean.yaml
@@ -15,7 +15,7 @@ spec:
       labels:
         app: digitalocean-cloud-controller-manager
       annotations:
-        "credentials-hash": "{{ .CredentialsCCMHash }}"
+        "kubeone.k8c.io/credentials-hash": "{{ .CredentialsCCMHash }}"
     spec:
       priorityClassName: system-cluster-critical
       dnsPolicy: Default
--- a/pkg/addons/applier.go
+++ b/pkg/addons/applier.go
@@ -18,7 +18,9 @@ package addons
 
 import (
 	"crypto/rsa"
+	"crypto/sha256"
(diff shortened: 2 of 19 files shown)


pkg/addons/applier.go
```diff
@@ -341,6 +365,29 @@ func mcCredentialsEnvVars(s *state.State) ([]byte, error) {
...
+	}
+
+	omap := orderedmap.New()
+	for k, v := range creds {
+		omap.Set(k, v)
+	}
+	omap.SortKeys(sort.Strings)
+
+	hash := fmt.Sprintf("kubeone-%s", s.Cluster.CloudProvider.CloudProviderName())
+	for k, v := range creds {
```
kron4eg, line 381: should be iteration over the omap, not creds

addons/ccm-digitalocean/ccm-digitalocean.yaml
```diff
@@ -14,6 +14,8 @@ spec:
     metadata:
       labels:
         app: digitalocean-cloud-controller-manager
+      annotations:
+        "credentials-hash": "{{ .CredentialsCCMHash }}"
```
ahmedwaleedmalik, line 18: Should we add some prefixes to this annotation like "kubeone.k8c.io/"? I think making it explicit that this annotation is managed by KubeOne can be valuable for the end user.

pkg/addons/applier.go
xmudrii: Fixed.

addons/ccm-digitalocean/ccm-digitalocean.yaml
xmudrii: I added the `kubeone.k8c.io/` prefix for all KubeOne annotations.

PR #2216, opened 2022-08-02
[release/v1.4] Rollout pods if credentials are changed

### LINKED ISSUE #2180: machine-controller pods should restart on rotation of credentials

xmudrii:
**What type of PR is this?**

/kind bug

This PR is a manual cherry-pick of #2214.

Fixes #2180 

```release-note
Rollout pods that are using `kubeone-*-credentials` Secrets if credentials are changed
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -97,4 +97,4 @@ issues:
     text: "cyclomatic complexity 36 of func `openstackValidationFunc` is high"
 
   - path: pkg/addons
-    text: "cyclomatic complexity 36 of func `newAddonsApplier` is high"
+    text: "cyclomatic complexity 39 of func `newAddonsApplier` is high"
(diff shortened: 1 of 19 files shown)


ahmedwaleedmalik (approved):
/approve

PR #2220, opened 2022-08-03
Update Cilium to v1.12.0

### LINKED ISSUE #2219: Bump cilium to version 1.12
### Our story

We need to use cilium on Hetzner Cloud, because traffic flowing from lb (internal ips) => kube-proxy (snat*) => svc/pod on the same node will always originate from the public eth0 ip address of the node, thus breaking a lot of things for us (i.e. services always need to know all public ips in the cluster to exclude them from x-forwarded-for/proxy protocol). We invested a lot of time and tried many tweaks to convince canal/flannel to use the internal ip, but without luck.

With a switch to cilium v1.11.x we essentially fixed this entirely in a snap using the strict kube-proxy replacement - cilium agent will use the private ip out of the box instead of the public one to talk to pods on the same host. We also benefit from the other cilium features (hubble insights, bpf performance ...) and are therefore very happy to find kubeone supporting cilium. However, afair kubeone 1.4.5 installs cilium v1.11.1, which is quite old and showed some bugs in our cluster, which have been resolved in the last months and are gone in cilium v1.12 (stable).

For testing purposes I updated cilium manually with your given helm command and everything seems to be working very well since over a week. I couldn't make out any breaking changes in the release notes. I'm aware cilium v1.12 is really new as it has been released just about 2 weeks ago and understand absolutely it probably needs some testing.

What would be needed to have you bump this version of cilium into kubeone? Alternatively, would it be possible to just be able to configure the desired cilium version in kubeone.yml?

### Solution details

- bump cilium to stable 1.12

### Alternative approaches

- allow to configure cilium version in kubeone.yml

### Use cases

- additional features like native service mesh and ingress support
- mainly bug fixes

xmudrii:
**What type of PR is this?**

/kind feature

This PR updates Cilium to v1.12.0.

Fixes #2219 

```release-note
Update Cilium to v1.12.0
```


addons/cni-cilium/cilium.yaml
```diff
@@ -119,14 +142,14 @@ data:
   cluster-name: default
```
xmudrii, line 142: Do we want to put some name here, e.g. name of the cluster?

addons/cni-cilium/cilium.yaml
```diff
@@ -119,14 +142,14 @@ data:
...
   # Unique ID of the cluster. Must be unique across all conneted clusters and
   # in the range of 1 and 255. Only relevant when building a mesh of clusters.
-  cluster-id: ""
+  cluster-id: "0"
```
xmudrii, line 145: We are using 0 in KKP, but the comment says that it should be in the range of 1 and 255. Should we change this?

https://github.com/kubermatic/kubermatic/blob/5ea2f529409b18b10a8c74b533265a5182aef48e/addons/cilium/cilium_v1.12.yaml#L150

addons/cni-cilium/cilium.yaml
```diff
@@ -138,7 +161,6 @@ data:
   install-no-conntrack-iptables-rules: "false"
 
   auto-direct-node-routes: "false"
-  enable-bandwidth-manager: "false"
```
xmudrii, line 141: Do we need this? It's removed from the upstream manifest.

addons/cni-cilium/cilium.yaml
```diff
@@ -176,13 +200,59 @@ data:
...
   set-cilium-is-up-condition: "true"
   unmanaged-pod-watcher-interval: "15"
+  tofqdns-dns-reject-response-code: "refused"
+  tofqdns-enable-dns-compression: "true"
+  tofqdns-endpoint-max-ip-per-hostname: "50"
+  tofqdns-idle-connection-grace-period: "0s"
+  tofqdns-max-deferred-connection-deletes: "10000"
+  tofqdns-min-ttl: "3600"
+  tofqdns-proxy-response-max-delay: "100ms"
+  agent-not-ready-taint-key: "node.cilium.io/agent-not-ready"
```
xmudrii, line 223: There is a bunch of added and removed options here and above that I'm unsure about. Should we leave everything as it is?

addons/cni-cilium/cilium.yaml
```diff
@@ -176,13 +200,59 @@ data:
   cluster-pool-ipv4-cidr: "{{.Config.ClusterNetwork.PodSubnet }}"
   cluster-pool-ipv4-mask-size: "24"
```
xmudrii, line 201: TODO: This is maybe not correct.

addons/cni-cilium/cilium.yaml
```diff
@@ -176,13 +200,59 @@ data:
...
+{{ if .Config.ClusterNetwork.CNI.Cilium.EnableHubble }}
+# Source: cilium/templates/hubble-relay/configmap.yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: hubble-relay-config
+  namespace: kube-system
+data:
+  config.yaml: |
+    cluster-name: default
```
xmudrii, line 234: Same as above, do we want to put some name here, e.g. name of the cluster?

addons/cni-cilium/cilium.yaml
```diff
@@ -176,13 +200,59 @@ data:
...
+    peer-service: "hubble-peer.kube-system.svc.cluster.local:443"
```
xmudrii, line 235: This was originally using a socket instead of a service. Is this change okay or should we revert to using socket?

addons/cni-cilium/cilium.yaml
WeirdMachine: We can do this but its not needed. IIRC this is only needed for cluster mesh deployment.

addons/cni-cilium/cilium.yaml
WeirdMachine: I think any value is fine here as long we do not use cluster mesh.
Maybe 0 is actually used to indicate its not a cluster mesh

addons/cni-cilium/cilium.yaml
WeirdMachine: Can be removed.

addons/cni-cilium/cilium.yaml
rastislavs: it is working fine in case of KKP (even upgrade from the socket version), so I would use the new service version

addons/cni-cilium/cilium.yaml
rastislavs: used only for clustermesh - not needed for single-cluster deployment

addons/cni-cilium/cilium.yaml
rastislavs: I would leave them as they are.

addons/cni-cilium/cilium.yaml
rastislavs: `{{.Config.ClusterNetwork.PodSubnet }}` looks good to me here.
Going forward it may be helpful to also make mask-size configurable as well, but if it was not demanded up til now, the default be fine.

xmudrii:
Some conformance tests are failing, but that's as expected (see https://github.com/cilium/cilium/issues/14287 and https://github.com/cilium/cilium/issues/12116).

```
    helpers.go:523: some e2e tests failed:
        [
          {
            "name": "[sig-network] HostPort validates that there is no conflict between pods with same hostPort but different hostIP and protocol [LinuxOnly] [Conformance]",
            "status": "failed",
            "details": {}
          },
          {
            "name": "[sig-network] Services should have session affinity work for service with type clusterIP [LinuxOnly] [Conformance]",
            "status": "failed",
            "details": {}
          }
        ]
```

/override pull-kubeone-e2e-aws-defaults-cilium-containerd-v1.22.11

PR #2236, opened 2022-08-05
Increase the minimum Kubernetes version to v1.22.0

xmudrii:
**What type of PR is this?**

/kind api-change

This PR proposes increasing the minimum Kubernetes version to v1.22.0 (from v1.20.0). Kubernetes v1.21 reached EOL on June 28th and v1.20 reached EOL on February 28th. Removing support for those EOL releases will make maintaining KubeOne 1.5 release easier because we don't have to worry about missing or deprecated/removed features/APIs.

Additionally, images used for Kubernetes 1.21 and 1.20 clusters are removed because those are not used any longer (in this case).

```release-note
[ACTION REQUIRED] The minimum Kubernetes version has been increased to v1.22.0. If you're still using Kubernetes v1.21 or v1.20, you have to upgrade the cluster to v1.22 or newer **before** upgrading to KubeOne 1.5.
```


kron4eg:
/approve
/lgtm

🦣

PR #2258, opened 2022-08-08
Mark `install` and `upgrade` subcommands deprecated

kron4eg:
**What type of PR is this?**
/kind cleanup
/kind deprecation

We had `kubeone apply` since long time, and nowadays even started to use it in e2e tests.

Fixes #

```release-note
[ACTION REQUIRED] We announced with the KubeOne 1.4.0 release that kubeone install and kubeone upgrade commands are deprecated in favor of kubeone apply. This time we're marking those commands as hidden, so they'll not show in the help output. In the next release, we'll completely remove those commands, so we strongly recommend migrating to kubeone apply as soon as possible.
```

### DIFF
--- a/pkg/cmd/install.go
+++ b/pkg/cmd/install.go
@@ -52,14 +52,17 @@ func installCmd(rootFlags *pflag.FlagSet) *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "install",
-		Short: "Install Kubernetes",
+		Short: "[DEPRECATED] Install Kubernetes",
 		Long: heredoc.Doc(`
+			[DEPRECATED] This command is deprecated, please use kubeone apply instead.
+
 			Install Kubernetes on pre-existing machines
 
 			This command takes KubeOne manifest which contains information about hosts and how the cluster should be provisioned.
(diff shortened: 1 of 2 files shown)


xmudrii:
@kron4eg Let's make this an action required release note. WDYT about something like this:

> [ACTION REQUIRED] We announced with the KubeOne 1.4.0 release that `kubeone install` and `kubeone upgrade` commands are deprecated in favor of `kubeone apply`. This time we're marking those commands as hidden, so they'll not show in the help output. In the next release, we'll completely remove those commands, so we strongly recommend migrating to `kubeone apply` as soon as possible.

kron4eg:
@xmudrii sounds good, updated!

pkg/cmd/install.go
```diff
@@ -54,12 +54,15 @@ func installCmd(rootFlags *pflag.FlagSet) *cobra.Command {
 		Use:   "install",
 		Short: "Install Kubernetes",
```
xmudrii, line 55: Nit: should we add `[DEPRECATED]` here as well? We have it for `upgrade`.

kron4eg: Changes

xmudrii (approved):
/lgtm
/approve

PR #2263, opened 2022-08-09
testv2: Fix mode name for Sonobuoy conformance tests

xmudrii:

We're using a mode called `conformance` for conformance tests, however, the such mode doesn't exist. The following modes exist:

```
  -m, --mode Mode                                What mode to run the e2e plugin in. Valid modes are [certified-conformance conformance-lite non-disruptive-conformance quick]. (default non-disruptive-conformance)
```

`non-disruptive-conformance` is the mode that we should be using instead:

```
Mode: non-disruptive-conformance
Description: Non-destructive conformance mode runs all of the conformance tests except those that would disrupt other cluster operations (e.g. tests that may cause nodes to be restarted or impact clus... (truncated) ...
E2E_FOCUS: \[Conformance\]
E2E_SKIP: \[Disruptive\]|NoExecuteTaintManager
E2E_PARALLEL: false
```

**What type of PR is this?**

/kind bug

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/testv2/e2e/sonobuoy.go
+++ b/testv2/e2e/sonobuoy.go
@@ -44,7 +44,7 @@ type sonobuoyReportDetails struct {
 type sonobuoyMode string
 
 const (
-	sonobuoyConformance     sonobuoyMode = "conformance"
+	sonobuoyConformance     sonobuoyMode = "non-disruptive-conformance"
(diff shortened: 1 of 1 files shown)


xmudrii:
/hold
To get #2264 merged first.

PR #2287, opened 2022-08-17
VCD: Fix for worker machine disk size

ahmedwaleedmalik:

This fixes a bug where we were erroneously assuming that disk size will be specified in MBs for the worker machines. Although this is not the case and it should be specified in GBs instead.

Fixes #

**What type of PR is this?**

/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


xmudrii (approved):
/approve

PR #2290, opened 2022-08-18
Add safeguards for master node-role and cloud-init provisioningUtility

xmudrii:

This PR introduces two new safeguards to the KubeOne reconciliation process:

- Disallow upgrading to Kubernetes 1.24 if there are pods with `node-role.kubernetes.io/master` NodeSelector or toleration. `node-role.kubernetes.io/master` node role is removed in Kubernetes 1.24. If there are pods using it as a NodeSelector, those pods will not be able to get scheduled after upgrading to 1.24. Similarly, if there are pods that tolerate the master node role, but not `node-role.kubernetes.io/control-plane`, those pods might get stuck as well. This can cause the eviction to fail (because pods can't get scheduled), which causes the upgrade process to fail when upgrading from 1.23 to 1.24.
- Warn users if there are Flatcar-based MachineDeployments that use cloud-init provisioningUtility. The cloud-init provisioningUtility is not supported on Flatcar with OSM. Machines using cloud-init will get stuck and never join the cluster. This is not a binding safeguard, i.e. we only warn users about this, we don't block the reconciliation process (e.g. upgrading the cluster or just running `kubeone apply`).

**What type of PR is this?**

/kind feature

```release-note
Introduce additional safeguards in the KubeOne reconciliation process to disallow upgrading to Kubernetes 1.24 if there are pods that use removed master node-role (`node-role.kubernetes.io/master`), and if there are Flatcar-based MachineDeployments that use the cloud-init provisioningUtility in a cluster with Operating System Manager (OSM) enabled.
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/pkg/tasks/probes.go
+++ b/pkg/tasks/probes.go
@@ -17,11 +17,15 @@ limitations under the License.
 package tasks
 
 import (
+	"encoding/json"
 	"fmt"
 	"regexp"
 	"strings"
 
 	"github.com/Masterminds/semver/v3"
+	clusterv1alpha1 "github.com/kubermatic/machine-controller/pkg/apis/cluster/v1alpha1"
+	"github.com/kubermatic/machine-controller/pkg/jsonutil"
+	providerconfigtypes "github.com/kubermatic/machine-controller/pkg/providerconfig/types"
 	"github.com/pkg/errors"
 	"gopkg.in/yaml.v2"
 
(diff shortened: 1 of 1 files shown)


pkg/tasks/probes.go
```diff
@@ -134,6 +147,139 @@ func safeguard(s *state.State) error {
...
+				return fail.RuntimeError{
+					Err: err,
+					Op:  "decoding operatingsystemconfig for machinedeployment",
+				}
+			}
+
+			if v, ok := osConfig[provisioningUtilityKey]; ok && v == provisioningUtilityCloudInit {
+				if aVal := md.Annotations[machineDeploymentOSMAnnotation]; aVal == "" || aVal == ospFlatcar {
+					cloudInitMDs = append(cloudInitMDs, md.Name)
+				}
```
ahmedwaleedmalik, line 270: We should omit this check. Since for existing machine deployments, the OSP annotation wouldn't exist anyways.

ahmedwaleedmalik: My bad, I didn't properly go through the condition before commenting. This looks good, as expected ;)

ahmedwaleedmalik (approved):
/approve

PR #2294, opened 2022-08-18
testv2: Clean PVCs when cleaning up the cloud provider tests

xmudrii:

Removing StatefulSet doesn't remove PVC, so we're now doing this explicitly.

**What type of PR is this?**

/kind bug

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```


xrstf (approved):
/approve

PR #2295, opened 2022-08-18
[Test] Run cluster upgrading tests on DigitalOcean

xmudrii:
Relevant to #2250.


ahmedwaleedmalik:
The upgrade tests failed because sonobuoy flaked out when writing the results to disk. However, all required tests (including Sonobuoy tests) passed, so I'm considering those tests as passed.

PR #2298, opened 2022-08-19
Add Plan block for the control plane VMs to the example Terraform configs for Azure

xmudrii:

## Problem Overview

If a VM is using an image that comes from the Azure Marketplace, the VM must have a Plan block defined and populated, otherwise, provisioning fails with the following error:

```
│ Error: Code="VMMarketplaceInvalidInput" Message="Creating a virtual machine from Marketplace image or a custom image sourced from a Marketplace image requires Plan information in the request. VM: '<redacted>'."
```

This wasn't a problem before because we only supported Ubuntu until #2266, and Ubuntu doesn't require Plan to be specified.

## Problem Statement

Depending on where the image is coming from, the Plan block:

* **MUST NOT** be **preset** if the image is a built-in Azure image (i.e. not coming from the marketplace)
  * You can't provide null or empty strings, the block must be omitted entirely
  * Ubuntu and CentOS are those images in our case
* **MUST** be **present** if the image is coming from the marketplace
  * RHEL, Rocky Linux, and Flatcar are images coming from the marketplace

We need to make Terraform to omit the Plan block if OS is Ubuntu or CentOS, and populate the block if it's an image coming from the marketplace.

## Solution

This is not easy out of the box with Terraform. I decided to utilize [Dynamic Blocks](https://www.terraform.io/language/expressions/dynamic-blocks) since this seems to be the only way you can dynamically set or remove a block.

To make Dynamic Blocks work, the object must be iterable, i.e. you can't use a simple object, you have to use a list or a map.

The `image_references` variable has been changed:

* Information about the image is moved to a new nested object (`image`)
* A new nested object (`plan`) type of list is added with information about the Plan
  * If a Plan is not supposed to be used, the list is empty
  * If a Plan is supposed to be used, the list must contain **ONLY ONE** Plan to be used
    * We have no validation for this, but I don't think anyone should provide more than one Plan

This approach seems to integrate very well with the dynamic Plan block.

**What type of PR is this?**

/kind bug

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #2273, opened 2022-08-11
[TESTING] Openstack testing PR

kron4eg:
xref #2242

PR #2299, opened 2022-08-22
Mount /etc/pki to the OpenStack CCM container and minor test fixes

xmudrii:

CA certificates in the `/etc/ssl/certs` directory on CentOS 7 are only symlinks pointing to `/etc/pki`. We only mount `/etc/ssl/certs` in the OpenStack CCM container, so symlinks are not working at all. This is causing the OpenStack CCM to end up in a CrashLoopBack. To fix this issue, we are now mounting `/etc/pki` to the OpenStack CCM container.

Addionally, this PR includes some fixes from #2273 needed to make the new E2E tests work on OpenStack.

**What type of PR is this?**

/kind bug

```release-note
- Mount `/etc/pki` to the OpenStack CCM container to fix CrashLoopBackoff on clusters running CentOS 7 
- Example Terraform configs for OpenStack are no longer attaching a Floating IP address to the initial MachineDeployment. This matches the behavior of not attaching Floating IP addresses to the control plane nodes.
```

**Documentation**:
```documentation
NONE
```


examples/terraform/openstack/output.tf
```diff
@@ -86,7 +86,7 @@ output "kubeone_workers" {
           securityGroups = [openstack_networking_secgroup_v2.securitygroup.name]
           network        = openstack_networking_network_v2.network.name
           subnet         = openstack_networking_subnet_v2.subnet.name
-          floatingIpPool = var.external_network_name
+          # floatingIpPool = var.external_network_name
```
ahmedwaleedmalik, line 89: We should probably revert this change?!

ahmedwaleedmalik (approved):
/approve

examples/terraform/openstack/output.tf
xmudrii: We decided to keep this change. The release note is updated accordingly.

xmudrii:
/override pull-kubeone-e2e-openstack-default-install-containerd-external-v1.24.3
We'll test this as part of #2273

PR #2302, opened 2022-08-22
Fix Flatcar: mkdir /opt/bin before untar anything in it

xmudrii:

We found an issue with Flatcar where we're trying to untar archives to `/opt/bin` before actually creating that directory. If that directory doesn't exist beforehand, the provisioning process will fail:

```
tar: /opt/bin: Cannot open: No such file or directory
```

This has been discovered while testing the Flatcar support on OpenStack in #2273

**What type of PR is this?**

/kind bug

```release-note
Explicitly create `/opt/bin` on Flatcar before trying to untar anything to that directory
```

**Documentation**:
```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #2303, opened 2022-08-23
[release/v1.4] Mount /etc/pki to the OpenStack CCM container

xmudrii:
**What this PR does / why we need it**:

This is a partial cherry-pick of #2299 to the `release/v1.4` branch to fix an issue with the OpenStack CCM crashing on clusters running CentOS 7.

**What type of PR is this?**

/kind bug

**Does this PR introduce a user-facing change? Then add your Release Note here**:
```release-note
Mount `/etc/pki` to the OpenStack CCM container to fix CrashLoopBackoff on clusters running CentOS 7 
```

**Documentation**:
```documentation
NONE
```

PR #2304, opened 2022-08-23
Build stable KubeOne binary instead of downloading it from GitHub

xmudrii:

Right now, we're downloading the latest stable KubeOne binary from GitHub and we use it to provision the cluster in the upgrade tests. After that, we build the binary based on PR's ref and use it to upgrade the cluster.

The issue with downloading the KubeOne binary from GitHub is that it might miss some changes required for tests to pass. For example, we had a problem with OpenStack CentOS upgrade tests failing because of missing volume mounts. We cherry-picked the fix in #2303, but we would have to release a new KubeOne patch release to be able to use that fix in tests. This is not optimal as we would have to cut patch releases way too often.

To mitigate this, this PR proposes building the stable binary directory from the release branch (e.g. `release/v1.4`). In this case, we have all the latest fixes without needing to cut a patch release.

The old logic is still in code, so we can easily switch back to download the latest stable binary from GitHub.

**What type of PR is this?**

/kind cleanup

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/testv2/e2e/helpers.go
+++ b/testv2/e2e/helpers.go
@@ -117,15 +117,30 @@ func requiredTemplateFunc(warn string, input interface{}) (interface{}, error) {
 }
 
 func makeBin(args ...string) *testutil.Exec {
+	return makeBinWithPath(filepath.Clean("../../"), args...)
+}
+
+func makeBinWithPath(path string, args ...string) *testutil.Exec {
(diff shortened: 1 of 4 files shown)


ahmedwaleedmalik (approved):
/approve ⚡ 

This approach makes sense to me. 👍

PR #2292, opened 2022-08-18
Move vmware csi driver to vmware-system-csi namespace

WeirdMachine:
**Which issue(s) this PR fixes**:

Fixes https://github.com/kubermatic/kubermatic/issues/10719

**What type of PR is this?**

/kind bug

**Special notes for your reviewer**:

```release-note
Fix a bug where the vmware csi driver requires to run in its dedicated namespace
```

**Documentation**:

```documentation
NONE
```

PR #2327, opened 2022-08-28
Ensure vsphere-csi-driver is removed in kube-system namespace

WeirdMachine:
Removes vsphere csi deployment in kube-system namespace.

Moved to dedicated namespace in https://github.com/kubermatic/kubeone/pull/2292

/kind bug

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/addons/csi-vsphere-ks/validatingwebhook.yaml
+++ b/addons/csi-vsphere-ks/validatingwebhook.yaml
@@ -1,0 +1,157 @@
+{{ if .CSIMigration }}
+# Requires k8s 1.19+
+---
+apiVersion: v1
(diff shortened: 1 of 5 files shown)


xmudrii:
Since we use `kubectl apply --prune`, shouldn't kubectl do this out of the box for us?

WeirdMachine:
> Since we use `kubectl apply --prune`, shouldn't kubectl do this out of the box for us?

Interesting point. For some reason it seems that this does not work between namespaces. Let me validate it.

xmudrii (approved):
/approve

Issue #2318, opened 2022-08-25
KubeOne is failing to provision Flatcar machines upgraded from versions prior to 2969.0.0


xmudrii:
### Problem statement

I'm not able to provision a Kubernetes cluster using KubeOne on Flatcar machines upgraded from a version prior to 2969.0.0. Flatcar was updated by running `sudo update_engine_client -update`, followed by a reboot.

KubeOne is showing the following error on all control plane nodes:

```
Process exited with status 1  node=137.116.220.127
[13.81.11.8] + sudo systemctl restart containerd
[13.81.11.8] Job for containerd.service failed because the control process exited with error code.
[13.81.11.8] See "systemctl status containerd.service" and "journalctl -xeu containerd.service" for details.
ERRO[20:29:50 CEST] ssh: installing kubeadm
```

containerd service is failing with the following error message:

```
Aug 25 18:50:32 marko-1-cp-0 systemd[1]: containerd.service: Main process exited, code=exited, status=1/FAILURE
Aug 25 18:50:32 marko-1-cp-0 systemd[1]: containerd.service: Failed with result 'exit-code'.
Aug 25 18:50:32 marko-1-cp-0 systemd[1]: Failed to start containerd container runtime.
Aug 25 18:50:37 marko-1-cp-0 systemd[1]: containerd.service: Scheduled restart job, restart counter is at 1.
Aug 25 18:50:38 marko-1-cp-0 systemd[1]: Stopped containerd container runtime.
Aug 25 18:50:38 marko-1-cp-0 systemd[1]: Starting containerd container runtime...
Aug 25 18:50:38 marko-1-cp-0 env[2376]: containerd: failed to load TOML: /usr/share/containerd/config-cgroupfs.toml: open /usr/share/containerd/config-cgroupfs.toml: no such file or directory
```

`sudo systemctl status containerd.service` shows the following information:

```
Aug 25 18:26:07 marko-1-cp-0 env[1107]: time="2022-08-25T18:26:07.052470499Z" level=info msg=serving... address=/run/containerd/containerd.sock.ttrpc
core@marko-1-cp-0 ~ $ sudo systemctl status containerd
● containerd.service - containerd container runtime
     Loaded: loaded (/run/systemd/system/containerd.service; enabled; vendor preset: disabled)
    Drop-In: /etc/systemd/system/containerd.service.d
             └─10-kubeone.conf, 10-use-cgroupfs.conf
[...]

PR #2272, opened 2022-08-11
[Test] Run cluster provisioning and upgrading tests on Azure

kron4eg:
Relevant to #2243 and #2249

PR #2331, opened 2022-08-29
Mount /usr/share/ca-certificates in the Azure CCM, update Flatcar to 3227.2.1, enforce NFT backend, and improve Terraform configs

xmudrii:

This PR brings several changes needed to make Azure work on all operating systems:

* Mount `/usr/share/ca-certificates` in the Azure CCM
  * Similar to #2308, this is needed to make the Azure CCM work on Flatcar
* Update Flatcar to 3227.2.1
  * We were initially using a very old version, so this PR bumps it to the latest. This also helps us to avoid issues described in #2318
* Enforce the NFT iptables backend for Canal on RHEL
  * Similar to Flatcar, RHEL uses the NFT backend by default, so this change ensures we don't run into problems with Calico/Canal autodetection mechanisms
* Use the same image reference and plan for the initial MachineDeployment as for the control plane
  * This is disabled for RHEL because machine-controller v1.43 has a stronger dependency on the RHEL version (it works only with RHEL 8.3)

All those changes were tested as part of #2272.

**What type of PR is this?**

/kind bug

```release-note
- Mount `/usr/share/ca-certificates` to the Azure CCM container to fix CrashLoopBackoff on clusters running Flatcar
- Update Flatcar to 3227.2.1 in the example Terraform configs for Azure
- Enforce the NFT iptables backend for Canal and Calico VXLAN on RHEL
- Use the same image reference and plan for the initial Azure MachineDeployment as for the control plane
```

**Documentation**:
```documentation
NONE
```


examples/terraform/azure/output.tf
```diff
@@ -91,6 +91,8 @@ output "kubeone_workers" {
           # assignAvailabilitySet = true/false
           securityGroupName = azurerm_network_security_group.sg.name
           assignPublicIP    = true
+          imageReference    = var.os != "rhel" ? var.image_references[var.os].image : null
+          imagePlan         = length(var.image_references[var.os].plan) > 0 && var.os != "rhel" ? var.image_references[var.os].plan[0] : null
```
kron4eg, line 95: I'd suggest to change this in variables.tf instead, like this:

```diff
diff --git examples/terraform/azure/variables.tf examples/terraform/azure/variables.tf
index 2e12c9ef..508ba4dd 100644
--- examples/terraform/azure/variables.tf
+++ examples/terraform/azure/variables.tf
@@ -159,17 +159,8 @@ variable "image_references" {
     }
 
     rhel = {
-      image = {
-        publisher = "RedHat"
-        offer     = "rhel-byos"
-        sku       = "rhel-lvm85"
-        version   = "8.5.20220316"
-      }
-      plan = [{
-        name      = "rhel-lvm85"
-        publisher = "redhat"
-        product   = "rhel-byos"
-      }]
+      image = null
+      plan = null
       ssh_username = "rhel-user"
       worker_os    = "rhel"
     }
```

xmudrii: This is needed for control plane, so we can't do this.

PR #2344, opened 2022-09-05
Initialize OperatingSystemManager in KubeOneCluster regardless if MachineController enabled or not

kron4eg:
**What this PR does / why we need it**:
Fixes NPE.

**Which issue(s) this PR fixes**:

Fixes #2343

**What type of PR is this?**
/kind bug

**Special notes for your reviewer**:
To trigger the issue config should have:
```
machineController:
  deploy: false

# OSM config have to be omitted from the config
# operatingSystemManager:
#   deploy: false
```

**Does this PR introduce a user-facing change? Then add your Release Note here**:

```release-note
Fixes NPE when machineController deployment is disabled
```

**Documentation**:

```documentation
NONE
```

Issue #2343, opened 2022-09-05
Nil pointer exception in probes upon re-running kubeone on bare metal cluster


embik:
### What happened?

I followed the instructions at https://docs.kubermatic.com/kubeone/v1.5/tutorials/creating-clusters-baremetal/ to create a bare metal cluster with kubeone. For that, I've adapted the `kubeone.yaml` example from that page.

It seems that re-running `kubeone` after the initial setup results in a nil pointer exception:

<details>

```
INFO[10:33:59 CEST] Building Kubernetes clientset...
INFO[10:33:59 CEST] Running cluster probes...
E0905 10:33:59.613224    7209 runtime.go:79] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 1 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic({0x103caf900?, 0x104d71cc0})
	k8s.io/apimachinery@v0.24.2/pkg/util/runtime/runtime.go:75 +0x7c
k8s.io/apimachinery/pkg/util/runtime.HandleCrash({0x0, 0x0, 0x1?})
	k8s.io/apimachinery@v0.24.2/pkg/util/runtime/runtime.go:49 +0x7c
panic({0x103caf900, 0x104d71cc0})
	runtime/panic.go:838 +0x204
k8c.io/kubeone/pkg/tasks.safeguardFlatcarMachineDeployments(0x140008bc1e0?)
	k8c.io/kubeone/pkg/tasks/probes.go:236 +0x2c
k8c.io/kubeone/pkg/tasks.safeguard(0x140008bc1e0)
	k8c.io/kubeone/pkg/tasks/probes.go:154 +0x4e0
k8c.io/kubeone/pkg/tasks.(*Task).Run.func1()
	k8c.io/kubeone/pkg/tasks/task.go:60 +0xa0
k8s.io/apimachinery/pkg/util/wait.ConditionFunc.WithContext.func1({0x14000a07668, 0x1029b1bd0})
	k8s.io/apimachinery@v0.24.2/pkg/util/wait/wait.go:220 +0x28
k8s.io/apimachinery/pkg/util/wait.runConditionWithCrashProtectionWithContext({0x103f5d020?, 0x140001aa010?}, 0x14000a076b8?)
	k8s.io/apimachinery@v0.24.2/pkg/util/wait/wait.go:233 +0x54
[...]

PR #2349, opened 2022-09-13
Initialize OperatingSystemManager even if v1beta1 API is used

kron4eg:
When v1beta1 API is used, internal Cluster doesn't get `.OperatingSystemManager` defaulted, since it's defaulted only for v1beta2 and later converted to internal Cluster.

This is an addition to #2344
regarding #2343

**What type of PR is this?**
/kind bug

```release-note
Fixed NPE regarding OSM for the v1beta1 API in use
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/apis/kubeone/helpers.go
+++ b/pkg/apis/kubeone/helpers.go
@@ -119,10 +119,6 @@ func (h *HostConfig) SetLeader(leader bool) {
 	h.IsLeader = leader
 }
 
-func (c KubeOneCluster) OperatingSystemManagerEnabled() bool {
-	return c.OperatingSystemManager != nil && c.OperatingSystemManager.Deploy
-}
-
 func (crc ContainerRuntimeConfig) MachineControllerFlags() []string {
 	var mcFlags []string
 	switch {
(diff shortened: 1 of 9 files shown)


ahmedwaleedmalik (approved):
/approve

pkg/apis/kubeone/helpers.go
```diff
@@ -119,10 +119,6 @@ func (h *HostConfig) SetLeader(leader bool) {
 	h.IsLeader = leader
 }
 
-func (c KubeOneCluster) OperatingSystemManagerEnabled() bool {
-	return c.OperatingSystemManager != nil && c.OperatingSystemManager.Deploy
-}
-
```
ahmedwaleedmalik, line 125: I still consider this as a safer option since we won't have to depend on the defaulting for this field in code to avoid nil pointer exceptions.

kron4eg: I totally agree, it's safer from "avoid NPE at any costs" perspective. Sometime having an NPE is a good thing though. NPE is a signal that we borked something, and if we suppress it with safety, we'd never know until it's too late. I'd prefer we never had `OperatingSystemManagerEnabled()` in the first place. In that case, probably, we'd hit the bug much earlier and had the fix in the v1.5.0 release :shrug:. But what's done is done.

Now let's see what NPEs have to say to us :joy:.

PR #2353, opened 2022-09-19
Upgrade kubernetes CNI package to v1.1.1

kron4eg:
This is needed in order to unblock kubelet v1.24.5 and v1.25.1, as they has dependency specified as `kubernetes-cni >= 1.1.1`, while older versions has `>= 0.8.7` (which is also can be `1.1.1`).

Otherwise at the moment it's not possible to install kubernetes v1.24.5.

**What type of PR is this?**
/kind bug

```release-note
Fixed installation of kubernetes v1.24.5+
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/scripts/os.go
+++ b/pkg/scripts/os.go
@@ -25,7 +25,7 @@ import (
 )
 
 const (
-	defaultKubernetesCNIVersion = "0.8.7"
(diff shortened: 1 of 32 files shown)


xmudrii:
/hold
I want to double-check something.

xmudrii (approved):
/approve

PR #2354, opened 2022-09-20
Hard-disable OSM when v1beta1 is used

kron4eg:
We don't want to have OSM when v1beta1 is used, to avoid unpleasant surprises,

Fixes #

**What type of PR is this?**
/kind bug

```release-note
Hard-disable OSM when v1beta1 is used
```

**Documentation**:

```documentation
NONE
```


xmudrii (approved):
/approve

Issue #2352, opened 2022-09-16
Cluster autoscaling - new nodes cannot join control plane


cristianrat:
### What happened?

I followed the guide to spin up a kube one cluster with 3 control plane nodes, 2 worker nodes and then I also enabled autoscaling and it worked ok - I tested it

Then a few days later when I scaled it up (by increasing a deployment replicas) new machines are created, but they can't join control plane.

Seems that is the case in 2 clusters - it's fine on the first day, then later it won't work
Note: this is on hetzner
Log says 
```
Sep 16 10:47:06 dev-pool1-77c8b59b86-9xrgs kubelet[6923]: I0916 10:47:06.652377    6923 kubelet_node_status.go:70] "Attempting to register node" node="dev-pool1-77c8b59b86-9xrgs"


Sep 16 10:47:06 dev-pool1-77c8b59b86-9xrgs kubelet[6923]: E0916 10:47:06.655857    6923 kubelet_node_status.go:92] "Unable to register node with API server" err="nodes is forbidden: User \"system:anonymous\" cannot create resource \"nodes\" in API group \"\" at the cluster scope" node="dev-pool1-77c8b59b86-9xrgs"
```
And that log just repeats with various flavours / components 

Deleting the MachineDeployment and MachineSet and re-creating the MachineDeployment usually fixes it - but not really a practical solution
I'm happy to investigate further, if you could point me in the right direction

### Expected behavior

Auto scaling to continue working

### How to reproduce the issue?

Spin up cluster
Setup auto scaling
Go away for 1-2 days
Come back and try to scale

### What KubeOne version are you using?

<details>

```console
$ kubeone version
{
  "kubeone": {
    "major": "1",
    "minor": "5",
    "gitVersion": "1.5.0-beta.0",
    "gitCommit": "3800c3d9d244f76d915bb05aa26e45acafc32158",
    "gitTreeState": "",
    "buildDate": "2022-08-04T14:03:44Z",
    "goVersion": "go1.19",
    "compiler": "gc",
    "platform": "linux/amd64"
  },
  "machine_controller": {
    "major": "1",
    "minor": "53",
    "gitVersion": "v1.53.0",
    "gitCommit": "",
    "gitTreeState": "",
    "buildDate": "",
    "goVersion": "",
    "compiler": "",
    "platform": "linux/amd64"
  }
}
```

</details>

### Provide your KubeOneCluster manifest here (if applicable)

<details>

```yaml
[...]

PR #2355, opened 2022-09-20
install.sh: change VERSION regex to match only stable releases

xmudrii:

The installation script right now uses the following regex to match the KubeOne version: `v([^"]*)`. This matches everything, including prereleases. Because of this, the installation script downloads KubeOne v1.5.0-rc.0 instead of v1.5.0.

This PR changes the regex to `v(([0-9])*\.([0-9])*\.([0-9])*)` so that only stable releases are matched by the script.

xref https://github.com/kubermatic/kubeone/issues/2352#issuecomment-1252114838

**What type of PR is this?**

/kind bug

```release-note
The installation script (`install.sh`) has been modified to match only the stable releases
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -21,7 +21,7 @@ set -eu
 # What OS is used
 OS="$(uname | tr '[:upper:]' '[:lower:]')"
 # Find out what's the latest version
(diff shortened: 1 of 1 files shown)


kron4eg:
/approve
/lgtm

/hold
waiting for #2353 to merge

PR #2356, opened 2022-09-20
New NodeLocalDNS API field

### LINKED ISSUE #2235: Add an Option to disable Nodelocal DNS Cache setup

As a user who runs a SDN specific cni plugin I would like to have an option to disable the default setup of the nodelocal dns cache feature in order to use kubone as k8s deployment tool.

### Solution details

We are running a Cisco ACI SDN in our on-premise data center and are already using the Cisco ACI cni plugin in our setup for running k8s instances. While trying to deploy a k8s cluster with kubeone I realized there is no option to disable the nodelocal dns cache setup. 

In order to have a proper DNS resolution in a k8s cluster when using the Cisco cni plugin a default dns setup without the nodelocal dns cache feature is needed.

kron4eg:

Fixes #2235

**What type of PR is this?**
/kind feature
/kind api-change

```release-note
New NodeLocalDNS API field
```

**Documentation**:

```documentation
TBD
```


xmudrii (changes requested):
machine-controller and OSM addons are using the NodeLocalDNS IP address. We need to adjust those addons as well.

kron4eg:
@xmudrii good catch. Fixed.

xmudrii (approved):
/approve

PR #2369, opened 2022-09-23
Remove old E2E tests in favor of new tests

xmudrii:

This is the last step of the migration to the new E2E (testv2) tests. The new testing framework has proven itself to be better than the old one, to give more coverage, and provides many more test combinations. The new tests also now support all cloud providers natively supported by KubeOne.

This also avoids confusion for contributors in terms of what tests they should run (new or old ones).

This PR:

* Moves `testutil` packages from the old tests to the new tests, and renames the package to `exec`
* Removes the old `test` package
* Renames the `testv2` package to `test`
* Removes old ProwJobs
* Enables AlwaysRun for the new AWS containerd tests

**What type of PR is this?**

/kind feature

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```

/hold
for lazy consensus until EOD


ahmedwaleedmalik:
@xmudrii the generated prow jobs are not updated.

ahmedwaleedmalik (approved):
/approve

test/e2e/helpers.go
```diff
@@ -118,16 +118,16 @@ func requiredTemplateFunc(warn string, input interface{}) (interface{}, error) {
...
 }
 
-func makeBinWithPath(path string, args ...string) *testutil.Exec {
-	return testutil.NewExec("make",
-		testutil.WithArgs(args...),
-		testutil.WithEnv(os.Environ()),
-		testutil.InDir(path),
-		testutil.StdoutDebug,
+func makeBinWithPath(path string, args ...string) *exec.Exec {
+	return exec.NewExec("make",
```
kron4eg, line 126: I'm really not fun of the name, in code it's confusing with `os/exec`. Let's imagine something like, `testexec`, `testbin` or something similar.

xmudrii: That's a good point, I'll update the PR.

xmudrii: @kron4eg Renamed to `testexec`, PTAL if it's okay now.

PR #2391, opened 2022-09-29
SSH host public keys verification

### LINKED ISSUE #1256: Allow users to provide SSH host key for instances

It would be useful if users could provide SSH host keys for instances, so the instance's identity can be validated when SSH-ing to them. Currently, we SSH with the `StrictHostKeyChecking: no` option, so we don't validate the identity.

kron4eg:
Introducing new API fields `HostConfig.SSHHostPublicKey` and `HostConfig.BastionHostPublicKey` which take base64 encoded ssh public keys. At ssh connection time those keys are being verified, to make sure that we are talking to the genuine SSH host.

Fixes #1256

**What type of PR is this?**
/kind feature

```release-note
SSH host public keys verification
```

**Documentation**:

```documentation
NONE
```


examples/terraform/aws/output.tf
```diff
@@ -41,9 +41,11 @@ output "kubeone_hosts" {
       ssh_port             = var.ssh_port
       ssh_private_key_file = var.ssh_private_key_file
       ssh_user             = local.ssh_username
+      ssh_hosts_keys       = var.ssh_hosts_keys
       bastion              = aws_instance.bastion.public_ip
       bastion_port         = var.bastion_port
       bastion_user         = local.bastion_user
+      bastion_host_key     = var.bastion_host_key
```
xmudrii, line 48: Can you please add this to other Terraform examples so that we don't have a big difference between them?

pkg/cmd/config.go
```diff
@@ -812,13 +812,15 @@ addons:
 #     bastion: '4.3.2.1'
 #     bastionPort: 22  # can be left out if using the default (22)
 #     bastionUser: 'root'  # can be left out if using the default ('root')
+#     bastionHostPublicKey: "AAAAC3NzaC1lZDI1NTE5AAAAIGpmWkI5dl7GB3E1hB9LDuju87x9hX5Umw9fih+xXNU+"
```
xmudrii, line 815: Can you add a comment describing this field similar to what we have for other fields?

pkg/cmd/config.go
```diff
@@ -812,13 +812,15 @@ addons:
...
 #     sshPort: 22 # can be left out if using the default (22)
 #     sshUsername: root
 #     # You usually want to configure either a private key OR an
 #     # agent socket, but never both. The socket value can be
 #     # prefixed with "env:" to refer to an environment variable.
 #     sshPrivateKeyFile: '/home/me/.ssh/id_rsa'
 #     sshAgentSocket: 'env:SSH_AUTH_SOCK'
+#     sshHostPublicKey: "AAAAC3NzaC1lZDI1NTE5AAAAIPwEDvXiKfvXrysf86VW5dJTKDlQ09e2tV0+T3KeFKmI"
```
xmudrii, line 823: Can you add a comment describing this field similar to what we have for other fields?

xmudrii (commented):
Some nits, otherwise LGTM.

kron4eg:
@xmudrii suggestions are addressed, PTAL again.

xmudrii (approved):
/approve

PR #2394, opened 2022-10-05
Allow overriding image repository for CoreDNS

### LINKED ISSUE #2023: Proxy registries

As a user I would like to use a proxy registry in order to have control of the accessed images. 

Because proxy registries forward to an upstream registries and not a mirror registries the coredns path breaks when using the `overwriteRegistry` feature. 

Mirror registries are not being utilized because they fallback to the default registry which is a problems due to security reasons. 
Because it is only a proxy the image of corends cannot be just pushed and adjusted.

### Solution details
For coredns to work we need to set `imageRepository` for coredns in the ClusterConfiguration of kubeadm.

### Alternative approaches

Consider mirror registries where it would not fallback to the original registry. Maybe there is an option to control the behavior.

xmudrii:

Right now, we have the overwriteRegistry mechanism that can be used to overwrite image registries for all images used by KubeOne and kubeadm. However, due to architecture decisions in kubeadm, overwriteRegistry doesn't work well for the CoreDNS image.

By default, the CoreDNS image is `registry.k8s.io/coredns/coredns:<version>` (or `k8s.gcr.io` for Kubernetes 1.24 and older). When we overwrite the image (via kubeadm's `dns.imageRepository`), we end up with `my.corp/coredns:<version>`. Note that we have only `/coredns` instead of `/coredns/coredns` when overwriting the CoreDNS image.

This is making the overwriteRegistry feature useless depending on the registry type that you use (e.g. it might be a problem for pull through registries). In the v1beta1 API, we had the AssetConfiguration API that allowed overriding the CoreDNS image explicitly, but we removed that API in v1beta2.

To address this issue, we're introducing a new field (`.features.coreDNS.imageRepository`) in the v1beta2 API that can be used to override the CoreDNS image repository. It remains impossible to override the image tag, it's only possible to override the repository. This field is decoupled from the overwriteRegistry feature, i.e. it can be used even if overwriteRegistry is not used.

Example configuration:

```yaml
kind: KubeOneCluster
versions:
  kubernetes: 1.24.6
cloudProvider:
  aws: {}
  external: false
registryConfiguration:
  overwriteRegistry: "localhost:8080"
features:
  coreDNS:
    imageRepository: "localhost:8080/coredns" # results in localhost:8080/coredns/coredns being used
```

Fixes #2023 

**What type of PR is this?**

/kind feature

```release-note
Allow overriding image repository for CoreDNS via `.features.coreDNS.imageRepository` field.
```

**Documentation**:
```documentation
TBD
```

/assign @kron4eg 
/hold
for discussion

### DIFF
--- a/docs/api_reference/v1beta2.en.md
+++ b/docs/api_reference/v1beta2.en.md
@@ -1,6 +1,6 @@
 +++
 title = "v1beta2 API Reference"
-date = 2022-09-29T16:34:01+03:00
+date = 2022-10-05T12:04:49+02:00
 weight = 11
 +++
 ## v1beta2
@@ -290,6 +290,7 @@ ControlPlaneConfig defines control plane nodes
 | ----- | ----------- | ------ | -------- |
 | replicas |  | *int32 | false |
 | deployPodDisruptionBudget |  | *bool | false |
+| imageRepository | ImageRepository allows users to specify the image registry to be used for CoreDNS. Kubeadm automatically appends `/coredns` at the end, so it's not necessary to specify it. By default it's empty, which means it'll be defaulted based on kubeadm defaults and if overwriteRegistry feature is used. ImageRepository has the highest priority, meaning that it'll override overwriteRegistry if specified. | string | false |
 
 [Back to Group](#v1beta2)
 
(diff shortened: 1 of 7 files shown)


kron4eg:
after giving long and heavy think, Allright let it be.

/lgtm
/approve

PR #2415, opened 2022-10-21
Use smaller instances for AWS E2E tests

xmudrii:

This PR introduces the following changes to the AWS E2E tests:

- Run E2E tests in the Ireland region (`eu-west-1`)
  - This is the cheapest Europen region available
- Use `t3a.small` for the control plane and worker nodes
  - This instance size provides 2 vCPUs and 2 GB RAM which is enough for running E2E tests
  - `t3a` instances are slightly slower than `t3` instances, but they're also cheaper. This doesn't seem to affect the speed of E2E tests at all
- Use `t3a.micro` instances for worker nodes in the upgrade tests
  - We don't run any conformance tests when running upgrade tests, so we don't need big machines
- Use spot instances for the worker nodes
- Use 25 GB disks (50 GB for RHEL) instead of 100 GB for the control plane and 50 GB for worker nodes
- Use `t3a.nano` (`t3.micro` for RHEL) instead of `t3.nano` for the bastion instance

Old setup:

- Paris (`eu-west-3`)
- `t3.medium` and `t3.nano` instances
- Total per run/hour assumes running 4 AWS E2E tests for 1 hour

Disclaimer: the following table doesn't take into account other cloud resources.

| Instance size | Price   | Amount | Total per job | Total per run/hour |
|---------------|---------|--------|---------------|--------------------|
| t3.medium     | $0.0472 | 6      | $0.2832       | $1.1328            |
| t3.nano       | $0.0059 | 1      | $0.0059       | $0.0236            |
|               |         |        | $0.2891       | $1.1564            |

New setup:

* Ireland (`eu-west-1`)
* `t3a.small` and `t3a.nano` instances
* Total per run/hour assumes running 4 AWS E2E tests for 1 hour

Disclaimer: the following table doesn't take into account other cloud resources and using spot instances for the worker nodes.

| Instance size | Price   | Amount | Total per job | Total per run/hour |
|---------------|---------|--------|---------------|--------------------|
| t3a.small     | $0.0204 | 6      | $0.1224       | $0.4896            |
| t3a.nano      | $0.0051 | 1      | $0.0051       | $0.0204            |
|               |         |        | $0.1275       | $0.51              |

**Savings per job/hour: $0,1616**
**Savings per run/hour: $0,6464**

Considered:

- Using spot instances for the control plane nodes
  - Requires significant changes to the Terraform configs
- Using even smaller instances for worker nodes
  - Tested `t3.micro`, but it only has 1 GB RAM which is not enough for running the conformance tests. All tests were failing.

**What type of PR is this?**

/kind chore

```release-note
Make volume size for worker nodes configurable in Terraform configs for AWS (50 GB by default)
```

**Documentation**:
```documentation
NONE
```


test/e2e/tests_definitions.go
```diff
@@ -41,6 +41,14 @@ var (
 				vars: []string{
 					"disable_kubeapi_loadbalancer=true",
 					"subnets_cidr=27",
+					// Use smaller instances in Ireland (the cheapest EU region)
```
kron4eg, line 44: looks like varsFile would be a better fit for most cases.

xmudrii: @kron4eg Updated, PTAL.

test/e2e/testdata/aws_stable.tfvars
```diff
@@ -0,0 +1,10 @@
+disable_kubeapi_loadbalancer = true
+subnets_cidr                 = 27
+
+# Use smaller instances in Ireland for E2E tests
+aws_region                                        = eu-west-1
+control_plane_type                                = t3a.small
+control_plane_volume_size                         = 25
+worker_type                                       = t3a.small
+bastion_type                                      = t3a.nano
```
kron4eg, line 9: looks like you need to take values in quotes

xmudrii: Fixed. :sweat_smile:

ahmedwaleedmalik (approved):
/approve

PR #2419, opened 2022-10-23
Update etcd to 3.5.5 or use the version provided by kubeadm if it's newer

xmudrii:

We are manually setting the etcd version to 3.5.3 because previous versions are affected by a data integrity bug. Additionally, versions up to 3.5.5 (not including 3.5.5) have a [durability bug](https://groups.google.com/a/kubernetes.io/g/dev/c/7q4tB_Vp3Uc/m/MrHalhCIBAAJ) affecting single node clusters.

With the current approach of manually setting the etcd version, we might quickly get behind on etcd versions. For example, Kubernetes might start using a newer version while we still enforce the old version.

To mitigate all those issues, we're now:

- Overriding etcd version only if used kubeadm version is known to use an affected etcd version
- Using the default etcd version provided by kubeadm if used kubeadm version uses etcd not affected by those bugs

**What type of PR is this?**

/kind feature

Kubernetes 1.25.4, 1.24.8, and 1.23.12 are not yet released.

```release-note
Update etcd to 3.5.5 or use the version provided by kubeadm if it's newer
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/pkg/templates/kubeadm/v1beta3/kubeadm.go
+++ b/pkg/templates/kubeadm/v1beta3/kubeadm.go
@@ -31,6 +31,7 @@ import (
 	"k8c.io/kubeone/pkg/fail"
 	"k8c.io/kubeone/pkg/features"
 	"k8c.io/kubeone/pkg/kubeflags"
+	"k8c.io/kubeone/pkg/semverutil"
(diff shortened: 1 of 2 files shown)


ahmedwaleedmalik (approved):
/approve

PR #2428, opened 2022-10-27
Helm addons

### LINKED ISSUE #2398: Write a proposal for supporting Helm charts as KubeOne addons
We should come up with a proposal (at least an informal one) for supporting Helm charts as KubeOne addons. The proposal should at least answer the following questions:

* Do we want to use Helm to deploy core components such as CNI plugins and CSI drivers? Is that possible at all and how that would look like?
* Do we want to allow users to provide their own Helm charts to be deployed by KubeOne (e.g. deploy some of their applications)?
* Do we want to allow users to override Helm charts used by KubeOne with their own Helm charts?
* Do we want to allow users to customize `values.yaml`?

Additionally, the proposal should show:

* How we can integrate Helm in KubeOne? Do we want to call it over SSH? Do we want to integrate it as a library?
  * We can see how is that done in KKP as an inspiration
* How could the API look like for interacting with Helm charts (e.g. how to provide values, enable Helm charts -- an example for each supported operation)?

We should consider the following when writing a proposal:

* KKP has a feature called Applications which is based on Helm. We should use it as an inspiration
* There's a proposal for managing CNI addons via Helm charts in KKP. We should try to follow the same approach as KKP for this

kron4eg:

Fixes #2398

**What type of PR is this?**
/kind design

helm libs imports are provided just for demonstration purposes, and will be deleted prior merge

```release-note
helm addons proposal
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/docs/proposals/20221021-helm-addons.md
+++ b/docs/proposals/20221021-helm-addons.md
@@ -1,0 +1,130 @@
+# KubeOne Helm Addons
+
+**Status:** **Draft**
+**Created:** 2022-10-27
+**Last updated:** 2022-10-27
+**Author:** Artiom Diomin ([@kron4eg](https://github.com/kron4eg))
+
+## Abstract
+
+There is a demand for addons that are helm charts. TBD.
+
+### FAQ
+---
+Q: Is that possible at all and how that would look like?
+
+A: Yes, it's possible. Helm can be integrated as library and helm charts can be
+embedded (like our current addons).
+
+---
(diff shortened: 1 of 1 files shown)


ahmedwaleedmalik (commented):
Was randomly going through this, looks great :D

kron4eg:
/assign @xmudrii 

PTAL

ahmedwaleedmalik (approved):
/approve ⭐

xmudrii (commented):
Retroactively,
/lgtm
/approve

PR #2463, opened 2022-11-08
Move the kubeone-e2e image to Quay

xmudrii:

The `kubeone-e2e` image is the only remaining image on Docker Hub that we have. It's time to move this image to Quay, which is done with this PR. I already pushed the image manually, so all jobs are updated as well.

**What type of PR is this?**
/kind chore

```release-note
The `kubeone-e2e` image is moved from Docker Hub to Quay (`quay.io/kubermatic/kubeone-e2e`)
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/.prow/postsubmits.yaml
+++ b/.prow/postsubmits.yaml
@@ -9,7 +9,7 @@ postsubmits:
     branches:
       - ^main$
     labels:
-      preset-docker-push: "true"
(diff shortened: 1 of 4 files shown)


ahmedwaleedmalik (approved):
/approve ⚡

PR #2501, opened 2022-11-24
Use `registry.k8s.io` instead of `k8s.gcr.io` for Kubernetes upstream images

### LINKED ISSUE #2459: Use `registry.k8s.io` in kubeadm for all Kubernetes versions
Starting with December patch releases, kubeadm will use `registry.k8s.io` instead of the `k8s.gcr.io` image registry by default. To avoid any potential confusion, starting with the December KubeOne patch releases, we'll unconditionally use `registry.k8s.io` for all Kubernetes versions.

### LINKED ISSUE #2460: Use `registry.k8s.io` instead of `k8s.gcr.io` for all components that we deploy
The scope of this ticket is replacing `k8s.gcr.io` with `registry.k8s.io` for all components that we deploy in [`images.go`](https://github.com/kubermatic/kubeone/blob/main/pkg/templates/images/images.go) and [`addons/`](https://github.com/kubermatic/kubeone/tree/main/addons).

xmudrii:

This change has been announced in the changelogs for v1.5.3 and v1.4.11:

> For the next series of KubeOne and KKP patch releases, image references will move from `k8s.gcr.io` to `registry.k8s.io`. This will be done to keep up with [the latest upstream changes](https://github.com/kubernetes/enhancements/tree/master/keps/sig-release/3000-artifact-distribution). Please ensure that any mirrors you use are able to host `registry.k8s.io` and/or that firewall rules are going to allow access to `registry.k8s.io` to pull images before applying the next KubeOne patch releases. This is not included in this patch release but just a notification of future changes.

`k8s.gcr.io` is facing many problems because its parent GCP project is running out of credits. Google credits CNCF/Kubernetes with 3M$ every year. 2/3 out of those credits go on `k8s.gcr.io`. This year that wasn't enough and Google had to give 600k$ more, followed by an additional 100k$.

Earlier, Kubernetes pushed its images to three GCP regions (US, EU, and Asia). To overcome issues, they're now distributing images to many more regions, but also to some AWS regions. `registry.k8s.io` is a new image registry that's able to determine what's the closest registry to you and based on that redirect your request. All images available on `k8s.gcr.io` are also available on `registry.k8s.io`.

Starting with the December Kubernetes patch releases, kubeadm and kubelet will use `registry.k8s.io` by default. We decided to join the upstream efforts and migrate to `registry.k8s.io` by default for all Kubernetes releases that we support. Not only that this is going to help upstream, but it's also going to reduce confusion on why is suddenly 1.25.5 using `registry.k8s.io`, but 1.25.4 is using `k8s.gcr.io`.

The same change has been made to KKP as part of https://github.com/kubermatic/kubermatic/pull/11079. Same as for KKP, we'll cherry-pick this change to the supported release branches (`release/v1.5` and `release/v1.4`).

Fixes #2459
Fixes #2460

**What type of PR is this?**

/kind feature
/kind api-change

```release-note
[ACTION REQUIRED] Image references are changed from `k8s.gcr.io` to `registry.k8s.io`. This is done to keep up with [the latest upstream changes](https://github.com/kubernetes/enhancements/tree/master/keps/sig-release/3000-artifact-distribution). Please ensure that any mirrors you use are able to host `registry.k8s.io` and/or that firewall rules are going to allow access to `registry.k8s.io` to pull images before applying the next KubeOne patch releases.
```

**Documentation**:
```documentation
TBD
```

### DIFF
--- a/pkg/scripts/kubeadm.go
+++ b/pkg/scripts/kubeadm.go
@@ -67,7 +67,7 @@ var (
 
 	kubeadmPauseImageVersionScriptTemplate = heredoc.Doc(`
 		sudo kubeadm config images list --kubernetes-version={{ .KUBERNETES_VERSION }} |
-			grep "k8s.gcr.io/pause" |
+			grep "registry.k8s.io/pause" |
 			cut -d ":" -f2
 	`)
 )
--- a/pkg/scripts/kubeadm.go
+++ b/pkg/scripts/kubeadm.go
@@ -66,7 +66,7 @@ var (
 	`)
 
 	kubeadmPauseImageVersionScriptTemplate = heredoc.Doc(`
-		sudo kubeadm config images list --kubernetes-version={{ .KUBERNETES_VERSION }} |
+		sudo kubeadm config images list --image-repository=registry.k8s.io --kubernetes-version={{ .KUBERNETES_VERSION }} |
 			grep "registry.k8s.io/pause" |
 			cut -d ":" -f2
 	`)
--- a/addons/csi-vmware-cloud-director/csi-controller.yaml
+++ b/addons/csi-vmware-cloud-director/csi-controller.yaml
@@ -108,7 +108,7 @@ spec:
       dnsPolicy: Default
       containers:
         - name: csi-attacher
-          image: k8s.gcr.io/sig-storage/csi-attacher:v3.2.1
+          image: {{ .InternalImages.Get "VMwareCloudDirectorCSIAttacher" }}
           imagePullPolicy: IfNotPresent
           args:
             - --csi-address=$(ADDRESS)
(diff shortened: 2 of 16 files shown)


pkg/scripts/kubeadm.go
```diff
@@ -67,7 +67,7 @@ var (
 
 	kubeadmPauseImageVersionScriptTemplate = heredoc.Doc(`
 		sudo kubeadm config images list --kubernetes-version={{ .KUBERNETES_VERSION }} |
-			grep "k8s.gcr.io/pause" |
+			grep "registry.k8s.io/pause" |
```
kron4eg, line 70: This script have to handle both cases, for older k8s versions will still return k8s.gcr.io.

xmudrii: @kron4eg I just pushed a commit to fix this.

kron4eg: The problem is: at the time when this script run, the config file is not yet uploaded. You have to use the flag `--image-repository {{ .REGISTRY }}`

xmudrii: Fixed, PTAL.

PR #2513, opened 2022-11-30
fix build date on macos

kossmac:
use correct date parameters when using "make build" on macos.
macos' date is not aware of --rfc3339 parameter.

> date: illegal option -- -
usage: date [-jnRu] [-r seconds|file] [-v[+|-]val[ymwdHMS]]
            [-I[date | hours | minutes | seconds]]
            [-f fmt date | [[[mm]dd]HH]MM[[cc]yy][.ss]] [+format]

Signed-off-by: Karsten Kosmala <kosmala@cosmocode.de>

/kind bug
/kind chore

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,13 +32,21 @@ MAKEFLAGS += --warn-undefined-variables
 # disable magic rules
 MAKEFLAGS += --no-builtin-rules
 
+detected_OS := $(shell sh -c 'uname 2>/dev/null || echo Unknown')
+
 export GOPATH?=$(shell go env GOPATH)
 export CGO_ENABLED=0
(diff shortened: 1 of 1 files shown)


embik:
Hi @kossmac, this logic is already macOS-aware but expects `coreutils` to be installed (for the `gdate` command, which doesn't exist on Linux and as such doesn't make sense in the "Linux" condition). Would that also work for you?

kossmac:
Hi @embik, sure thing, since you have to install other things to build anyway. Tested it with coreutils from homebrew.

embik:
Glad to hear that things are working as expected now!

PR #2528, opened 2022-12-08
Add an option for tasks to mutate the State object

xmudrii:

We have a task to determine the pause image by asking kubeadm what pause image we should use: https://github.com/kubermatic/kubeone/blob/5a489d65729cbb4cac0166af31f6e556a541e736/pkg/tasks/kubeadm_config.go#L30-L54

The output from kubeadm is taken and the State object is updated by using `state.PauseImage = ...`. However, that doesn't work because when running a task, we're creating a shallow copy of the State object and we use that one for running the task: https://github.com/kubermatic/kubeone/blob/5a489d65729cbb4cac0166af31f6e556a541e736/pkg/state/task.go#L67

Copying the State object is required because we might be running a task concurrently and multiple goroutines editing the same struct concurrently is never a good idea. But we still need a way to mutate the State from a task, for example, in this case, the only way to get the pause image is to ask kubeadm.

This issue is addressed by adding a state mutator function to the task executor (`RunTaskOnNodes`). This mutator function is guarded with a mutex so that it's safe to run it concurrently. This PR also integrates this mutator into a task responsible for getting the pause image, so this is now supposed to work properly.

**What type of PR is this?**

/kind bug

```release-note
Use the pause image from `registry.k8s.io` for all Kubernetes releases
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/pkg/state/task.go
+++ b/pkg/state/task.go
@@ -54,11 +54,14 @@ func (s *State) runTask(node *kubeoneapi.HostConfig, task NodeTask) error {
 	return fail.Runtime(task(s, node, conn), "")
 }
 
+type stateMutatorFn func(original *State, tmp *State)
+
 // RunTaskOnNodes runs the given task on the given selection of hosts.
-func (s *State) RunTaskOnNodes(nodes []kubeoneapi.HostConfig, task NodeTask, parallel RunModeEnum) error {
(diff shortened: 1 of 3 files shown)


embik (approved):

ahmedwaleedmalik (approved):
/approve

PR #2541, opened 2022-12-12
Run E2E tests only if relevant files are changed

xmudrii:

This PR ensures that we run the E2E tests only if the relevant files are changed: packages, addons, scripts, and tests themselves.

The key reason behind this change is that we don't want to run the full E2E suite when updating changelogs. That's slowing down the release process a lot, as we have to wait for all the E2E tests on both main and release branches. The same goes for updating the in-repo docs.

**What type of PR is this?**

/kind cleanup

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/test/e2e/helpers.go
+++ b/test/e2e/helpers.go
@@ -385,15 +385,16 @@ func parseContainerImageVersion(image string) (*semver.Version, error) {
 }
 
 type ProwJob struct {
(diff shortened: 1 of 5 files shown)


ahmedwaleedmalik (approved):
/approve

Issue #2492, opened 2022-11-21
Add IPv6+IPv4 family support for OSM/MC


PratikDeoghare:
In KubeOne we have introduced IPFamily IPv6+IPv4. 
OSM/MC does not know about it yet. 
This story will track its implementation.

PR #2543, opened 2022-12-14
add support for AWS x Cilium x IPv6+IPv4

PratikDeoghare:

Add support for AWS x Cilium x IPv6+IPv4.

Fixes https://github.com/kubermatic/kubeone/issues/2492

**What type of PR is this?**

/kind feature 

```release-note
Adds support for  IPv6+IPv4 family for AWS x Cilium
```

**Documentation**:

```documentation
TBD
```


pkg/etcdutil/etcd.go
```diff
@@ -52,8 +52,15 @@ func NewClientConfig(s *state.State, host kubeoneapi.HostConfig) (*clientv3.Conf
 		return nil, fail.Etcd(err, "TLS config creating")
 	}
 
+	var endpoints []string
+	if s.Cluster.ClusterNetwork.IPFamily.IsIPv6Primary() {
+		endpoints = []string{fmt.Sprintf("[%s]:2379", host.IPv6Addresses[0])}
```
xmudrii, line 57: Maybe we should verify that this IPv6 address/element of the slice exists to avoid eventual NPEs. I know we have some validation for this, but better to be on the safe side.

pkg/templates/kubeadm/v1beta3/kubeadm.go
```diff
@@ -98,6 +98,13 @@ func NewConfig(s *state.State, host kubeoneapi.HostConfig) ([]runtime.Object, er
 
 	controlPlaneEndpoint := fmt.Sprintf("%s:%d", cluster.APIEndpoint.Host, cluster.APIEndpoint.Port)
 
+	var advertiseAddress string
+	if s.Cluster.ClusterNetwork.IPFamily.IsIPv6Primary() {
+		advertiseAddress = host.IPv6Addresses[0]
```
xmudrii, line 103: The same goes here.

addons/ccm-aws/ccm-aws.yaml
```diff
@@ -165,7 +165,26 @@ spec:
...
+          secret:
+            defaultMode: 420
+            secretName: cloud-config
+---
+apiVersion: v1
+data:
+  config: {{ awsCCMCloudConfig .Config.Name .Config.ClusterNetwork.IPFamily | b64enc }}
+kind: Secret
+metadata:
+  name: cloud-config
```
xmudrii, line 188: We tend to allow users to provide their own cloud-config whenever that's possible. I think we should do the same here. That being said, the cloud-config should be taken from the API (`.cloudProvider.cloudConfig`). We can default cloud-config there if needed.

pkg/addons/manifest.go
```diff
@@ -295,10 +296,33 @@ func txtFuncMap(overwriteRegistry string) template.FuncMap {
...
+	}
+
+	switch ipFamily {
+	case kubeoneapi.IPFamilyIPv4:
+		lines = append(lines, fmt.Sprintf("NodeIPFamilies=%q", "ipv4"))
+	case kubeoneapi.IPFamilyIPv6:
+		lines = append(lines, fmt.Sprintf("NodeIPFamilies=%q", "ipv6"))
+	case kubeoneapi.IPFamilyIPv4IPv6:
+		lines = append(lines, fmt.Sprintf("NodeIPFamilies=%q", "ipv4"))
+		lines = append(lines, fmt.Sprintf("NodeIPFamilies=%q", "ipv6"))
```
xmudrii, line 317: Note for future readers: I checked to make sure it's possible to specify `NodeIPFamilies` multiple times like this and it is. Here's an example I found in the AWS CCM tests: https://github.com/kubernetes/cloud-provider-aws/blob/15fa84711702e2a57997c934d2963b29284f8bdc/pkg/providers/v1/aws_test.go#L238

addons/ccm-aws/ccm-aws.yaml
```diff
@@ -165,7 +165,17 @@ spec:
             - --cloud-provider=aws
             - --use-service-account-credentials=false
             - --configure-cloud-routes=false
+            - --cloud-config=/etc/kubernetes/cloud/cloud-config
```
PratikDeoghare, line 168: todo: add conditions

PratikDeoghare:
Cyclo complexity went up a little. Lemme see if it can be fixed.

examples/terraform/aws-dualstack/credentials_ipv4+ipv6.yaml
```diff
@@ -0,0 +1,5 @@
+cloudConfig: |
```
xmudrii, line 1: Let's replace `+` in the filename with `_` if possible. I tend to avoid using special symbols in filenames for compatibility reasons.

examples/terraform/aws-dualstack/credentials_ipv6+ipv4.yaml
```diff
@@ -0,0 +1,5 @@
+cloudConfig: |
```
xmudrii, line 1: We need some documentation regarding how this file is supposed to be used. You can update the README file for those configs for now.

pkg/apis/kubeone/v1beta2/defaults.go
```diff
@@ -69,6 +70,15 @@ func SetDefaults_KubeOneCluster(obj *KubeOneCluster) {
 	SetDefaults_HelmReleases(obj)
 	SetDefaults_SystemPackages(obj)
 	SetDefaults_Features(obj)
+	SetDefaults_CloudConfig(obj)
+}
+
+func SetDefaults_CloudConfig(obj *KubeOneCluster) {
+	if obj.CloudProvider.AWS != nil {
```
xmudrii, line 77: Shouldn't we do this only if external CCM is used? Or this should be used with the in-tree CCM as well?

pkg/apis/kubeone/validation/validation.go
```diff
@@ -174,11 +174,14 @@ func ValidateAPIEndpoint(a kubeoneapi.APIEndpoint, fldPath *field.Path) field.Er
...
 
 // ValidateCloudProviderSpec validates the CloudProviderSpec structure
-func ValidateCloudProviderSpec(p kubeoneapi.CloudProviderSpec, fldPath *field.Path) field.ErrorList {
+func ValidateCloudProviderSpec(p kubeoneapi.CloudProviderSpec, networkConfig kubeoneapi.ClusterNetworkConfig, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
 	providerFound := false
 	if p.AWS != nil {
+		if networkConfig.IPFamily.IsDualstack() && len(p.CloudConfig) == 0 {
+			allErrs = append(allErrs, field.Required(fldPath.Child("cloudConfig"), ".cloudProvider.cloudConfig is required for dualstack clusters for aws provider"))
```
xmudrii, line 183: Field is already included in the error message:

```suggestion
			allErrs = append(allErrs, field.Required(fldPath.Child("cloudConfig"), "cloudConfig is required for dualstack clusters for aws provider"))
```

pkg/apis/kubeone/validation/validation.go
```diff
@@ -174,11 +174,14 @@ func ValidateAPIEndpoint(a kubeoneapi.APIEndpoint, fldPath *field.Path) field.Er
 }
 
 // ValidateCloudProviderSpec validates the CloudProviderSpec structure
-func ValidateCloudProviderSpec(p kubeoneapi.CloudProviderSpec, fldPath *field.Path) field.ErrorList {
+func ValidateCloudProviderSpec(p kubeoneapi.CloudProviderSpec, networkConfig kubeoneapi.ClusterNetworkConfig, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
 	providerFound := false
 	if p.AWS != nil {
+		if networkConfig.IPFamily.IsDualstack() && len(p.CloudConfig) == 0 {
```
xmudrii, line 182: Do we need cloud-config for IPv6-only clusters?

pkg/apis/kubeone/validation/validation_test.go
```diff
@@ -738,7 +738,7 @@ func TestValidateCloudProviderSpec(t *testing.T) {
 	for _, tc := range tests {
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
-			errs := ValidateCloudProviderSpec(tc.providerConfig, nil)
+			errs := ValidateCloudProviderSpec(tc.providerConfig, kubeoneapi.ClusterNetworkConfig{}, nil)
```
xmudrii, line 741: Please add some tests for the case you added.

xmudrii (changes requested):

pkg/apis/kubeone/v1beta2/defaults.go
PratikDeoghare: only external 
fixing...

pkg/apis/kubeone/validation/validation.go
PratikDeoghare: No. It worked without any special config when I hacked it together.

pkg/apis/kubeone/validation/validation_test.go
PratikDeoghare: done

examples/terraform/aws-dualstack/credentials_ipv6+ipv4.yaml
PratikDeoghare: done

xmudrii (approved):
/lgtm
/approve

Issue #2520, opened 2022-12-01
Evaluate different UI options for the interactive subcommand


xmudrii:

xmudrii:
## Charm / Bubble Tea

[Bubble Tea](https://github.com/charmbracelet/bubbletea) from [Charm](https://charm.sh/) is a tool for building powerful and beautiful terminal user interfaces (TUIs). They support many components, such as text fields, text areas, various types of lists, spinners, progress bars, and many more... In general, it's an amazing tool that provides many possibilities.

However, all those features are increasing the complexity. Bubble Tea brings a lot of options but requires a lot of code to wire different features together. [The `views` example](https://github.com/charmbracelet/bubbletea/tree/master/examples/views) is a good example of what it takes to build a more complex UI: you need to handle a lot of UI-related tasks yourself, such as creating a layout, choosing colors, handling keystrokes... Overall, you need a solid knowledge of the whole ecosystem (Bubble Tea and relevant libraries) to be efficient with it and build nice-looking TUIs.

## go-survey

[`go-survey`](https://github.com/go-survey/survey) is a simple library for building interactive prompts. It offers inputs, selects, and multi-selects. There are basic options, such as tooltips, validators, filters (for select/multi-select), and suggestions. It doesn't bring much flexibility, e.g. you can't choose colors, change the views, or run it as a full-screen application.

As a result, building an application using `go-survey` is very simple. You just need to define questions and use `survey.Ask` or `survey.AskOne`. You can easily decide what questions to ask based on user input just by calling `survey.Ask` with a different set of questions, while with Bubble Tea, you might need to build different views and wire them together.

## Verdict

I believe we should go further with `go-survey`. Bubble Tea is definitely an amazing tool, but what we need is a relatively simple questionnaire to guide users through the init process. We could build a very nice-looking TUI with Bubble Tea, but it would be very hard to maintain it for folks who don't have experience with the Charm ecosystem.

`go-survey` seems to be well-maintained and it works pretty well. GitHub uses `go-survey` for their CLI, so if it works for them, I think it should work for us as well.

/close

PR #2552, opened 2022-12-22
Implement an interactive mode for `kubeone init` subcommand

### LINKED ISSUE #2455: Implement an interactive mode for `kubeone init` subcommand

We want to implement an interactive mode for the `kubeone init` subcommand to make getting started with KubeOne as easiest as possible. The interactive mode should guide users through creating the proper KubeOneCluster manifest and `terraform.tfvars` file. We should ask questions such as (non-exhaustive list):

* What cluster name do you want to use?
* What could provider do you want to use?
* Depending on the selected cloud provider, ask users to provide values for the most important and relevant Terraform variables
  * For the most of cloud providers, that might be region/availability zone, instance size, disk size...
  * Ask users what operating system they want to use
* Ask users what number of worker nodes they want
* Recommend users some of the features that we have and ask them if they want to enable any of those features

Each question should be accompanied by a description of what the value means and how should you choose a value.

The result should be the same as with non-interactive mode, but with values that the user provided (i.e. generated Terraform configs, `terraform.tfvars`, and `kubeone.yaml`).

### Solution details

- Users can run something like `kubeone init --interactive` and get the behavior described above
- Users can use `kubeone init` in non-interactive mode (i.e. keep the current behavior as well)

### Alternative approaches

We considered not implementing this feature, but we think this might have a very positive effect on the getting started experience, so we want to give it a try.

xmudrii:

This PR implements an interactive mode for the `kubeone init` command. In this mode, KubeOne asks questions about the desired cluster leading to generating KubeOneCluster manifests and Terraform configs based on the user's input. More about the rationale can be found in #2455.

This has been implemented using the [go-survey](https://github.com/go-survey/survey) library. The decision to use this library comes from the research documented in #2520.

The interactive mode covers questions such as:
* Cluster name, (cloud) provider, Kubernetes version
* Features to enable (encryption providers and CoreDNS PodDisruptionBudget for now)
* Addons to enable (cluster-autoscaler and backups-restic for now)
* Number of control plane and worker nodes, operating system, SSH public key

The interactive mode is invoked as:

```shell
kubeone init -i
```

Note: other `kubeone init` flags are currently ignored if the interactive mode is used.

Here's an asciinema presenting the interactive mode: https://asciinema.org/a/547401

Fixes #2455 

**What type of PR is this?**
/kind feature

```release-note
Implement an interactive mode for the `kubeone init` subcommand
```

**Documentation**:
```documentation
TBD
```


pkg/cmd/initcmd/validators.go
```diff
@@ -0,0 +1,99 @@
...
+		i, err := strconv.Atoi(str)
+		if err != nil {
+			return fmt.Errorf("provided value must be a valid integer, but got %v", reflect.TypeOf(val).Name())
+		}
+		if i <= 0 {
+			return fmt.Errorf("provided value must be positive, but got %d", i)
+		}
+	}
+
+	return nil
```
kron4eg, line 71: This is a bug, if `val` would be an int32 (or any other int*) the function will not check its positiveness

pkg/cmd/initcmd/validators.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func oddNumberValidator(val interface{}) error {
+	if i, ok := val.(int); ok {
+		if i <= 0 {
+			return fmt.Errorf("provided value must be positive, but got %d", i)
```
kron4eg, line 77: probably `positiveNumberValidator(val)` should be used here

pkg/cmd/initcmd/helpers.go
```diff
@@ -0,0 +1,120 @@
...
+
+func stringSliceIncludes(slice []string, val string) bool {
+	for _, v := range slice {
+		if v == val {
+			return true
+		}
+	}
+
+	return false
+}
```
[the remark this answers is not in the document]
xmudrii: I love generics.

pkg/cmd/initcmd/providers.go
```diff
@@ -0,0 +1,412 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package initcmd
+
+import (
+	"github.com/MakeNowJust/heredoc"
```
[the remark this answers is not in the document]
xmudrii: My IDE has failed me. :joy:

pkg/cmd/initcmd/validators.go
xmudrii: We can't check for every int type as that would get really ugly, but I'll at least return an error if we encounter an "unknown" type (instead of returning nil).

pkg/cmd/initcmd/validators.go
xmudrii: We would anyways need to ask for type, so we wouldn't get much. I at least DRY-ed the function a little bit.

xmudrii:
@kron4eg Comments are addressed.

kron4eg:
/lgtm
/approve

:zap:

PR #2560, opened 2022-12-26
Remove leftover `/tmp/k1-etc-environment` file

### LINKED ISSUE #2547: Kubeone fails on consecutive runs if the ssh username changed
This is a very niche problem, I guess, so not sure a fix is desired. On the other hand, the fix should be really simple.

### What happened?

If the ssh username changes, kubeone will fail on this line:
https://github.com/kubermatic/kubeone/blob/main/pkg/scripts/proxy.go#L62

This is because the file still exists from the previous run, owned by the old user.

### Expected behavior

Kubeone should not fail, either because it first removes the temporary file or because the previous run deleted it at the end.

1. Create a cluster with kubeone where the ssh username in the TF output is set to X
2. Change the TF output such that the ssh username is now Y
3. Run kubeone again and see it fail because /tmp/k1-etc-environment is owned by X

Not sure whether this always happens or whether kubeone must be triggered to run a certain task. In our case, one of the 3 nodes was replaced, i.e. a new node was added. It failed on the existing nodes, but not the new node.

1.5.4 (sorry, this is running in a pipeline, can't just run kubeone version there easily)

### What cloud provider are you running on?

OpenStack

### What operating system are you running in your cluster?

Flatcar and Ubuntu 22.04

### Additional information

Kubeone fails here:
- https://github.com/kubermatic/kubeone/blob/main/pkg/scripts/proxy.go#L62
- https://github.com/kubermatic/kubeone/blob/main/pkg/scripts/proxy.go#L64

The fix should be easy, either:
1. delete \$envtmp after line 65 (thus the file won't exist on the next run and thus won't have the wrong owner), OR
2. delete \$envtmp before line 62 with sudo provileges

Last but not least, our use case: we're changing the OS, and thus the ssh username changed as well (from core to ubuntu).

xmudrii:

This PR fixes the issue reported in #2547 where we leave a file owned by potentially a non-root user.

Fixes #2547 

**What type of PR is this?**
/kind bug

```release-note
Remove the leftover `/tmp/k1-etc-environment` file. This fixes an issue with `kubeone apply` failing if the username is changed
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/pkg/scripts/proxy.go
+++ b/pkg/scripts/proxy.go
@@ -59,6 +59,8 @@ export NO_PROXY no_proxy
 EOF
 
 envtmp=/tmp/k1-etc-environment
+# ensure we delete old file to avoid errors e.g. if username changed
+sudo rm -f /tmp/k1-etc-environment
 grep -v '#kubeone$' /etc/environment > $envtmp || true
 set +o pipefail # grep exits non-zero without match
 grep = /etc/kubeone/proxy-env | sed 's/$/#kubeone/' >> $envtmp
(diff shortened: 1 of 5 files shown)


kron4eg:
@xmudrii you've forgot to update fixtures

PR #2602, opened 2023-01-18
Use main revision for 1.26 upgrade tests

xmudrii:

This PR adds support for specifying what KubeOne version to use for initializing a cluster in the upgrade tests.

We use this feature for 1.26 upgrade tests because the current stable release (1.5) doesn't support 1.25, so we can't create a 1.25 cluster with KubeOne 1.5, and then upgrade it to 1.26 with KubeOne based on the latest revision.

This is supposed to be a temporary workaround, hoping that it'll not happen that we add support for two minor Kubernetes releases in a single KubeOne release.

**What type of PR is this?**
/kind failing-test

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/test/e2e/tests_definitions.go
+++ b/test/e2e/tests_definitions.go
@@ -1244,6 +1244,10 @@ type Scenario interface {
 	Run(context.Context, *testing.T)
 }
 
+type ScenarioStable interface {
+	SetInitKubeOneVersion(version string)
+}
+
 type ProwConfig struct {
 	AlwaysRun    bool
 	RunIfChanged string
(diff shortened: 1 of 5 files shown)


test/e2e/tests_definitions.go
```diff
@@ -1240,6 +1240,7 @@ const (
 type Scenario interface {
 	SetInfra(infrastructure Infra)
 	SetVersions(versions ...string)
+	SetInitKubeOneVersion(version string)
```
kron4eg, line 1243: Given that only 1 scenario needs it, it looks like this is the case for [interface smuggling](https://utcc.utoronto.ca/~cks/space/blog/programming/GoInterfaceSmuggling).

kron4eg: I.e. make SetInitKubeOneVersion() own interface, and use type assertion if scenario implements it.

xmudrii: @kron4eg Good point. Fixed, PTAL.

Issue #2607, opened 2023-01-20
kubernetes-cni pinned to 1.1.1*, but kubeadm requires >=1.2.0 for k8s 1.24.10


7oku:
### What happened?
Upgrading kubernetes from 1.24.9 to 1.24.10 via kubeone 1.5.5 fails. 

### Expected behavior
kubeone should be able to upgrade kubernetes

### How to reproduce the issue?
run kubeone apply -f kubeone.yml 

### What KubeOne version are you using?
kubeone 1.5.5

```
{
  "kubeone": {
    "major": "1",
    "minor": "5",
    "gitVersion": "1.5.5",
    "gitCommit": "99303c67363dc3788e24ff47b4ae0eae3ac7b8ae",
    "gitTreeState": "",
    "buildDate": "2023-01-17T10:27:32Z",
    "goVersion": "go1.19.4",
    "compiler": "gc",
    "platform": "linux/amd64"
  },
  "machine_controller": {
    "major": "1",
    "minor": "54",
    "gitVersion": "v1.54.2",
    "gitCommit": "",
    "gitTreeState": "",
    "buildDate": "",
    "goVersion": "",
    "compiler": "",
    "platform": "linux/amd64"
  }
}
```
### Provide your KubeOneCluster manifest here (if applicable)

```
apiVersion: kubeone.k8c.io/v1beta2
kind: KubeOneCluster
versions:
  kubernetes: '1.24.10'
cloudProvider:
  hetzner: {}
  external: true
clusterNetwork:
  kubeProxy:
    skipInstallation: true
  cni:
    cilium:
      enableHubble: true
      kubeProxyReplacement: strict
apiEndpoint:
  alternativeNames: ["<domain>"]
addons:
  enable: true
  addons:
  - name: default-storage-class
```

### What cloud provider are you running on?
Hetzner

### What operating system are you running in your cluster?
Ubuntu 20.04 Focal

### Additional information
We're failing to upgrade k8s to the latest 1.24.10 version, because `kubeadm` requires package `kubernetes-cni` v1.2.0. However, somehow the `kubernetes-cni` version is pinned to 1.1.1* :
```
[<ip>] + sudo DEBIAN_FRONTEND=noninteractive apt-get install --option Dpkg::Options::=--force-confold --no-install-recommends -y 'kubeadm=1.24.10*' 'kubernetes-cni=1.1.1*'
[<ip>] Reading package lists...
[<ip>] Building dependency tree...
[...]

PR #2628, opened 2023-01-27
[release/v1.4] Updates to kubernetes-cni, cri-tools, and supported Kubernetes versions

xmudrii:

The latest (January 2023) Kubernetes patch releases introduced some changes to packages that are described in #2607.

We were able to easily mitigate issues on main and `release/v1.5` because those branches/versions are using containerd 1.6.

However, we cannot do that easily for `release/v1.4` because it's using containerd 1.5 and cri-tools v1.26.0 doesn't support containerd 1.5. Updating containerd to 1.6 at this point would be too complicated because we have to update it in machine-controller, OSM, and KubeOne itself.

The idea for this PR is:

* We forbid using 1.23.16 and newer with KubeOne 1.4
  * We will revisit this if the issue gets solved on the Kubernetes side
* We strongly recommend upgrading to KubeOne 1.5 as soon as possible
* We'll upgrade kubernetes-cni to v1.2.0 because it's an easy win
* We'll upgrade and enforce cri-tools to v1.25.0
  * On CentOS and Amazon Linux 2, KubeOne will automatically downgrade cri-tools to v1.25.0 if newer is present
  * On Ubuntu, downgrading is possible with the `--force` flag

**What type of PR is this?**
/kind bug
/kind api-change

```release-note
Update kubernetes-cni to v1.2.0 and cri-tools to v1.25.0 to solve the issue with `crictl` not being able to interact with containerd. Set the maximum Kubernetes version to v1.23.15 because v1.23.16+ is requiring cri-tools v1.26.0 which doesn't work with containerd 1.5
```

**Documentation**:
```documentation
TBD
```

### DIFF
--- a/pkg/apis/kubeone/validation/validation.go
+++ b/pkg/apis/kubeone/validation/validation.go
@@ -38,7 +38,7 @@ const (
 	// lowerVersionConstraint defines a semver constraint that validates Kubernetes versions against a lower bound
 	lowerVersionConstraint = ">= 1.20"
 	// upperVersionConstraint defines a semver constraint that validates Kubernetes versions against an upper bound
-	upperVersionConstraint = "<= 1.23"
+	upperVersionConstraint = "<= 1.23.15"
 )
 
 var (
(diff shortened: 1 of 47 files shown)


ahmedwaleedmalik (approved):
/approve

dbagarozza:
we have this problem, but --force flag on ubuntu doesn't exist .

xmudrii:
We'll tackle this in another PR.

PR #2630, opened 2023-01-30
Add an option to disable templating for addons

ahmedwaleedmalik:

Introduce a new field in the addons spec `.addons.addon[x].DisableTemplate`. With this the manifests will be consumed as is without any templatization or addtions. This is useful when the addon manifests include Go text template tags i.e. {{ .value }}.

The main use case right now for this feature is better integration of custom operating system profiles and to avoid things like:

```
Make sure to escape all occurrences of Go text template in your custom OperatingSystemProfiles. For example, if you want have {{.Token}} in your OSP then it should be replaced with {{ `{{.Token}}` }}.
```

Fixes #

**What type of PR is this?**

/kind feature
/kind api-change

```release-note
Introduce a new `.Addons.Addons[x].DisableTemplating` field to the KubeOneCluster API which can be used to disable templatization for an addon.
```

**Documentation**:

```documentation
https://docs.kubermatic.com/kubeone/v1.6/architecture/operating-system-manager/usage/
```


pkg/apis/kubeone/v1beta2/types.go
```diff
@@ -882,6 +882,9 @@ type Addon struct {
 	// Params to the addon, to render the addon using text/template, this will override globalParams
 	Params map[string]string `json:"params,omitempty"`
 
+	// Disable templatization for the addon
```
xmudrii, line 885: Please make this a valid godoc comment.

pkg/apis/kubeone/v1beta2/types.go
```diff
@@ -882,6 +882,9 @@ type Addon struct {
...
+	DisableTemplate bool `json:"disableTemplate,omitempty"`
```
xmudrii, line 886: I'm wondering if we should call this with `-ing`, like `DisableTemplating`. I don't have strong opinions, but that might be a better name.

ahmedwaleedmalik: Good point, updated.

pkg/apis/kubeone/v1beta2/types.go
ahmedwaleedmalik: Good point, updated.

xmudrii (approved):
/lgtm
/approve

PR #2633, opened 2023-02-02
Hotfix: Fix decoding for bastion host public key

ahmedwaleedmalik:

**What this PR does / why we: need it**:
We unmarshal terraform output(tf.json) into [this](https://github.com/kubermatic/kubeone/blob/main/pkg/terraform/v1beta2/config.go#L35) Go struct. We use `[]byte` type for [BastionHostKey](https://github.com/kubermatic/kubeone/blob/main/pkg/terraform/v1beta2/config.go#L102) which will fail in most cases since during unmarshalling it will be interpreted as Base64 encoded data.

```
runtime: strict unmarshal of *v1beta2.Config
illegal base64 data at input byte 3
```

To fix this, we have updated the type from `[]byte` to `string`.

Fixes #

**What type of PR is this?**

/kind bug

This is a bug fix for an unreleased feature so we don't need any release notes.

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/docs/api_reference/v1beta2.en.md
+++ b/docs/api_reference/v1beta2.en.md
@@ -1,6 +1,6 @@
 +++
 title = "v1beta2 API Reference"
-date = 2023-01-30T19:59:26+05:00
+date = 2023-02-02T13:51:03+05:00
 weight = 11
 +++
 ## v1beta2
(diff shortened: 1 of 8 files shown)


xmudrii:
@ahmedwaleedmalik We also have `SSHHostPublicKey` which is `[]byte`. Can you please also refactor it to `string`?

ahmedwaleedmalik:
> Introducing new API fields HostConfig.SSHHostPublicKey and HostConfig.BastionHostPublicKey which take base64 encoded ssh public keys

https://github.com/kubermatic/kubeone/pull/2391

We will resolve it by updating documentation instead.

PR #2643, opened 2023-02-15
do not assign DynamicClient to state unless error is checked

WeirdMachine:

client.New returns a nil interface on error which will evaluate s.DynamicClient == nil to be false and causes nil pointer exceptions. 

Fixes  a nil pointer exception in case building dynamic kubernetes client failed on a previous try.
https://kubermatic.slack.com/archives/CF50UEULV/p1676476519084139

**What type of PR is this?**
/kind bug

```release-note
Fixes a nil pointer exception in case building dynamic kubernetes client failed on a previous try.
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/kubeconfig/kubeconfig.go
+++ b/pkg/kubeconfig/kubeconfig.go
@@ -69,10 +69,12 @@ func BuildKubernetesClientset(s *state.State) error {
 		return err
 	}
 
-	s.DynamicClient, err = client.New(s.RESTConfig, client.Options{})
+	dynamicClient, err := client.New(s.RESTConfig, client.Options{})
 	if err != nil {
 		return fail.KubeClient(err, "building dynamic kubernetes client")
 	}
+
+	s.DynamicClient = dynamicClient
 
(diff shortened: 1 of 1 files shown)


pkg/kubeconfig/kubeconfig.go
```diff
@@ -69,11 +69,13 @@ func BuildKubernetesClientset(s *state.State) error {
 		return err
 	}
 
-	s.DynamicClient, err = client.New(s.RESTConfig, client.Options{})
+	dynamicClient, err := client.New(s.RESTConfig, client.Options{})
```
kron4eg, line 72: This doesn't change anything, as if client.New() returns error it also returns nil client, meaning we assign nil to already nil (no actual change).

WeirdMachine: As described above, client.New() returns a nil interface. This looks something like:

`*client.client(nil)` instead of real `nil` and does not evaluate == nil as true

Issue #2611, opened 2023-01-26
Test cluster provisioning on Azure - Test Release KubeOne 1.6


xmudrii:
This is a ticket for tracking the progress of testing cluster provisioning on Azure. PR: #2621

### v1.26.1

* Containerd (OSM)
  * [x] Ubuntu: `/test pull-kubeone-e2e-azure-default-install-containerd-v1.26.1`
  * [x] CentOS: `/test pull-kubeone-e2e-azure-centos-install-containerd-v1.26.1`
  * [x] RHEL: `/test pull-kubeone-e2e-azure-rhel-install-containerd-v1.26.1`
  * [x] Rocky Linux: `/test pull-kubeone-e2e-azure-rockylinux-install-containerd-v1.26.1`
  * [x] Flatcar: `/test pull-kubeone-e2e-azure-flatcar-install-containerd-v1.26.1`
* Containerd (no OSM)
  * [x] Ubuntu: `/test pull-kubeone-e2e-azure-default-legacy-machine-controller-containerd-v1.26.1`
  * [x] CentOS: `/test pull-kubeone-e2e-azure-centos-legacy-machine-controller-containerd-v1.26.1`
  * [x] RHEL: `/test pull-kubeone-e2e-azure-rhel-legacy-machine-controller-containerd-v1.26.1`
  * [x] Rocky Linux: `/test pull-kubeone-e2e-azure-rockylinux-legacy-machine-controller-containerd-v1.26.1`
  * [x] Flatcar: `/test pull-kubeone-e2e-azure-flatcar-legacy-machine-controller-containerd-v1.26.1`
* External CCM (OSM)
  * [x] Ubuntu: `/test pull-kubeone-e2e-azure-default-install-containerd-external-v1.26.1`
  * [x] CentOS: `/test pull-kubeone-e2e-azure-centos-install-containerd-external-v1.26.1`
  * [x] RHEL: `/test pull-kubeone-e2e-azure-rhel-install-containerd-external-v1.26.1`
  * [x] Rocky Linux: `/test pull-kubeone-e2e-azure-rockylinux-install-containerd-external-v1.26.1`
  * [x] Flatcar: `/test pull-kubeone-e2e-azure-flatcar-install-containerd-external-v1.26.1`
* External CCM (no OSM)
  * [x] Ubuntu: `/test pull-kubeone-e2e-azure-default-legacy-machine-controller-containerd-external-v1.26.1`
[...]

PR #2647, opened 2023-02-18
azure: Fix CCM deployment and ssh reconnects

kron4eg:
While testing #2611 it was discovered that:
* azure CCM fails with "unknown --port flag", removing it fixes external CCM
* Flatcar control plane hosts like (for update) to reboot right in the middle of cluster initialization, which broke SSH client and left it in the "half-broken" state

Fixes #

**What type of PR is this?**
/kind bug

```release-note
* azure CCM: Fix unknown port flag in deployment
* fix ssh client to re-establish broken ssh connections
```

**Documentation**:

```documentation
NONE
```


examples/terraform/azure/variables.tf
```diff
@@ -270,3 +270,9 @@ variable "rhsm_offline_token" {
   type        = string
   sensitive   = true
 }
+
+variable "disable_auto_update" {
+  description = "Flatcar specific"
```
xmudrii, line 275: It's obvious from the name, but let's add a proper description so it renders well in docs.

kron4eg: Done

Issue #2705, opened 2023-02-27
KubeOne fails to download the GPG key from `packages.cloud.google.com`


dcardellino:
### What happened?

Upgrading Kubernetes from Version 1.25.6 to 1.26.1 fails with the following error:

```
WARN[12:39:14 CET] Task failed, error was: runtime: running task on "116.203.139.92"
ssh: upgrading kubeadm and CNI plugins
ssh: exec
ssh: popen
Process exited with status 2
set -xeuo pipefail
...
...
...
W: https://download.docker.com/linux/ubuntu/dists/jammy/InRelease: Key is stored in legacy trusted.gpg keyring (/etc/apt/trusted.gpg), see the DEPRECATION section in apt-key(8) for details.
W: http://apt.kubernetes.io/dists/kubernetes-xenial/InRelease: Key is stored in legacy trusted.gpg keyring (/etc/apt/trusted.gpg), see the DEPRECATION section in apt-key(8) for details.
+ sudo DEBIAN_FRONTEND=noninteractive apt-get install --option Dpkg::Options::=--force-confold -y --no-install-recommends apt-transport-https ca-certificates curl gnupg lsb-release rsync
+ curl -fsSL https://packages.cloud.google.com/apt/doc/apt-key.gpg
+ sudo apt-key add -
Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8)).
curl: (22) The requested URL returned error: 500
gpg: no valid OpenPGP data found.
```

Running the command `curl -fsSL https://packages.cloud.google.com/apt/doc/apt-key.gpg` is the one which is causes the actual error.

### Expected behavior

KubeOne can upgrade Kubernetes.

### How to reproduce the issue?

See in What happened?

### What KubeOne version are you using?

<details>

```console
$ kubeone version
{
  "kubeone": {
    "major": "1",
    "minor": "6",
    "gitVersion": "1.6.0",
    "gitCommit": "8b0973ca77856dca920798bbd5ff6b5c0f3f4856",
    "gitTreeState": "",
    "buildDate": "2023-02-23T19:25:26Z",
    "goVersion": "go1.19.6",
    "compiler": "gc",
    "platform": "linux/amd64"
  },
  "machine_controller": {
    "major": "1",
    "minor": "56",
    "gitVersion": "v1.56.0",
    "gitCommit": "",
    "gitTreeState": "",
    "buildDate": "",
    "goVersion": "",
    "compiler": "",
[...]

PR #2706, opened 2023-02-27
Use apt/GPG key from `dl.k8s.io`

xmudrii:

The purpose of this PR is to start downloading GPG key for the APT repo from `dl.k8s.io` instead of `packages.cloud.google.com`.

As we got to see in #2705, it might happen that `packages.cloud.google.com` is unavailable. Kubernetes doesn't have any control over that endpoint, so all we can do is wait for the Google folks to fix it. Sometimes that might take time, so the upstream decided to mitigate that issue by hosting the GPG key themselves. The new `dl.k8s.io` endpoint works in the following way:

- usually, it's going to be just a redirect to `packages.cloud.google.com`
- if `packages.cloud.google.com` is down for a longer period, the upstream folks are going to point `dl.k8s.io` to some mirror (e.g. GitHub)

Such setup will minimize the down time as much as possible.

Ref: https://github.com/kubernetes/k8s.io/pull/4837

xref #2705 

**What type of PR is this?**
/kind regression

```release-note
Download GPG key for the `apt` repository from `dl.k8s.io` instead of `packages.cloud.google.com`
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/pkg/scripts/os_debian.go
+++ b/pkg/scripts/os_debian.go
@@ -61,7 +61,7 @@ sudo systemctl enable --now iscsid
 {{- end }}
 
 {{- if .CONFIGURE_REPOSITORIES }}
-curl -fsSL https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
+curl -fsSL https://dl.k8s.io/apt/doc/apt-key.gpg | sudo apt-key add -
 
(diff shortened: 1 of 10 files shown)


xmudrii:
/hold
for testing

xmudrii:
Let's merge this one on main to get the signal how it performs. We can decide later if we need to backport it (it's not a priority any longer because the original issue has been fixed on Google's side).
/hold cancel

PR #2751, opened 2023-04-03
API token authentication support for VMware Cloud Director

ahmedwaleedmalik:
This PR adds support for API token authentication for VMware Cloud Director.

Required for https://github.com/kubermatic/kubermatic/issues/11624

**What type of PR is this?**

/kind feature

```release-note
VMware Cloud Director now supports authentication using API Token
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/examples/terraform/vmware-cloud-director/README.md
+++ b/examples/terraform/vmware-cloud-director/README.md
@@ -20,27 +20,34 @@ Following environment variables or terraform variables can be used to authentica
 | Environment Variable | Terraform Variable |
 |------|---------|
 | VCD_USER | vcd.user |
-| VCD_PASSWORD | vcd.user |
+| VCD_PASSWORD | vcd.password |
 | VCD_ORG | vcd.org |
 | VCD_URL | vcd.url |
+
+To use access token instead of username and password:
+
+
+| Environment Variable | Terraform Variable |
+|------|---------|
+| VCD_AUTH_TYPE | vcd.auth_type |
+| VCD_API_TOKEN | vcd.api_token |
 
 #### References
 
 - <https://registry.terraform.io/providers/vmware/vcd/latest/docs#connecting-as-sys-admin-with-default-org-and-vdc>
 - <https://registry.terraform.io/providers/vmware/vcd/latest/docs#argument-reference>
-
 ## Requirements
 
 | Name | Version |
 |------|---------|
 | <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 1.0.0 |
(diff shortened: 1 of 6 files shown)


ahmedwaleedmalik:
Still WIP and blocked by https://github.com/kubermatic/machine-controller/pull/1612

kron4eg:
Looks good. But I think you also need to update the MC image version to reference the latest commit.

pkg/credentials/credentials.go
```diff
@@ -598,5 +599,61 @@ func vmwareCloudDirectorValidationFunc(creds map[string]string) error {
...
+		}
+	}
+
+	// Username and password are required when using default credentials i.e. username and password.
+	if !(username && password) {
+		return fail.CredentialsError{
+			Op:       "validating",
+			Provider: "VMware Cloud Director",
+			Err: errors.Errorf(
+				"key %v or %v is required but isn't present",
```
xmudrii, line 651: Nit:
```suggestion
				"key %v and %v are required but not present",
```

examples/terraform/vmware-cloud-director/README.md
```diff
@@ -20,27 +20,34 @@ Following environment variables or terraform variables can be used to authentica
...
 | VCD_ORG | vcd.org |
 | VCD_URL | vcd.url |
 
+To use access token instead of username and password:
+
+
+| Environment Variable | Terraform Variable |
+|------|---------|
+| VCD_AUTH_TYPE | vcd.auth_type |
+| VCD_API_TOKEN | vcd.api_token |
```
xmudrii, line 33: One thing that confuses me is that:
- we call this API token in Terraform and KubeOne
- it's called refresh token in the CSI config

What's the official name out of those two?

xmudrii (commented):
Just two nits.

pkg/credentials/credentials.go
ahmedwaleedmalik: I think the original one is fine.

xmudrii: The original one implies that you need only one (`or`), but my understanding is that we need both, right?

examples/terraform/vmware-cloud-director/README.md
xmudrii: Okay, that makes sense. Thanks!

pkg/credentials/credentials.go
ahmedwaleedmalik: I have committed whatever your suggestion was, should be good now.

xmudrii (approved):
/approve

PR #2763, opened 2023-04-12
Fix kubeone setup on Hetzner

ThWoywod:

Fixes Control-Plane setup issue. 
Sometimes the following issue ocure:
```
[10.10.0.6] Reading package lists...
[10.10.0.6] E: Could not get lock /var/lib/apt/lists/lock. It is held by process 1125 (apt-get)
[10.10.0.6] E: Unable to lock directory /var/lib/apt/lists/
time="11:16:15 UTC" level=error msg="ssh: installing kubeadm\nssh: popen\nProcess exited with status 100" 
```

This happen because any `apt-get update` process is still running in the background and the process is frozen. Even during a retry the `apt-get update` process is on hold.

**What type of PR is this?**
/kind bug

### DIFF
--- a/pkg/scripts/os_debian.go
+++ b/pkg/scripts/os_debian.go
@@ -43,7 +43,7 @@ Acquire::http::Proxy "{{ .HTTP_PROXY }}";
 {{- end }}
 EOF
 
-update_pid="$(lsof | grep /var/lib/apt/lists/lock || true)"
+update_pid="$(lsof | grep ' /var/lib/apt/lists/lock' || true)"
 if [ ! -z "$update_pid" ]; then
 	# make sure that no other update process is running in the background
 	lsof -t /var/lib/apt/lists/lock | xargs -r sudo kill
@@ -91,7 +91,7 @@ fi
 # contains neither kubeadm nor kubelet, and the docs themselves suggest using xenial repo.
 echo "deb http://apt.kubernetes.io/ kubernetes-xenial main" | sudo tee /etc/apt/sources.list.d/kubernetes.list
 
-update_pid="$(lsof | grep /var/lib/apt/lists/lock || true)"
+update_pid="$(lsof | grep ' /var/lib/apt/lists/lock' || true)"
 if [ ! -z "$update_pid" ]; then
 	# make sure that no other update process is running in the background
 	lsof -t /var/lib/apt/lists/lock | xargs -r sudo kill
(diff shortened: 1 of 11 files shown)


embik:
Hey @ThWoywod, thank you for your submission. I'll let others voice their opinion on this but I find this change a bit scary - how can we make sure this is a "frozen" apt process we're killing and not a legitimate one? What does "frozen" even mean in the first place? `apt` puts these locks in place because messing with packages while `apt` is working can be highly destructive.

I think it would be more suited to figure out why apt processes freeze for you instead of this workaround, but again, that's a personal opinion and not necessarily that of other KubeOne folks.

kron4eg:
Sometimes unattended-upgrades gets too fast & furious, and run in your way. I'm also against killing apt in general. That problem should not be solved at kubeone level.

ThWoywod:
Hey @embik,
thank you for your feedback. I totaly agree with you that this is quite a hacky way around. 
I invest some time to figure out why this only sometime happen but did not really get it. 

For me it is not really clear why there should be an update process running in the background. But the fact is that it does. My guess is that this happens when an installation error occurs and the retry is executed. The retry does not make sure that the processes are really terminated before it will start. 
Maybe there is also the right point to make sure that the update process has been terminated. 

I am open for any feedback and would like to adjust the code to fit better to the kubeoen way.

ThWoywod:
@kron4eg you mean this is an issue with the default Hetzner `unattended-upgrades` configuration?

I mean the only unattended-upgrades stuff that is configured during the kubeone installation process is the Kubernetes DeamonSet and the setup of this Kubernetes Ressources should be done after the setup of the controle-plane nodes.

kron4eg:
Yeah, hetzner integrated unattended-upgrades in the base ubuntu image and sometimes it kicks in right after the VM boot (and before kubeone).

ThWoywod:
@kron4eg How about we stop the process of unattended-upgrades before the kubernetes installation steps are executed?

kron4eg:
@ThWoywod well, this can be done in a custom OSP. See more at https://docs.kubermatic.com/operatingsystemmanager/guides/customize-operating-system-profile/

xmudrii:
@ThWoywod Is it an option for you to provide cloud-init script to disable or reconfigure unattended-upgrades? You can do that [using Terraform](https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs/resources/server#user_data) as well.

ThWoywod:
Yeah i gess the OSP are not really the way to go because its an issue with the controle plane nodes not the machine deployment nodes.

Well yes @xmudrii this is a good suggestion. I will experiment with it a bit but at the end the kubeone code need to have an change. Disableing unattended-upgrades in generall does not sound practicable for me.

I will get back as soon as I have found a solution to the problem.

PR #2776, opened 2023-04-26
support for Oracle Linux

jserviceorg:
Fixes #

This adds Oracle Linux support for on premise installations. Oracle Linux is basically just another RHEL clone like Rocky.

**What type of PR is this?**
/kind feature


kron4eg:
@jserviceorg thank you for your efforts!

Before we can proceed on this, please take a look at the https://github.com/kubermatic/kubeone/blob/main/docs/adding_provider_support.md, especially Prerequisites section. Please keep in mind, this doc has been written prior we even had an idea about OSM that we are currently using. I mean, besides integrating to machine-controller you will also need an OperatingSystemProfile object integrated in OSM project.

Once MC and OSM has support for new OS, then and only then we can integrate it to Kubeone.

PR #2784, opened 2023-05-11
Disable CSI driver deploy by API flag

### LINKED ISSUE #2750: Possibility to deactivate vSphere CSI

As a KubeOne user
I would like to have the ability to deactivate the vSphere CSI
in order to be able to deploy a cluster without vSphere CSI

### Solution details

- disable vSphere CSI
- deploy cluster without putting a `csiConfig` to the `KubeOneCluster` configuration

### Use cases

I want to use the vSphere provider but without the vSphere CSI because I want to use Ceph as a storage backend.
The problem is if I only put the following lines of code into the `KubeOneCluster` config is that the pods of the `vsphere-csi-controller` deployment run into a `CrashLoopBackOff`.

```yaml
cloudProvider:
  vsphere: {}
  external: true
  csiConfig: |
    [VirtualCenter "<VCENTER>"]
cloudConfig: |
<omitted for brevity>
```

### Additional information

You can also check https://github.com/kubermatic/kubeone/issues/2737 for some background information.

kron4eg:

Fixes #2750

**What type of PR is this?**
/kind feature

```release-note
Disable CSI driver deploy by API flag
```

**Documentation**:
```documentation
TBD
```

### DIFF
--- a/pkg/apis/kubeone/types.go
+++ b/pkg/apis/kubeone/types.go
@@ -322,8 +322,8 @@ type CloudProviderSpec struct {
 	// External
 	External bool `json:"external,omitempty"`
 
-	// DisableBundledCSIDriver used to cancel automatic deploy of bundled CSI drivers
-	DisableBundledCSIDriver bool `json:"disableBundledCSIDriver"`
+	// DisableBundledCSIDrivers used to cancel automatic deploy of bundled CSI drivers
+	DisableBundledCSIDrivers bool `json:"disableBundledCSIDrivers"`
 
 	// CloudConfig
 	CloudConfig string `json:"cloudConfig,omitempty"`
--- a/pkg/apis/kubeone/types.go
+++ b/pkg/apis/kubeone/types.go
@@ -322,6 +322,9 @@ type CloudProviderSpec struct {
 	// External
 	External bool `json:"external,omitempty"`
 
+	// DisableBundledCSIDriver used to cancel automatic deploy of bundled CSI drivers
+	DisableBundledCSIDriver bool `json:"disableBundledCSIDriver"`
+
 	// CloudConfig
 	CloudConfig string `json:"cloudConfig,omitempty"`
 
--- a/pkg/apis/kubeone/types.go
+++ b/pkg/apis/kubeone/types.go
@@ -322,7 +322,7 @@ type CloudProviderSpec struct {
 	// External
 	External bool `json:"external,omitempty"`
 
-	// DisableBundledCSIDrivers used to cancel automatic deploy of bundled CSI drivers
+	// DisableBundledCSIDrivers disables automatic deployment of CSI drivers bundled with KubeOne
 	DisableBundledCSIDrivers bool `json:"disableBundledCSIDrivers"`
 
 	// CloudConfig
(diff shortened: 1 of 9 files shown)


pkg/apis/kubeone/v1beta2/types.go
```diff
@@ -319,6 +319,9 @@ type CloudProviderSpec struct {
 	// External
 	External bool `json:"external,omitempty"`
 
+	// DisableBundledCSIDriver used to cancel automatic deploy of bundled CSI drivers
+	DisableBundledCSIDriver bool `json:"disableBundledCSIDriver"`
```
xmudrii, line 323: I kinda prefer "positive" fields, I think it's more in line with our API. It would be similar to `external` which you need to set to `true` to enable, and with many other flags which are just `enable`.

That said, maybe we can name this `DeployBundledCSIDriver` or even just `DeployCSIDriver`, make it a bool pointer, and default it to true if it's not explicitly set.

kron4eg: Having "negative" fields very nicely exploits zero value of Go's type system, avoids any potential NPE with pointers, avoids additional logic to initialized the field with reverse logic. And is literally what a user wants ("disable deployment of CSI drivers").

pkg/apis/kubeone/types.go
```diff
@@ -322,6 +322,9 @@ type CloudProviderSpec struct {
 	// External
 	External bool `json:"external,omitempty"`
 
+	// DisableBundledCSIDriver used to cancel automatic deploy of bundled CSI drivers
```
xmudrii, line 325: Nit: potential improvements
```suggestion
	// DisableBundledCSIDriver disables automatic deployment of CSI drivers bundled with KubeOne
```

pkg/apis/kubeone/v1beta2/types.go
xmudrii: Just a tiny nit on naming: should it be DisableBundledCSIDriver or DisableBundledCSIDrivers (in plural), because some providers might have multiple CSI drivers.

pkg/apis/kubeone/validation/validation.go
```diff
@@ -269,7 +271,7 @@ func ValidateCloudProviderSpec(p kubeoneapi.CloudProviderSpec, networkConfig kub
 		allErrs = append(allErrs, field.Invalid(fldPath, "", "provider must be specified"))
 	}
 
-	if p.Vsphere == nil && len(p.CSIConfig) > 0 {
+	if providerSpec.Vsphere == nil && len(providerSpec.CSIConfig) > 0 {
```
xmudrii, line 274: I think we should also reject CSIConfig if CSI drivers are disabled.

pkg/apis/kubeone/v1beta2/types.go
kron4eg: > Just a tiny nit on naming: should it be DisableBundledCSIDriver or DisableBundledCSIDrivers (in plural), because some providers might have multiple CSI drivers.

done

pkg/apis/kubeone/types.go
kron4eg: done

pkg/apis/kubeone/validation/validation.go
kron4eg: done

xmudrii (approved):
/lgtm
/approve

ahmedwaleedmalik (approved):
/approve
/retest

PR #2787, opened 2023-05-16
Explicitly configure TLS cert path for machine-controller webhooks

ahmedwaleedmalik:
With https://github.com/kubermatic/machine-controller/pull/1644, the default values for machine-controller webhook flags `tls-cert-path` and `tls-key-path` were changed to `/tmp/cert/tls.crt` and `/tmp/cert/tls.key` respectively. To make sure it's still compatible with KubeOne we need to configure explicitly pass these flags with the values that KubeOne generates.

Fixes #

**What type of PR is this?**

/kind chore

An alternative to this is to rename the keys in the `machinecontroller-webhook-serving-cert` secret.

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/addons/machinecontroller/webhook.yaml
+++ b/addons/machinecontroller/webhook.yaml
@@ -87,6 +87,8 @@ spec:
             - -v=4
             - -listen-address=0.0.0.0:9876
             - -namespace=kube-system
+            - -tls-cert-path=/tmp/cert/cert.pem
+            - -tls-key-path=/tmp/cert/tls.key
             {{ if .Config.CloudProvider.External }}
             - -node-external-cloud-provider
             {{ end }}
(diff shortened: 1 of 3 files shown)


addons/machinecontroller/webhook.yaml
```diff
@@ -87,6 +87,8 @@ spec:
             - -v=4
             - -listen-address=0.0.0.0:9876
             - -namespace=kube-system
+            - -tls-cert-path=/tmp/cert/cert.pem
+            - -tls-key-path=/tmp/cert/tls.key
```
xmudrii, line 91: Isn’t it key.pem in the Secret?

xmudrii (approved):

PR #2790, opened 2023-05-18
[EXPERIMENTAL] Add experimental support for OpenBuildService

xmudrii:

This is very experimental and unsupported, do not use.

**What type of PR is this?**
/kind cleanup

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/pkg/scripts/os_amzn.go
+++ b/pkg/scripts/os_amzn.go
@@ -49,9 +49,9 @@ cat <<EOF | sudo tee /etc/yum.repos.d/kubernetes.repo
 [kubernetes]
 name=Kubernetes
 type=rpm-md
-baseurl=https://download.opensuse.org/repositories/isv:/kubernetes:/core:/stable:/v{{ .OBS_VERSION }}:/rpm_stream/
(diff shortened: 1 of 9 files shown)


pkg/scripts/os_amzn.go
```diff
@@ -214,6 +226,8 @@ func KubeadmAmazonLinux(cluster *kubeoneapi.KubeOneCluster, force bool) (string,
 		"INSTALL_CONTAINERD":     cluster.ContainerRuntime.Containerd,
 		"USE_KUBERNETES_REPO":    cluster.AssetConfiguration.NodeBinaries.URL == "",
 		"CILIUM":                 ciliumCNI(cluster),
+		"USE_OBS":                kubeoneapi.IsOpenBuildServiceEnabled(),
+		"OBS_VERSION":            cluster.Versions.KubernetesMajorMinorVersion(),
```
kron4eg, line 230: Theoretically we can get away with just 1 variable instead of 2, just "OBS_VERSION", and in template use it like this
```
{{ with .OBS_VERSION }}
  this block will be omitted if .OBS_VERSION above is empty. .OBS_VERSION from above here can be used as {{ . }}.
{{ end }}
```

PR #2791, opened 2023-05-19
bug: remove duplicate err print

boranx:

Maybe a small thing, but I came across this while I was playing with kubermatic stack and kubeone
This PR removes the second error print as it's already being handled by cobra

before:
```shell
kubeone --version
Error: unknown flag: --version
Error: unknown flag: --version
```
after:
```shell
./dist/kubeone --version
Error: unknown flag: --version
```

```release-note
Fix duplicate CLI print
```

```documentation
NONE
```

### DIFF
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -51,11 +51,7 @@ func Execute() {
 	}
 
 	rootCmd := newRoot()
-
 	if err := rootCmd.Execute(); err != nil {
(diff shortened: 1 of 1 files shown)


kron4eg:
/release-note-edit ```release-note Fix duplicate CLI print```

kron4eg:
@boranx thanks for contribution!

Issue #2792, opened 2023-05-23
Control Plane Node CIS issues


kron4eg:
### What happened?

Running [CIS benchmark](https://github.com/aquasecurity/kube-bench) results to failures:

```
[INFO] 1 Control Plane Security Configuration
[INFO] 1.1 Control Plane Node Configuration Files
[WARN] 1.1.9 Ensure that the Container Network Interface file permissions are set to 600 or more restrictive (Manual)
[WARN] 1.1.10 Ensure that the Container Network Interface file ownership is set to root:root (Manual)
[FAIL] 1.1.12 Ensure that the etcd data directory ownership is set to etcd:etcd (Automated)
[WARN] 1.1.20 Ensure that the Kubernetes PKI certificate file permissions are set to 600 or more restrictive (Manual)
[INFO] 1.2 API Server
[WARN] 1.2.1 Ensure that the --anonymous-auth argument is set to false (Manual)
[FAIL] 1.2.5 Ensure that the --kubelet-certificate-authority argument is set as appropriate (Automated)
[WARN] 1.2.9 Ensure that the admission control plugin EventRateLimit is set (Manual)
[WARN] 1.2.11 Ensure that the admission control plugin AlwaysPullImages is set (Manual)
[WARN] 1.2.12 Ensure that the admission control plugin SecurityContextDeny is set if PodSecurityPolicy is not used (Manual)
[FAIL] 1.2.15 Ensure that the admission control plugin NodeRestriction is set (Automated)
[FAIL] 1.2.17 Ensure that the --profiling argument is set to false (Automated)
[FAIL] 1.2.18 Ensure that the --audit-log-path argument is set (Automated)
[FAIL] 1.2.19 Ensure that the --audit-log-maxage argument is set to 30 or as appropriate (Automated)
[FAIL] 1.2.20 Ensure that the --audit-log-maxbackup argument is set to 10 or as appropriate (Automated)
[FAIL] 1.2.21 Ensure that the --audit-log-maxsize argument is set to 100 or as appropriate (Automated)
[WARN] 1.2.22 Ensure that the --request-timeout argument is set as appropriate (Manual)
[WARN] 1.2.29 Ensure that the --encryption-provider-config argument is set as appropriate (Manual)
[WARN] 1.2.30 Ensure that encryption providers are appropriately configured (Manual)
[WARN] 1.2.31 Ensure that the API Server only makes use of Strong Cryptographic Ciphers (Manual)
[...]

PR #2797, opened 2023-05-26
Fixed some CIS issues on control plane nodes

kron4eg:

Following reported CIS fails and warnings been fixed:
* [WARN] 1.1.20 Ensure that the Kubernetes PKI certificate file permissions are set to 600 or more restrictive (Manual)
* [FAIL] 1.2.5 Ensure that the --kubelet-certificate-authority argument is set as appropriate (Automated)
* [FAIL] 1.2.17 Ensure that the --profiling argument is set to false (Automated)
* [WARN] 1.2.22 Ensure that the --request-timeout argument is set as appropriate (Manual)
* [WARN] 1.2.31 Ensure that the API Server only makes use of Strong Cryptographic Ciphers (Manual)
* [WARN] 1.3.1 Ensure that the --terminated-pod-gc-threshold argument is set as appropriate (Manual)
* [FAIL] 1.3.2 Ensure that the --profiling argument is set to false (Automated)
* [FAIL] 1.4.1 Ensure that the --profiling argument is set to false (Automated)

xref #2792

**What type of PR is this?**
/kind bug

```release-note
Fixed some CIS issues on control plane nodes
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/scripts/kubeadm.go
+++ b/pkg/scripts/kubeadm.go
@@ -28,6 +28,7 @@ var (
 
 		sudo kubeadm {{ .VERBOSE }} join \
 			--config={{ .WORK_DIR }}/cfg/master_{{ .NODE_ID }}.yaml
+		sudo find /etc/kubernetes/pki/ -name *.crt -exec chown 600 {} \;
 	`)
 
 	kubeadmWorkerJoinScriptTemplate = heredoc.Doc(`
@@ -35,11 +36,13 @@ var (
 
 		sudo kubeadm {{ .VERBOSE }} join \
 			--config={{ .WORK_DIR }}/cfg/worker_{{ .NODE_ID }}.yaml
+		sudo find /etc/kubernetes/pki/ -name *.crt -exec chown 600 {} \;
 	`)
 
 	kubeadmCertScriptTemplate = heredoc.Doc(`
 		sudo kubeadm {{ .VERBOSE }} init phase certs all \
 			--config={{ .WORK_DIR }}/cfg/master_{{ .NODE_ID }}.yaml
+		sudo find /etc/kubernetes/pki/ -name *.crt -exec chown 600 {} \;
 	`)
 
 	kubeadmInitScriptTemplate = heredoc.Doc(`
@@ -49,6 +52,7 @@ var (
 		fi
 
 		sudo kubeadm {{ .VERBOSE }} init {{ if .SKIP_PHASE }}--skip-phases={{ .SKIP_PHASE }} {{ end}}--config={{ .WORK_DIR }}/cfg/master_{{ .NODE_ID }}.yaml
+		sudo find /etc/kubernetes/pki/ -name *.crt -exec chown 600 {} \;
 	`)
 
 	kubeadmResetScriptTemplate = heredoc.Doc(`
@@ -63,6 +67,7 @@ var (
 
 	kubeadmUpgradeScriptTemplate = heredoc.Doc(`
 		sudo {{ .KUBEADM_UPGRADE }}{{ if .LEADER }} --config={{ .WORK_DIR }}/cfg/master_{{ .NODE_ID }}.yaml{{ end }}
+		sudo find /etc/kubernetes/pki/ -name *.crt -exec chown 600 {} \;
 	`)
 
 	kubeadmPauseImageVersionScriptTemplate = heredoc.Doc(`
--- a/pkg/templates/kubeadm/v1beta3/kubeadm.go
+++ b/pkg/templates/kubeadm/v1beta3/kubeadm.go
@@ -17,6 +17,7 @@ limitations under the License.
 package v1beta3
 
 import (
+	"crypto/tls"
 	"fmt"
 	"path/filepath"
 	"strconv"
@@ -177,9 +178,13 @@ func NewConfig(s *state.State, host kubeoneapi.HostConfig) ([]runtime.Object, er
 		APIServer: kubeadmv1beta3.APIServer{
 			ControlPlaneComponent: kubeadmv1beta3.ControlPlaneComponent{
 				ExtraArgs: map[string]string{
-					"endpoint-reconciler-type": "lease",
-					"service-node-port-range":  cluster.ClusterNetwork.NodePortRange,
-					"enable-admission-plugins": kubeflags.DefaultAdmissionControllers(kubeSemVer),
+					"enable-admission-plugins":      kubeflags.DefaultAdmissionControllers(kubeSemVer),
+					"endpoint-reconciler-type":      "lease",
+					"kubelet-certificate-authority": "/etc/kubernetes/pki/ca.crt",
+					"profiling":                     "false",
+					"request-timeout":               "300s",
+					"service-node-port-range":       cluster.ClusterNetwork.NodePortRange,
+					"tls-cipher-suites":             safeTLSCiphers(),
 				},
 				ExtraVolumes: []kubeadmv1beta3.HostPathMount{},
 			},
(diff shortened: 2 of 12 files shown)


pkg/templates/kubeadm/v1beta3/kubeadm.go
```diff
@@ -177,17 +178,29 @@ func NewConfig(s *state.State, host kubeoneapi.HostConfig) ([]runtime.Object, er
...
 			ControlPlaneComponent: kubeadmv1beta3.ControlPlaneComponent{
 				ExtraArgs: map[string]string{
-					"endpoint-reconciler-type": "lease",
-					"service-node-port-range":  cluster.ClusterNetwork.NodePortRange,
-					"enable-admission-plugins": kubeflags.DefaultAdmissionControllers(kubeSemVer),
+					"enable-admission-plugins":      kubeflags.DefaultAdmissionControllers(kubeSemVer),
+					"endpoint-reconciler-type":      "lease",
+					"kubelet-certificate-authority": "/etc/kubernetes/pki/ca.crt",
+					"profiling":                     "false",
+					"request-timeout":               "300s",
```
xmudrii, line 185: Why is this 300s when the [upstream default](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/) is 1m?

> --request-timeout duration     Default: 1m0s
>
> An optional field indicating the duration a handler must keep a request open before timing it out. This is the default request timeout for requests but may be overridden by flags such as --min-request-timeout for specific types of requests.

pkg/templates/kubeadm/v1beta3/kubeadm.go
```diff
@@ -177,17 +178,29 @@ func NewConfig(s *state.State, host kubeoneapi.HostConfig) ([]runtime.Object, er
...
 				ExtraVolumes: []kubeadmv1beta3.HostPathMount{},
 			},
 			CertSANs: certSANS,
 		},
 		ControllerManager: kubeadmv1beta3.ControlPlaneComponent{
 			ExtraArgs: map[string]string{
-				"flex-volume-plugin-dir": "/var/lib/kubelet/volumeplugins",
+				"flex-volume-plugin-dir":      "/var/lib/kubelet/volumeplugins",
+				"profiling":                   "false",
+				"terminated-pod-gc-threshold": "100",
```
xmudrii, line 197: I'm kinda confused with this parameter. The [upstream default value](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/) is 12500 but it seems ridiculously high. How did we come up with 100?

>  --terminated-pod-gc-threshold int32     Default: 12500
>
> Number of terminated pods that can exist before the terminated pod garbage collector starts deleting terminated pods. If <= 0, the terminated pod garbage collector is disabled.

pkg/templates/kubeadm/v1beta3/kubeadm.go
```diff
@@ -177,17 +178,29 @@ func NewConfig(s *state.State, host kubeoneapi.HostConfig) ([]runtime.Object, er
...
+					"profiling":                     "false",
+					"request-timeout":               "300s",
+					"service-node-port-range":       cluster.ClusterNetwork.NodePortRange,
+					"tls-cipher-suites":             safeTLSCiphers(),
 				},
 				ExtraVolumes: []kubeadmv1beta3.HostPathMount{},
 			},
 			CertSANs: certSANS,
 		},
 		ControllerManager: kubeadmv1beta3.ControlPlaneComponent{
```
xmudrii, line 193: Why don't we set `tls-cipher-suites` for kube-controller-manager and kube-scheduler?

pkg/scripts/kubeadm.go
```diff
@@ -28,18 +28,21 @@ var (
 
 		sudo kubeadm {{ .VERBOSE }} join \
 			--config={{ .WORK_DIR }}/cfg/master_{{ .NODE_ID }}.yaml
+		sudo find /etc/kubernetes/pki/ -name *.crt -exec chown 600 {} \;
```
xmudrii, line 31: This should be `chmod` instead of `chown`.

```suggestion
		sudo find /etc/kubernetes/pki/ -name *.crt -exec chmod 600 {} \;
```

xmudrii (changes requested):

pkg/templates/kubeadm/v1beta3/kubeadm.go
kron4eg: Report says:
```
[WARN] 1.2.22 Ensure that the --request-timeout argument is set as appropriate (Manual)

1.2.22 Edit the API server pod specification file /etc/kubernetes/manifests/kube-apiserver.yaml
and set the below parameter as appropriate and if needed.
For example, --request-timeout=300s
```

pkg/templates/kubeadm/v1beta3/kubeadm.go
kron4eg: Report
```
[WARN] 1.3.1 Ensure that the --terminated-pod-gc-threshold argument is set as appropriate (Manual)

1.3.1 Edit the Controller Manager pod specification file /etc/kubernetes/manifests/kube-controller-manager.yaml
on the control plane node and set the --terminated-pod-gc-threshold to an appropriate threshold,
for example, --terminated-pod-gc-threshold=10
```

But 10, I thought it, is too small. So I just took a sensible guess.

pkg/templates/kubeadm/v1beta3/kubeadm.go
kron4eg: Report didn't have any problems about `--tls-cipher-suites` them, only kube-apiserver :shrug:

pkg/scripts/kubeadm.go
kron4eg: :+1: fixed

pkg/templates/kubeadm/v1beta3/kubeadm.go
xmudrii: My understanding of that report is that they provided an example how to set the flag. I think we should use same values as used by default by API server.

pkg/templates/kubeadm/v1beta3/kubeadm.go
xmudrii: Same as above, my understanding of that report is that they provided an example how to set the flag. I agree that 10 is too small, but I think we should use the same value as kube-controller-manager by default.

pkg/templates/kubeadm/v1beta3/kubeadm.go
xmudrii: I double checked in KKP and we also set this flag only for kube-apiserver.

pkg/templates/kubeadm/v1beta3/kubeadm.go
kron4eg: I don't get it, "ridiculously high" and "use the default" are mutually exclusive. We have to pick one. Default: 12500 means that 12500 pod specs could potentially litter etcd storage for no good reason. Let's pick something smaller, 500? 1000? whatever, but not 12500 :exploding_head:

kron4eg: Let's take 20k per pod, that's very roughly 230Mb (12500 * 20kb) of excess bytes in etcd.

kron4eg: Updated to 1000.

pkg/templates/kubeadm/v1beta3/kubeadm.go
kron4eg: done

PR #2799, opened 2023-06-05
Update canal and cilium cni

WeirdMachine:

Fixes #

**What type of PR is this?**

```release-note
Update canal to v3.26.0 and cilium to v1.13.3
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/addons/cni-canal/canal.yaml
+++ b/addons/cni-canal/canal.yaml
@@ -1,4 +1,41 @@
-# Source: https://projectcalico.docs.tigera.io/archive/v3.23/manifests/canal.yaml
+# Source: https://github.com/projectcalico/calico/blob/v3.26.0/manifests/canal.yaml
+---
+# Source: calico/templates/calico-kube-controllers.yaml
+# This manifest creates a Pod Disruption Budget for Controller to allow K8s Cluster Autoscaler to evict
+
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: calico-kube-controllers
+  namespace: kube-system
+  labels:
+    k8s-app: calico-kube-controllers
+spec:
+  maxUnavailable: 1
+  selector:
+    matchLabels:
+      k8s-app: calico-kube-controllers
+---
+# Source: calico/templates/calico-kube-controllers.yaml
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: calico-kube-controllers
+  namespace: kube-system
+---
+# Source: calico/templates/calico-node.yaml
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: canal
+  namespace: kube-system
+---
+# Source: calico/templates/calico-node.yaml
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: calico-cni-plugin
+  namespace: kube-system
 ---
 # Source: calico/templates/calico-config.yaml
 # This ConfigMap is used to configure a self-hosted Canal installation.
@@ -41,18 +78,7 @@ data:
           "mtu": __CNI_MTU__,
           "ipam": {
               "type": "host-local",
-              {{ if and .Config.ClusterNetwork.HasIPv4 .Config.ClusterNetwork.HasIPv6 }}
-              "ranges": [
-                  [{"subnet": "usePodCidr"}],
-                  [{"subnet": "usePodCidrIPv6"}]
-              ]
-              {{ else }}
-              {{ if .Config.ClusterNetwork.HasIPv6 }}
-              "subnet": "usePodCidrIPv6"
-              {{ else }}
               "subnet": "usePodCidr"
-              {{ end }}
-              {{ end }}
           },
           "policy": {
               "type": "k8s"
@@ -87,10 +113,8 @@ data:
         "Type": "vxlan"
       }
     }
-
 ---
 # Source: calico/templates/kdd-crds.yaml
-
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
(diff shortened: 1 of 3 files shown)


reneleonhardt:
Wow, that was a lot of work, I hope it will be merged 👍

addons/cni-canal/canal.yaml
```diff
@@ -1,4 +1,41 @@
...
+
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: calico-kube-controllers
+  namespace: kube-system
+  labels:
+    k8s-app: calico-kube-controllers
+spec:
+  maxUnavailable: 1
```
wozniakjan, line 14: I guess we just copy/paste upstream manifests but I'm curious, does this make any sense with
https://github.com/kubermatic/kubeone/blob/c21ab06001fd5d35095eb9182aa76cf850cd1cb2/addons/cni-canal/canal.yaml#L4952-L4953

WeirdMachine: I think the comment above describe the reason for this PDB:
`# This manifest creates a Pod Disruption Budget for Controller to allow K8s Cluster Autoscaler to evict`

wozniakjan: sure, I was just pondering whether having deployment where it's explicitly stated there can only be 1 replica and then defining PDB with `maxUnavailable=1` doesn't have the exact same behaviour as if no PDB was defined at all.

And I came to the conclusion it probably does but at the same time, there should be no harm in defining PDB.

wozniakjan (approved):
/approve
/lgtm

xmudrii:
/assign
I'll take a look on Monday.

addons/cni-canal/canal.yaml
```diff
@@ -41,18 +78,7 @@ data:
...
-                  [{"subnet": "usePodCidr"}],
-                  [{"subnet": "usePodCidrIPv6"}]
-              ]
-              {{ else }}
-              {{ if .Config.ClusterNetwork.HasIPv6 }}
-              "subnet": "usePodCidrIPv6"
-              {{ else }}
               "subnet": "usePodCidr"
-              {{ end }}
-              {{ end }}
```
xmudrii, line 55: Those are actually important options for IPv6, please revert this.

addons/cni-cilium/cilium.yaml
```diff
@@ -261,20 +264,16 @@ metadata:
...
...nginx.conf: "server {\n    listen       8081;\n    server_name  localhost;\n    root /app;\n    index index.html;\n    client_max_body_size 1G;\n\n    location / {\n        proxy_set_header Host $host;\n        proxy_set_header X-Real-IP $remote_addr;\n\n        # CORS\n        add_header Access-Control-Allow-Methods \"GET, POST, PUT, HEAD, DELETE, OPTIONS\";\n        add_header Access-Control-Allow-Origin *;\n        add_header Access-Control-Max-Age 1728000;\n        add_header Access-Control-Expose-Headers content-length,grpc-status,grpc-message;\n        add_header Access-Control-Allow-Headers range,keep-alive,user-agent,cache-control,content-type,content-transfer-encoding,x-accept-content-transfer-encoding,x-accept-response-streaming,x-user-agent,x-grpc-web,grpc-timeout;\n        if ($request_method = OPTIONS) {\n            return 204;\n        }\n        # /CORS\n\n        location /api {\n            proxy_http_version 1.1;\n            proxy_pass_request_headers on;\n            proxy_hide_header Access-Control-Allow-Origin;\n            proxy_pass http://127.0.0.1:8090;\n        }\n\n        location / {\n            try_files $uri $uri/ /index.html;\n        }\n    }\n}"
```
xmudrii, line 269: I think we still need this.

pkg/templates/images/images.go
```diff
@@ -213,11 +213,11 @@ func FindResource(name string) (Resource, error) {
...
 	return map[Resource]map[string]string{
-		CalicoCNI:        {"*": "quay.io/calico/cni:v3.23.5"},
-		CalicoController: {"*": "quay.io/calico/kube-controllers:v3.23.5"},
-		CalicoNode:       {"*": "quay.io/calico/node:v3.23.5"},
+		CalicoCNI:        {"*": "quay.io/calico/cni:v3.26.0"},
+		CalicoController: {"*": "quay.io/calico/kube-controllers:v3.26.0"},
+		CalicoNode:       {"*": "quay.io/calico/node:v3.26.0"},
 		DNSNodeCache:     {"*": "registry.k8s.io/dns/k8s-dns-node-cache:1.22.15"},
-		Flannel:          {"*": "quay.io/coreos/flannel:v0.15.1"},
+		Flannel:          {"*": "docker.io/flannel/flannel:v0.21.3"},
```
xmudrii, line 220: That's bad, we should mention this in the release note for heads up.

WeirdMachine: They are Sponsored OSS on docker.

addons/cni-cilium/cilium.yaml
WeirdMachine: Why? If you have a look at the updated conf you can see its listening by default on ipv4 and ipv6: `listen 8081;\n listen [::]:8081;\n`

xmudrii: That's the issue. There are some setups where listening on IPv6 doesn't work, so we drop it from the server config. I don't recall exact details now, but I think we should still keep this behavior.

WeirdMachine: Ahh okay so we support ipv6 disabled hosts in a clusters.

xmudrii (approved):
/approve

Issue #2802, opened 2023-06-13
Adding a new static worker node results in a preflight check failure on existing nodes


xmudrii:
### What happened?

Trying to add a new static worker node results in the following error:

```
+ sudo kubeadm init phase preflight --config=./kubeone/cfg/master_0.yaml
W0613 19:21:47.950292   27890 initconfiguration.go:331] [config] WARNING: Ignored YAML document with GroupVersionKind kubeadm.k8s.io/v1beta3, Kind=JoinConfiguration
W0613 19:21:47.958412   27890 initconfiguration.go:119] Usage of CRI endpoints without URL scheme is deprecated and can cause kubelet errors in the future. Automatically prepending scheme "unix" to the "criSocket" with value "/run/containerd/containerd.sock". Please update your configuration!
W0613 19:21:47.958515   27890 utils.go:69] The recommended value for "clusterDNS" in "KubeletConfiguration" is: [10.96.0.10]; the provided value is: [169.254.20.10]
	[WARNING DirAvailable--var-lib-etcd]: /var/lib/etcd is not empty
error execution phase preflight: [preflight] Some fatal errors occurred:
	[ERROR Port-6443]: Port 6443 is in use
	[ERROR Port-10259]: Port 10259 is in use
	[ERROR Port-10257]: Port 10257 is in use
	[ERROR FileAvailable--etc-kubernetes-manifests-kube-apiserver.yaml]: /etc/kubernetes/manifests/kube-apiserver.yaml already exists
	[ERROR FileAvailable--etc-kubernetes-manifests-kube-controller-manager.yaml]: /etc/kubernetes/manifests/kube-controller-manager.yaml already exists
	[ERROR FileAvailable--etc-kubernetes-manifests-kube-scheduler.yaml]: /etc/kubernetes/manifests/kube-scheduler.yaml already exists
	[ERROR FileAvailable--etc-kubernetes-manifests-etcd.yaml]: /etc/kubernetes/manifests/etcd.yaml already exists
	[ERROR Port-10250]: Port 10250 is in use
	[ERROR Port-2379]: Port 2379 is in use
	[ERROR Port-2380]: Port 2380 is in use
[preflight] If you know what you are doing, you can make a check non-fatal with `--ignore-preflight-errors=...`
To see the stack trace of this error execute with --v=5 or higher
```

[...]

PR #2803, opened 2023-06-13
Ignore some kubeadm preflight checks when validating cluster requirements

xmudrii:

Ignore some `kubeadm` preflight checks when validating cluster requirements to account for adding new static worker nodes. For more details about this, see #2802

xref #2802

**What type of PR is this?**

/kind bug
/kind regression

```release-note
Ignore some `kubeadm` preflight checks when validating cluster requirements to account for adding new static worker nodes
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/pkg/tasks/prerequisites.go
+++ b/pkg/tasks/prerequisites.go
@@ -50,6 +50,7 @@ func kubeadmPreflightChecks(s *state.State) error {
 			ctx.Logger.Info("	preflight...")
 			_, _, err := ctx.Runner.Run(heredoc.Docf(`
 				sudo kubeadm init phase preflight \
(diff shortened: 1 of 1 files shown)


pkg/tasks/prerequisites.go
```diff
@@ -50,6 +50,7 @@ func kubeadmPreflightChecks(s *state.State) error {
 			ctx.Logger.Info("	preflight...")
 			_, _, err := ctx.Runner.Run(heredoc.Docf(`
 				sudo kubeadm init phase preflight \
+					--ignore-preflight-errors=DirAvailable--var-lib-etcd,FileAvailable--etc-kubernetes-manifests-kube-apiserver.yaml,FileAvailable--etc-kubernetes-manifests-kube-controller-manager.yaml,FileAvailable--etc-kubernetes-manifests-kube-scheduler.yaml,FileAvailable--etc-kubernetes-manifests-etcd.yaml,Port-6443,Port-10259,Port-10257,Port-10250,Port-2379,Port-2380 \
```
kron4eg, line 53: shouldn't we use kubeadm config? and we already partually have some

```golang
	nodeRegistration := newNodeRegistration(s, host)
	nodeRegistration.IgnorePreflightErrors = []string{
		"DirAvailable--var-lib-etcd",
		"DirAvailable--etc-kubernetes-manifests",
		"ImagePull",
	}
```

xmudrii: No, that would apply ignore list to `kubeadm init` as well. We only want to apply this ignore list to `kubeadm init` running preflight checks.

PR #2807, opened 2023-06-23
Use buildx instead of buildah for building images

xmudrii:

Detailed reasoning behind this PR can be found here: https://github.com/kubermatic/kubermatic/pull/12393

Short version: `golang` images are now using Debian 12 (bookworm) as the base image. This broke our `buildah` because it's too old and doesn't work with Debian 12. We decided to switch to `buildx` because it's integrated in Docker and providers a bit better experience. This is also matching what are we doing for other projects such KKP.

**What type of PR is this?**

/kind cleanup

```release-note
Use buildx instead of Buildah to create multi-architecture KubeOne container images.
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/hack/ci/push-image.sh
+++ b/hack/ci/push-image.sh
@@ -27,55 +27,56 @@ set -o monitor
 
 source $(dirname $0)/../lib.sh
 
-DOCKER_REPO="${DOCKER_REPO:-quay.io/kubermatic}"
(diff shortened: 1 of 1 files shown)


xmudrii:
/assign @xrstf 
@xrstf already did this migration for KKP

Issue #2781, opened 2023-05-10
Support Kubernetes 1.27


xmudrii:
This is a collector issue for Kubernetes 1.27 support in KubeOne. The following tasks should be taken care of:

* [x] Check the Kubernetes changelog to ensure there are no breaking changes and removals
* [x] Update the `build` image to update Sonobuoy (and other dependencies if needed)
* [x] Update the latest supported Kubernetes version in [the API validation](https://github.com/kubermatic/kubeone/blob/main/pkg/apis/kubeone/validation/validation.go#L40-L41) (#2812)
* [x] Update [default admission controllers](https://github.com/kubermatic/kubeone/blob/main/pkg/kubeflags/data.go) if needed (#2812)
* [x] Update [the stable version marker in Makefile](https://github.com/kubermatic/kubeone/blob/5273f9a372736569c6b09b38f2959019d29e4d6a/Makefile#L24) (#2812)
* [x] Update [machine-controller and operating-system-manager](https://github.com/kubermatic/kubeone/blob/a5c3a14725bf135cb2682c4ec96d5872b165b98f/pkg/templates/images/images.go#L222-L224) (#2812)
* [x] Add E2E tests (#2812)
* [x] Update daily periodics to use the latest Kubernetes release
* [x] Update [the Compatibility Matrix](https://docs.kubermatic.com/kubeone/main/architecture/compatibility/supported-versions/) (https://github.com/kubermatic/docs/pull/1492)
* [x] Create an issue to track updating [images](https://github.com/kubermatic/kubeone/blob/main/pkg/templates/images/images.go) (#2782)
* [x] Run the full conformance tests suite using [Sonobuoy](https://github.com/vmware-tanzu/sonobuoy) (#2850)

kron4eg:
all tasks done

PR #2812, opened 2023-07-05
Add support for Kubernetes 1.27 and update Go to 1.20.5

xmudrii:

This PR adds support for Kubernetes 1.27 to KubeOne. 🎉 

Along with that, this PR brings some important changes:

- Go is updated to 1.20.5
- KubeOne container image is now based on Alpine 3.17
- Default stable version is now 1.27
- Explicitly bind the sandbox (pause) image for both kubeadm and containerd
  - This is to avoid the following warning:
    ```
    W0705 09:09:16.961128    4667 checks.go:835] detected that the sandbox image "registry.k8s.io/pause:3.6" of the container runtime is inconsistent with that used by kubeadm. It is recommended that using "registry.k8s.io/pause:3.9" as the CRI sandbox image.
    ```
  - This warning is caused by kubeadm being configured (by default) to use newer sandbox image than containerd. This should be fixed now that we explicitly bind the image
- External CCM is now required for AWS because in-tree provider got removed in 1.27
- The list of enabled-by-default admission controllers is updated
  - It turns out that we forgot to update this list for 1.26
- Get rid of deprecated and removed flags such as `--container-runtime`
- Ensure CRI socket is provided with `unix://` prefix to avoid the following warning:
    ```
    W0705 09:09:29.145737    4745 initconfiguration.go:120] Usage of CRI endpoints without URL scheme is deprecated and can cause kubelet errors in the future. Automatically prepending scheme "unix" to the "criSocket" with value "/run/containerd/containerd.sock". Please update your configuration!
    ```
- machine-controller and OSM are updated to the latest stable version
- E2E tests are updated to reflect external CCM requirements

xref #2781 

**What type of PR is this?**

/kind feature

```release-note
- Add support for Kubernetes 1.27
- AWS-based clusters require using external cloud controller manager (CCM) with Kubernetes 1.27 and newer. Existing clusters running in-tree cloud provider must migrate to external CCM before upgrading to Kubernetes 
- Explicitly bind the pause image (version depends on Kubernetes version) to avoid version drift between kubeadm/kubelet and containerd
- Update Go to 1.20.5
- Update base image for KubeOne container image to alpine:3.17
- Update machine-controller to v1.57.0
- Update operating-system-manager to v1.3.0
```

**Documentation**:
```documentation
TBD
```


kron4eg:
letting @ahmedwaleedmalik to take a look (thus not approved yet)

PR #2828, opened 2023-07-21
Rename helm values files to avoid rendering them to kube api

kron4eg:
Our addons rendering mechanism specifically looking for yaml files to render and send to kube-api. Renamed files should not end up in kube-api as they don't contain kubernetes objects.

https://github.com/kubermatic/kubeone/blob/ea3e7bf58fab30b79b68c03864a28ee5e02f843b/pkg/addons/manifest.go#L133

**What type of PR is this?**
/kind bug

```release-note
Rename helm values files to avoid rendering them to kube api
```

**Documentation**:

```documentation
NONE
```


xmudrii (approved):
/lgtm
/approve

PR #2836, opened 2023-07-31
chartURL field

kron4eg:
Currently it is possible to install a chart simply by assigning the direct archive URL of the chart. For example:
```yaml
helmReleases:
  - releaseName: flannel
    namespace: kube-system
    chart: https://github.com/flannel-io/flannel/releases/download/v0.22.1/flannel.tgz
```

Except when this didn't worked. Even run of `kubeone apply` will install the chart, and the odd run it will uninstall the chart. This happens because the `chart` is also a NAME of the chart, when it's already installed our internal heuristics will proceed to compare it to the chart from the release history. Because names are different (one is URL and the other is actual chart name, extracted from that archive) that causes KubeOne to determinate that it should be uninstalled as missing from the manifest.

So to still provide this capability and stop install/uninstall cycle the new chartURL field is introduced:
```yaml
helmReleases:
  - releaseName: flannel
    chart: flannel
    namespace: kube-system
    chartURL: https://github.com/flannel-io/flannel/releases/download/v0.22.1/flannel.tgz
```

**What type of PR is this?**
/kind bug

This is the strange case of a bug turned into a feature, or rather a feature introduction to help mitigate the bug :shrug: 

```release-note
Add chartURL field to helmRelease
```

**Documentation**:

```documentation
NONE
```


xmudrii (approved):
/lgtm
/approve

xmudrii:
Test are failing
/lgtm cancel

xmudrii (approved):

PR #2854, opened 2023-08-10
Take into account TEST_TIMEOUT env for internal context timeout

kron4eg:
We have some E2E tests that take too long, and we specifically configure them with long timeouts (TEST_TIMEOUT: 120m). But we forgot to configure this internal context, and it fail the test regardless of TEST_TIMEOUT :shrug: 

**What type of PR is this?**
/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


xrstf (approved):
/approve

PR #2855, opened 2023-08-14
Drop v1.27 constraint for vSphere, since CCM has been updated

kron4eg:
Previously there was no updated vSphere CCM release to work with kubernetes v1.27, but since that time there update has been integrated so I'm removing the "blocking" code.

**What type of PR is this?**
/kind bug

```release-note
Drop v1.27 constraint for vSphere, since CCM has been updated
```

**Documentation**:

```documentation
NONE
```


xmudrii:
I don't think this PR needs a release note.
/release-note-none

PR #2872, opened 2023-08-28
add option to skip local storage for autoscaler addon

c4tz:

The cluster-autoscaler addon often does not scale down the nodepool if there are pods running which have an `emptyDir` volume, e.g. The default behaviour of keeping these nodes is perfectly OK, but there are also many reasons why someone wouldn't want that. In order to enable KubeOne users to choose the option they prefer, I added another parameter to the addon.

None, its a new feature

**What type of PR is this?**

/kind feature

```release-note
Add `CLUSTER_AUTOSCALER_SKIP_LOCAL_STORAGE` parameter for cluster-autoscaler addon used to skip local storage when downscaling nodes (see https://github.com/kubermatic/kubeone/tree/main/addons/cluster-autoscaler for more details)
```

**Documentation**:
```documentation
TBD
```

I already updated the `addons/cluster-autoscaler/README.md` in my commit :slightly_smiling_face:

### DIFF
--- a/addons/cluster-autoscaler/README.md
+++ b/addons/cluster-autoscaler/README.md
@@ -100,14 +100,16 @@ kubectl annotate machinedeployment -n kube-system <machinedeployment-name> clust
 
 You need to replace the following values with the actual ones:
 
-* `AUTOSCALER_VERSION` needs to be replaced with the appropriate Cluster
-  Autoscaler version
+* `CLUSTER_AUTOSCALER_IMAGE` can be used to replace the Cluster Autoscaler image
   * The minor versions of Cluster Autoscaler and Kubernetes cluster should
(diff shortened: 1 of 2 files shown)


c4tz:
@kron4eg I just realised I set the wrong default (normally its `true`), so this might be a breaking change if you release it like this in v1.7.

Should I open another PR or could you switch the default value for this option to `true` (Also in the readme)? :man_facepalming:

kron4eg:
> @kron4eg I just realised I set the wrong default (normally its `true`), so this might be a breaking change if you release it like this in v1.7.
> 
> Should I open another PR or could you switch the default value for this option to `true` (Also in the readme)? 🤦‍♂️

please open a new PR

PR #2916, opened 2023-09-15
Add `CCM_CONCURRENT_SERVICE_SYNCS` addon parameter to all CCMs

### LINKED ISSUE #2895: Allow configuration of parallel service reconciliation in CCM

As a user deploying clusters with KubeOne
I would like to be able to configure `--concurrent-service-syncs` on the CCM deployed by KubeOne
in order to optimise load balancer reconcile performance.

### Solution details

- Be able to configure the`--concurrent-service-syncs` flag in the CCM `Deployment` (or `StatefulSet`) through [a global addon parameter](https://docs.kubermatic.com/kubeone/v1.6/guides/addons/#parameters), e.g. `CCM_CONCURRENT_SERVICE_SYNCS`. This should default to 1 (the upstream default).
- Not per cloud provider but one generic setting. They all support this, the flag is part of the framework to create a CCM.

### Alternative approaches

Make this a per CCM addon (so, for each cloud provider) independent setting. Since it works the same way for all CCMs, and only a single CCM can be deployed anyway, this seems less effective and more maintenance overhead.

### Additional information

The source of the flag in `k8s.io/cloud-provider`: https://github.com/kubernetes/cloud-provider/blob/742510807782c9f7f7ac207bb2a9dad11d209ff3/options/servicecontroller.go#L35

embik:
This PR adds an addon parameter called `CCM_CONCURRENT_SERVICE_SYNCS` to all CCM addons. It will set the `--concurrent-service-syncs` flag, which is added by the generic CCM framework (`k8s.io/cloud-provider`) and determines how many load balancer `Services` are reconciled at the same time. This can be useful if a single loop is too slow to catch up on all changes in the cluster.

Fixes #2895

**What type of PR is this?**
/kind feature

```release-note
Allow setting `CCM_CONCURRENT_SERVICE_SYNCS` parameter on CCM addons to configure number of concurrent `LoadBalancer` service reconciles
```

**Documentation**:

```documentation
https://github.com/kubermatic/docs/pull/1517
```

### DIFF
--- a/addons/ccm-aws/ccm-aws.yaml
+++ b/addons/ccm-aws/ccm-aws.yaml
@@ -177,6 +177,9 @@ spec:
             - --configure-cloud-routes=false
             - --cloud-config=/etc/kubernetes/cloud/cloud-config
             - --v=2
+            {{- with .Params.CCM_CONCURRENT_SERVICE_SYNCS }}
+            - --concurrent-service-syncs={{ . }}
+            {{- end }}
           resources:
             requests:
               cpu: 200m
--- a/addons/ccm-azure/ccm-azure.yaml
+++ b/addons/ccm-azure/ccm-azure.yaml
@@ -289,6 +289,9 @@ spec:
             - "--route-reconciliation-period=10s"
             - "--secure-port=10268"
             - "--v=2"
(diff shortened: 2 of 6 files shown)


xmudrii (approved):
Thank you for taking care of this! ✨ 
/lgtm
/approve

PR #2925, opened 2023-10-09
Apply registry override to sandbox (pause) image

embik:
For static nodes provisioned by KubeOne, there is no way to override the sandbox/pause image. This is possible already for MC and OSM, and KubeOne sets the `--pause-image` flags on those correctly (by applying registry overrides), so I believe that this is a bug and should be applied to static nodes as well.

Fixes #

**What type of PR is this?**
/kind bug

```release-note
`registryConfiguration.OverrideRegistry` is correctly applied to the pause image configured in static nodes (control plane and static workers)
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/apis/kubeone/helpers.go
+++ b/pkg/apis/kubeone/helpers.go
@@ -241,21 +241,23 @@ func (crc ContainerRuntimeConfig) CRISocket() string {
 	return ""
 }
 
-func (v VersionConfig) SandboxImage() (string, error) {
+func (v VersionConfig) SandboxImage(imageRegistry func(string) string) (string, error) {
 	kubeSemVer, err := semver.NewVersion(v.Kubernetes)
 	if err != nil {
 		return "", fail.Config(err, "parsing kubernetes semver")
 	}
 
+	registry := imageRegistry("registry.k8s.io")
+
 	switch {
 	case v124Constraint.Check(kubeSemVer):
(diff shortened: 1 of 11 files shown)


xmudrii (approved):
Thank you for fixing this @embik!
/lgtm
/approve

pkg/apis/kubeone/helpers.go
```diff
@@ -241,21 +241,23 @@ func (crc ContainerRuntimeConfig) CRISocket() string {
 	return ""
 }
 
-func (v VersionConfig) SandboxImage() (string, error) {
+func (v VersionConfig) SandboxImage(imageRegistry func(string) string) (string, error) {
```
kron4eg, line 244: If we anyway always use `cluster.RegistryConfiguration.ImageRegistry` as an argument, maybe we don't need to take it as an argument and just use directly :shrug:

kron4eg:
/hold

to resolve a discussion

pkg/apis/kubeone/helpers.go
embik: Can you specify what you mean? `v VersionConfig` doesn't have access to `cluster.RegistryConfiguration.ImageRegistry` as far as I know.

kron4eg: oh well :shrug: 
OK then

PR #2945, opened 2023-10-31
vcd: don't fail credential validation when only an API token is passed

embik:

This is a follow-up to #2751 as reported via INC-6166 to us. The linked PR updates the credential validation for VCD, but runs into the problem that it checks for two things in consecutive if conditions:

1. user credentials (username + password) and API token are mutually exclusive to each other
2. if username and password are not set, fail with them being required

These two conditions stacked on top of each other meant that `VCD_API_TOKEN` was never usable, as using it alongside user credentials was blocked by 1, but using it _without_ user credentials was blocked by 2.

The reporter helpfully pointed out the condition that this PR is updating. In addition, I've added a few simple test cases that should catch such errors in the future.

Fixes #

**What type of PR is this?**
/kind bug

```release-note
Fix incorrect validation that made `VCD_API_TOKEN` unusable for VMware Cloud Director
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/credentials/credentials.go
+++ b/pkg/credentials/credentials.go
@@ -642,7 +642,7 @@ func vmwareCloudDirectorValidationFunc(creds map[string]string) error {
 	}
 
 	// Username and password are required when using default credentials i.e. username and password.
-	if !(username && password) {
+	if !apiToken && !(username && password) {
 		return fail.CredentialsError{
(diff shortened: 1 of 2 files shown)


xmudrii (approved):
/lgtm
/approve

Issue #2835, opened 2023-07-27
Update auxiliary images to support Kubernetes 1.28


xmudrii:
Reference to the previous issue for updating images: https://github.com/kubermatic/kubeone/issues/2782

Action items:

- [x] Update the issue template to add/remove [images](https://github.com/kubermatic/kubeone/blob/main/pkg/templates/images/images.go) as appropriate

The following components/images should be updated:

### General

- [x] [Canal CNI](https://github.com/projectcalico/calico) (#2949)
- [x] [Calico VXLAN CNI](https://github.com/projectcalico/calico) (#2949)
- [x] [Cilium CNI](https://github.com/cilium/cilium) (#2949)
- [x] [NodeLocalDNS](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml) (already the latest)
- [x] [metrics-server](https://github.com/kubernetes-sigs/metrics-server) (#2949)
- [x] [Cluster Autoscaler](https://github.com/kubernetes/autoscaler) (#2949)

### Cloud provider components

- [x] [AWS CCM](https://github.com/kubernetes/cloud-provider-aws) 
- [x] [AWS CSI](https://github.com/kubernetes-sigs/aws-ebs-csi-driver) #3013
- [x] [Azure CCM](https://github.com/kubernetes-sigs/cloud-provider-azure) #3019
- [x] [AzureDisk CSI](https://github.com/kubernetes-sigs/azuredisk-csi-driver) #3019
- [x] [AzureFile CSI](https://github.com/kubernetes-sigs/azurefile-csi-driver) #3019
- [x] [DigitalOcean CCM](https://github.com/digitalocean/digitalocean-cloud-controller-manager) #3027
- [x] [DigitalOcean CSI](https://github.com/digitalocean/csi-digitalocean) #3026
- [x] [EquinixMetal CCM](https://github.com/equinix/cloud-provider-equinix-metal) #3028
- [x] [GCP CCM](https://github.com/kubernetes/cloud-provider-gcp) #3038
- [x] [GCP Compute Persistent Disk CSI](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver) #3023
- [x] [Hetzner CCM](https://github.com/hetznercloud/hcloud-cloud-controller-manager) #3022
- [x] [Hetzner CSI](https://github.com/hetznercloud/csi-driver) #3022
[...]

PR #2949, opened 2023-11-05
Update metrics-server, cluster-autoscaler, Canal, Calico VXLAN, and Cilium (Kubernetes 1.28)

xmudrii:

- Update Canal to v3.26.3
- Update Calico VXLAN addon to v3.26.3
- Update Cilium to v1.14.3
- Update metrics-server to v0.6.4
- Add cluster-autoscaler support for v1.28, and update cluster-autoscaler to v1.27.3, v1.26.4, v1.25.3

xref #2835

**What type of PR is this?**

/kind feature

```release-note
- Update Canal to v3.26.3
- Update Calico VXLAN addon to v3.26.3
- Update Cilium to v1.14.3
- Update metrics-server to v0.6.4
- Add cluster-autoscaler support for v1.28, and update cluster-autoscaler to v1.27.3, v1.26.4, v1.25.3
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/addons/cni-cilium/cilium.yaml
+++ b/addons/cni-cilium/cilium.yaml
@@ -1,5 +1,5 @@
 # Generated from:
-#   helm template cilium cilium/cilium --version 1.14.1 --namespace kube-system --set operator.replicas=1 \
+#   helm template cilium cilium/cilium --version 1.14.3 --namespace kube-system --set operator.replicas=1 \
 #     --set hubble.tls.auto.method=cronJob --set hubble.relay.enabled=true --set hubble.ui.enabled=true \
 #     --set kubeProxyReplacement=strict --set k8sServiceHost=CHANGEME --set k8sServicePort=CHANGEME
 #
@@ -70,8 +70,6 @@ data:
   cilium-endpoint-gc-interval: "5m0s"
   nodes-gc-interval: "5m0s"
   skip-cnp-status-startup-clean: "false"
-  # Disable the usage of CiliumEndpoint CRD
-  disable-endpoint-crd: "false"
 
   # If you want to run cilium in debug mode change this value to true
   debug: "false"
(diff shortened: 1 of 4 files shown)


addons/cni-cilium/cilium.yaml
```diff
@@ -193,7 +191,7 @@ data:
   enable-k8s-networkpolicy: "true"
   # Tell the agent to generate and write a CNI configuration file
   write-cni-conf-when-ready: /host/etc/cni/net.d/05-cilium.conflist
-  cni-exclusive: "false"
+  cni-exclusive: "true"
```
xmudrii, line 194: I'm not sure about this change, I have to double-check where is it coming from

addons/cni-cilium/cilium.yaml
```diff
@@ -70,8 +70,6 @@ data:
   cilium-endpoint-gc-interval: "5m0s"
   nodes-gc-interval: "5m0s"
   skip-cnp-status-startup-clean: "false"
-  # Disable the usage of CiliumEndpoint CRD
-  disable-endpoint-crd: "false"
```
xmudrii, line 74: I'm not sure about this change, I have to double-check where is it coming from

xmudrii: https://github.com/cilium/cilium/commit/c6bc4d61c2d807decdb61674c5115ad555580df4

addons/cni-cilium/cilium.yaml
xmudrii: I found the following description for this parameter:

```
  # -- Make Cilium take ownership over the `/etc/cni/net.d` directory on the
  # node, renaming all non-Cilium CNI configurations to `*.cilium_bak`.
  # This ensures no Pods can be scheduled using other CNI plugins during Cilium
  # agent downtime.
```

I wasn't able to find why we disabled it in KubeOne, but I stumbled across the following comment in the [KKP codebase](https://github.com/kubermatic/kubermatic/blob/2e30b91169a4f69a13486a58706ea4ce3ec832a6/pkg/cni/cilium/cilium.go#L241-L242):

> // we run Cilium as non-exclusive CNI to allow for Multus use-cases

I believe this might apply to KubeOne too, so I disabled this parameter and commented why is it disabled.

xmudrii:
Decoupled from #2948 and ready for review

PR #2967, opened 2023-11-13
Conditional Templating for VCD_USER, VCD_Password, VCD_API_TOKEN

SimonTheLeg:

This came in via a support ticket. Without any defaults, the b64enc func fails if username/password are being left empty

```
INFO[10:49:53 CET] Parsing addons manifest 'vcloud-basic-auth.yaml'
WARN[10:49:53 CET] Task failed, error was: runtime: executing addons manifest template "vcloud-basic-auth.yaml"
template: addons-base:8:39: executing "addons-base" at <b64enc>: invalid value; expected string
```

**What type of PR is this?**

/kind bug

I think the requester assumes that leaving username and password empty should be possible. I am not super familiar with vmware vcloud, so I am not sure if that makes sense. It would be good if you could double check this. 
If the two are fields required, then maybe we could make use of the `required` template func in order to avoid confusion?

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/addons/csi-vmware-cloud-director/vcloud-basic-auth.yaml
+++ b/addons/csi-vmware-cloud-director/vcloud-basic-auth.yaml
@@ -5,6 +5,12 @@ metadata:
   name: vcloud-basic-auth
   namespace: kube-system
 data:
-  username: {{ default "" .Credentials.VCD_USER | b64enc }}
-  password: {{ default "" .Credentials.VCD_PASSWORD | b64enc }}
-  refreshToken: {{ default "" .Credentials.VCD_API_TOKEN | b64enc }}
+{{- with .Credentials.VCD_USER }}
+  username: {{ . | b64enc }}
+{{- end }}
+{{- with .Credentials.VCD_PASSWORD }}
+  password: {{ . | b64enc }}
+{{- end }}
+{{- with .Credentials.VCD_API_TOKEN }}
+  refreshToken: {{ . | b64enc }}
+{{- end }}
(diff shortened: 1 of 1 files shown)


addons/csi-vmware-cloud-director/vcloud-basic-auth.yaml
```diff
@@ -5,6 +5,6 @@ metadata:
   name: vcloud-basic-auth
   namespace: kube-system
 data:
-  username: {{ .Credentials.VCD_USER | b64enc }}
-  password: {{ .Credentials.VCD_PASSWORD | b64enc }}
+  username: {{ default "" .Credentials.VCD_USER | b64enc }}
+  password: {{ default "" .Credentials.VCD_PASSWORD | b64enc }}
   refreshToken: {{ default "" .Credentials.VCD_API_TOKEN | b64enc }}
```
xmudrii, line 10: I asked @ahmedwaleedmalik about this change, and we think we should drop fields that are not used instead of doing this. For example, if `username` and `password` are not used, we should leave only `refreshToken`, same other way around. Can you please update the PR to reflect this requirement?

xmudrii (changes requested):

addons/csi-vmware-cloud-director/vcloud-basic-auth.yaml
SimonTheLeg: Most recent change should do the trick

xmudrii (approved):
Thank you!
/lgtm
/approve

PR #1486, opened 2021-09-02
Enable Addon api

WeirdMachine:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
xref #1459 

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #2981, opened 2023-12-14
Update backup-restic to etcd 3.5.11 and fix addon documentation

### LINKED ISSUE #2980: Update etcd image in `backups-restic` addon to distroless versions

The `backups-restic` addon includes usage of an `etcd:v3.5.6` image. That version should be updated to the latest `v3.5` image available. They key change is that newer versions of the image are based on `distroless` and as such, do not have a shell available.

### Solution details

- Update `backups-restic` Job definition to use latest `etcd:v3.5.x` image available
- Change logic to account for etcd images no longer providing a shell

### Alternative approaches

N/A

### Use cases

N/A

### Additional information

Internal ref: INC-6251

embik:
This updates the `backup-restic` addon to use etcd 3.5.11, the latest available 3.5.x image. Because the images are now using distroless, I have moved the parts that copy the certificates into the backup into the `restic` container before uploading the backup. Given that the container will have access to these files anyway (since its uploading them to s3), I did not see any gain from having another initContainer doing it.

In addition, while trying out the addon, I ran into a documentation issue - The `s3Bucket` param does _not_ do what it's documented to do. If you just put a bucket name as is documented, the backups will be written to a local folder (because `RESTIC_REPOSITORY` is set to the contents of `s3Bucket` verbatim).

Now I had two choices, either fix the way `RESTIC_REPOSITORY` is set from `s3Bucket` or update documentation for the field. I have chosen to go with the latter, because anyone using the addon and validating their backups (which I hope everyone does ... right?) would have figured this out already. I'm afraid that breaking the existing way is going to create more issues (break it for everyone who already sets the `s3:/` part).

I'm open to change this in a follow-up PR, but it wasn't the focus of this one so I just wanted to write down the as-is situation correctly.

Fixes #2980

**What type of PR is this?**
/kind feature

```release-note
Update backup-restic addon to use etcd 3.5.11 for creating etcd snapshots
```

**Documentation**:

```documentation
https://github.com/kubermatic/docs/pull/1591
```

### DIFF
--- a/addons/backups-restic/backups-restic.yaml
+++ b/addons/backups-restic/backups-restic.yaml
@@ -52,24 +52,14 @@ spec:
               path: /etc/kubernetes/pki
           initContainers:
           - name: snapshoter
-            image: {{ Registry "gcr.io" }}/etcd-development/etcd:v3.5.6
+            image: {{ Registry "gcr.io" }}/etcd-development/etcd:v3.5.11
             imagePullPolicy: IfNotPresent
             command:
-            - /bin/sh
-            - -c
-            - |-
-              set -euf
-              mkdir -p /backup/pki/kubernetes
-              mkdir -p /backup/pki/etcd
-              cp -a /etc/kubernetes/pki/etcd/ca.crt /backup/pki/etcd/
-              cp -a /etc/kubernetes/pki/etcd/ca.key /backup/pki/etcd/
-              cp -a /etc/kubernetes/pki/ca.crt /backup/pki/kubernetes
-              cp -a /etc/kubernetes/pki/ca.key /backup/pki/kubernetes
-              cp -a /etc/kubernetes/pki/front-proxy-ca.crt /backup/pki/kubernetes
-              cp -a /etc/kubernetes/pki/front-proxy-ca.key /backup/pki/kubernetes
-              cp -a /etc/kubernetes/pki/sa.key /backup/pki/kubernetes
-              cp -a /etc/kubernetes/pki/sa.pub /backup/pki/kubernetes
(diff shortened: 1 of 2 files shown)


xmudrii:
> Now I had two choices, either fix the way RESTIC_REPOSITORY is set from s3Bucket or update documentation for the field. I have chosen to go with the latter, because anyone using the addon and validating their backups (which I hope everyone does ... right?) would have figured this out already. I'm afraid that breaking the existing way is going to create more issues (break it for everyone who already sets the s3:/ part).

I agree with going forward with the latter option. The former option would be a breaking change for sure.

addons/backups-restic/backups-restic.yaml
```diff
@@ -52,24 +52,14 @@ spec:
               path: /etc/kubernetes/pki
           initContainers:
           - name: snapshoter
-            image: {{ Registry "gcr.io" }}/etcd-development/etcd:v3.5.6
+            image: {{ Registry "gcr.io" }}/etcd-development/etcd:v3.5.11
```
xmudrii, line 55: We might want to use `k8s.gcr.io` instead in the future.

xmudrii (approved):
/lgtm
/approve

addons/backups-restic/backups-restic.yaml
embik: We should, KKP is also still using this image.

embik:
Re: s3 bucket name, this was accidentally changed in https://github.com/kubermatic/kubeone/pull/1486/commits/d751b26aebeea2043ab142dd179e7c31f6fba4c4.

embik:
/retest

guess that is a flake.

PR #3035, opened 2024-02-08
Replace JSON6902 with Strategic Merge in Nutanix CSI driver

kron4eg:
This PR replaces required manual edits for `caBundleEnvVar`, `caBundleVolumeMount`, `caBundleVolume` with automated strategic merge patch. Instead of using error prone index (position in array) based JSON6902 patches, strategic merge patch uses stable names!

The trick that allow such configMap volume definition to work is `optional: true`. If we have caBundle defined, it will mount the volume, if no, nothing happens (no such path to SSL cabundle, so it's ignored).

**What type of PR is this?**
/kind cleanup

```release-note
Replace JSON6902 with Strategic Merge in Nutanix CSI driver
```

**Documentation**:

```documentation
NONE
```


addons/csi-nutanix/nutanix-csi-storage.yaml
```diff
@@ -339,15 +339,14 @@ spec:
...
+              value: /etc/kubeone/certs/ca-certificates.crt
             - name: CSI_ENDPOINT
               value: unix:///var/lib/csi/sockets/pluginproxy/csi.sock
             - name: NODE_ID
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
-{{ if .Config.CABundle }}
-{{ caBundleEnvVar | indent 12 }}
-{{ end }}
```
xmudrii, line 350: Is this expected?

kron4eg: yes. I've replaced those manual out of YAML/kustomize edit with static env/volumes/volumeMounts. volume sourced from configMap however is marked as optional.

xmudrii (approved):
/lgtm
/approve

PR #3040, opened 2024-02-14
Fix: Use containerd binary from path in Flatcar

4ch3los:
With the latest stable release of flatcar(https://www.flatcar.org/releases#release-3815.2.0), they removed the old containerd installation with torcx and moved to the upstream gentoo builds. This results in a broken containerd.service, because the kubeone dropin refers to the torcx path. Both old(LTS) and new versions PATH var contains the binary, so we are safe to just use the binary from PATH.

NONE

**What type of PR is this?**

/kind bug

```release-note
fixed support for flatcar stable channel 3815.2.0
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/pkg/scripts/render.go
+++ b/pkg/scripts/render.go
@@ -122,7 +122,7 @@ var containerRuntimeTemplates = map[string]string{
 			Restart=always
 			Environment=CONTAINERD_CONFIG=/etc/containerd/config.toml
 			ExecStart=
-			ExecStart=/usr/bin/env PATH=\${TORCX_BINDIR}:\${PATH} \${TORCX_BINDIR}/containerd --config \${CONTAINERD_CONFIG}
+			ExecStart=/usr/bin/env containerd --config \${CONTAINERD_CONFIG}
 			EOF
 		`),
 }
--- a/pkg/scripts/render.go
+++ b/pkg/scripts/render.go
@@ -122,7 +122,7 @@ var containerRuntimeTemplates = map[string]string{
 			Restart=always
 			Environment=CONTAINERD_CONFIG=/etc/containerd/config.toml
 			ExecStart=
-			ExecStart=/usr/bin/env containerd --config \${CONTAINERD_CONFIG}
+			ExecStart=/usr/bin/env PATH=\${TORCX_BINDIR}:\${PATH} containerd --config \${CONTAINERD_CONFIG}
 			EOF
 		`),
 }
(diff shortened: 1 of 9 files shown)


kron4eg:
Thank you @4ch3los!

You will need to update fixtures too. Please run `hack/update-fixtures.sh` and commit the results.

@ahmedwaleedmalik I think we need same change for the OSM, don't we?

kron4eg:
I wonder if we can make it compatible with the previous flatcar versions (some people run outdated images).

Since containerd already is in the `${TORCX_BINDIR}`, and `${TORCX_BINDIR}` is in the PATH, we can simply do this, that supposedly would make it forward and backward compatible.

```diff
diff --git pkg/scripts/render.go pkg/scripts/render.go
index c1d3f76c..3576bb9f 100644
--- pkg/scripts/render.go
+++ pkg/scripts/render.go
@@ -122,7 +122,7 @@ var containerRuntimeTemplates = map[string]string{
                        Restart=always
                        Environment=CONTAINERD_CONFIG=/etc/containerd/config.toml
                        ExecStart=
-                       ExecStart=/usr/bin/env PATH=\${TORCX_BINDIR}:\${PATH} \${TORCX_BINDIR}/containerd --config \${CONTAINERD_CONFIG}
+                       ExecStart=/usr/bin/env PATH=\${TORCX_BINDIR}:\${PATH} containerd --config \${CONTAINERD_CONFIG}
                        EOF
                `),
 }
```

pkg/scripts/render.go
```diff
@@ -122,7 +122,7 @@ var containerRuntimeTemplates = map[string]string{
 			Restart=always
 			Environment=CONTAINERD_CONFIG=/etc/containerd/config.toml
 			ExecStart=
-			ExecStart=/usr/bin/env PATH=\${TORCX_BINDIR}:\${PATH} \${TORCX_BINDIR}/containerd --config \${CONTAINERD_CONFIG}
+			ExecStart=/usr/bin/env containerd --config \${CONTAINERD_CONFIG}
```
kron4eg, line 125:
```suggestion
			ExecStart=/usr/bin/env PATH=\${TORCX_BINDIR}:\${PATH} containerd --config \${CONTAINERD_CONFIG}
```

please change it to make it backward compatible

kron4eg (changes requested):
small suggestion

pkg/scripts/render.go
kron4eg: LTS was also release "just now", so no wander it works too. I means some older images, before the Release Date: Feb 14, 2024 :shrug:

4ch3los: Ahh well, the lts release wasnt released yesterday, when i tested it :D But yeah i will just add it, as it shouldnt hurt

kron4eg (approved):
/approve

kron4eg:
Okey, I'll create manual cherrypicks

PR #3048, opened 2024-02-19
Add initial support for kubernetes 1.29

kron4eg:

xref #2997

**What type of PR is this?**
/kind chore

cri-tools and kubernetes-cni packages don't need version specification anymore as we are already using version locked repository per kubernetes version.

```release-note
* Add initial support for kubernetes 1.29
* Set cloudProvider.external = true by default for supported cloud providers in kubernetes 1.29+
```

**Documentation**:

```documentation
NONE
```


kron4eg:
Some more PRs will follow up

ahmedwaleedmalik (approved):
/approve 🚀

test/tests.yml
```diff
@@ -30,760 +47,872 @@
...
+
+- scenario: csi_ccm_migration
+  initVersion: v1.27.11
+  infrastructures:
+    - name: aws_default
+
+- scenario: csi_ccm_migration
+  initVersion: v1.28.7
+  infrastructures:
+    - name: aws_default
```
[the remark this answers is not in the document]
xmudrii: We should also add CCM/CSI migration tests for GCP (same situation as Azure).

test/tests.yml
```diff
@@ -30,760 +47,872 @@
...
+    - name: hetzner_default
+    - name: hetzner_rockylinux
+    - name: openstack_centos
+    - name: openstack_default
+    - name: openstack_flatcar
+    - name: openstack_rhel
+    - name: openstack_rockylinux
+    - name: vsphere_centos
+    - name: vsphere_default
+    - name: vsphere_flatcar
```
xmudrii, line 107: vSphere doesn't support in-tree CCM as of 1.25.

test/tests.yml
```diff
@@ -30,760 +47,872 @@
...
-    - name: azure_centos
-    - name: azure_flatcar
-    - name: azure_rhel
-    - name: azure_rockylinux
-    - name: gce_default
+    - name: aws_centos_stable
+    - name: aws_default_stable
+    - name: aws_flatcar_stable
+    - name: aws_rhel_stable
+    - name: aws_rockylinux_stable
```
xmudrii, line 190: Please replace AWS with Azure or GCP because AWS doesn't support in-tree CCM in v1.27.

test/tests.yml
```diff
@@ -30,760 +47,872 @@
...
```
xmudrii, line 200: Same here

test/tests.yml
```diff
@@ -30,760 +47,872 @@
...
+    - name: vsphere_default_stable
+    - name: vsphere_flatcar_stable
 
-- scenario: cilium_containerd_external
-  initVersion: v1.28.3
+###########################
+# ALTERNATIVE CNI PLUGINS #
+###########################
+
+- scenario: calico_containerd
```
xmudrii, line 276: Let's remove `calico_containerd` scenario and only keep `calico_containerd_external` tests

test/tests.yml
```diff
@@ -30,760 +47,872 @@
...
+    - name: openstack_flatcar
+    - name: openstack_rhel
+    - name: openstack_rockylinux
+    - name: vsphere_centos
+    - name: vsphere_default
+    - name: vsphere_flatcar
 
-- scenario: csi_ccm_migration
-  initVersion: v1.25.15
+- scenario: cilium_containerd
```
xmudrii, line 480: Let's remove `cilium_containerd` scenario and only keep `cilium_containerd_external` tests

test/tests.yml
```diff
@@ -30,760 +47,872 @@
...
+    - name: openstack_rhel
+    - name: openstack_rockylinux
     - name: vsphere_centos
+    - name: vsphere_default
     - name: vsphere_flatcar
-    - name: gce_default
 
-- scenario: install_containerd_external
-  initVersion: v1.27.7
+- scenario: upgrade_cilium_containerd
```
xmudrii, line 650: Same here, let's only keep external tests

test/tests.yml
```diff
@@ -30,760 +47,872 @@
...
+    - name: openstack_rhel
+    - name: openstack_rockylinux
     - name: vsphere_centos
+    - name: vsphere_default
     - name: vsphere_flatcar
-    - name: gce_default
 
-- scenario: install_containerd_external
-  initVersion: v1.28.3
+- scenario: upgrade_cilium_containerd
```
xmudrii, line 685: Same here, let's only keep external tests

test/tests.yml
```diff
@@ -30,760 +47,872 @@
...
+  initVersion: v1.27.11
+  infrastructures:
+    - name: aws_long_timeout_default
+
+- scenario: conformance_containerd_external
+  initVersion: v1.28.7
+  infrastructures:
+    - name: aws_long_timeout_default
+
+- scenario: conformance_containerd
```
xmudrii, line 804: Let's use `conformance_containerd_external` here to be explicit

test/tests.yml
```diff
@@ -30,760 +47,872 @@
...
+
+- scenario: legacy_machine_controller_containerd
+  initVersion: v1.28.7
+  infrastructures:
+    - name: aws_default
+
+- scenario: legacy_machine_controller_containerd
+  initVersion: v1.29.2
+  infrastructures:
+    - name: aws_default
```
xmudrii, line 831: Replace AWS with Azure or GCP because of in-tree CCM

pkg/apis/kubeone/validation/validation.go
```diff
@@ -731,59 +728,45 @@ func ValidateHelmReleases(helmReleases []kubeoneapi.HelmRelease, fldPath *field.
...
 				allErrs = append(allErrs, field.Invalid(fldPath.Child("labels"), labelValue, "label to remove cannot have value"))
 			}
 		}
-		if gte125Constraint.Check(v) {
-			for _, taint := range h.Taints {
-				if taint.Key == "node-role.kubernetes.io/master" {
-					allErrs = append(allErrs, field.Forbidden(fldPath.Child("taints"), fmt.Sprintf("%q taint is forbidden for clusters running Kubernetes 1.25+", "node-role.kubernetes.io/master")))
-				}
-			}
-		}
```
xmudrii, line 786: I'm conflicted on this change. I would kind of prefer keeping this safeguard as long as manifests/addons tolerate taint for `node-role.kubernetes.io/master`. Once we remove this taint from manifests, removing this safeguard will be much safer. Until them, someone could try to apply this taint again for any reason (e.g. outdated KubeOneCluster manifest) and then run into problems once tolerations are removed. Also, kubeadm might be a little weird when it comes to this taint, so it further increases the risk of something going wrong.

pkg/apis/kubeone/v1beta2/defaults.go
```diff
@@ -73,6 +74,18 @@ func SetDefaults_KubeOneCluster(obj *KubeOneCluster) {
...
+func SetDefaults_CloudProvider(obj *KubeOneCluster) {
+	gteKube129Condition, _ := semver.NewConstraint(">= 1.29")
+	actualVer, err := semver.NewVersion(obj.Versions.Kubernetes)
+	if err != nil {
+		return
+	}
+
+	if gteKube129Condition.Check(actualVer) && obj.CloudProvider.None == nil {
+		obj.CloudProvider.External = true
+	}
```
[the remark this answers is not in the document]
xmudrii: @kron4eg Can you please point me to the code where this is handled? I found this https://github.com/kubermatic/kubeone/blob/c33fec9b5a81429bfe5cd9fa0eda0f123ef60c8f/pkg/templates/kubeadm/v1beta3/kubeadm.go#L274-L307 but I don't see that we handle VMware vCloud Director specifically here. I wasn't able to find any other (meaningful) place where we handle this.

PR #3074, opened 2024-03-07
WIP: Support for passing additional args to API server

### LINKED ISSUE #2275: Ability to override arbitary apiserver args
I want to be able to configure the api servers args here: https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/

Specifically I want to do service-account-jwks-uri and service-account-issuer, however I assume others would be nice too. Just ability to configure the args.

ahmedwaleedmalik:
This PR adds a new field `apiServer.additionalArgs` which can be used to pass additional arguments to the Kubernetes API server of the KubeOne cluster. 

Fixes #2275

**What type of PR is this?**

/kind feature

```release-note
Support for propagating additional arguments to the API server
```

**Documentation**:

```documentation
NONE
```


kron4eg:
I don't have objections to the technical aspect of this PR, it's all clean. However we should consider future implications that will be opened by it.

/cc @xmudrii

kron4eg:
For example, feature gates (if used) should be synchronized across the components) and we can't provide that if we have open-ended config flags

xmudrii:
I'm -1 on this PR. This has been discussed many times since KubeOne exists and we don't want to provide open-ended configuration options. We however can provide specific configuration options and expose those options via the KubeOneCluster API. Please let us know what specific option do you need and we can discuss the next steps.

ahmedwaleedmalik:
Had no idea that this has already been discussed; closing the PR :D

PR #3075, opened 2024-03-07
Update CSI feature gates

kron4eg:
Use DisableCloudProviders feature gate instead of each individual provider InTreePluginXXXUnregister

This is a follow up PR #3048

**What type of PR is this?**
/kind cleanup

```release-note
* Use DisableCloudProviders feature gate as a replacement for InTreePluginXXXUnregister for each former in-tree provider
```

**Documentation**:

```documentation
https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates-removed/
```

### DIFF
--- a/pkg/tasks/probes.go
+++ b/pkg/tasks/probes.go
@@ -53,8 +53,13 @@ const (
 
 	kubeletInitializedCMD = `test -f /etc/kubernetes/kubelet.conf`
 
-	k8sAppLabel               = "k8s-app"
+	k8sAppLabel = "k8s-app"
+
+	awsCCMAppLabelValue       = "aws-cloud-controller-manager"
+	azureCCMAppLabelValue     = "azure-cloud-controller-manager"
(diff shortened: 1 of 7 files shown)


xmudrii (approved):
/lgtm
/approve
/hold
I'm double checking how should we handle `InTreePlugin*` feature gates as of v1.29

xmudrii (approved):

pkg/tasks/probes.go
```diff
@@ -790,12 +795,11 @@ func detectCCMMigrationStatus(s *state.State) (*state.CCMStatus, error) {
 				if csiFlagRegex.MatchString(command) {
 					status.CSIMigrationEnabled = true
 				}
```
xmudrii, line 797: Set unconditionally to `true` for providers that previously had the in-tree cloud provider.

xmudrii (changes requested):

xmudrii (changes requested):

xmudrii (approved):
/lgtm
/approve

PR #3080, opened 2024-03-12
Cleanup e2e tests after introduction of kubernetes v1.29

kron4eg:
This PR is a follow up to address most of the comments from #3048

**What type of PR is this?**
/kind cleanup

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


kron4eg:
lol, seems like I've removed 1.29 too :D

test/tests.yml
```diff
@@ -183,21 +135,21 @@
...
-    - name: aws_centos_stable
-    - name: aws_default_stable
-    - name: aws_flatcar_stable
-    - name: aws_rhel_stable
-    - name: aws_rockylinux_stable
+    - name: azure_centos_stable
+    - name: azure_default_stable
+    - name: azure_flatcar_stable
+    - name: azure_rhel_stable
+    - name: azure_rockylinux_stable
```
xmudrii, line 152: This scenario is not possible because there are no in-tree CCMs in v1.29.

xmudrii (changes requested):

test/tests.yml
kron4eg: Dropped!

xmudrii (approved):
/lgtm
/approve

test/tests.yml
```diff
@@ -183,21 +135,11 @@
...
```
xmudrii, line 142: Just realized we're missing GCE here. Can you quickly add it?

xmudrii (changes requested):
/hold
@kron4eg PTAL at this comment

xmudrii (approved):
/lgtm
/approve

PR #3081, opened 2024-03-12
New API to configure TLS cipher suites for kube-apiserver, etcd and kubelet

### LINKED ISSUE #3079: New API to configure TLS cipher suites
### Problem

We use a [list of TLS cipher suites](https://github.com/kubermatic/kubeone/blob/c33fec9b5a81429bfe5cd9fa0eda0f123ef60c8f/pkg/templates/kubernetesconfigs/kubelet.go#L30-L54) for components like kube-apiserver, etcd and kubelet that are extracted from CIS benchmark, but some of those are known to be not secure.

It could be easy to switch to fully secure list if it wouldn't break clusters. Unfortunately that's not the case. For example AWS Loadbalancer healthcheck client understands only those insecure ciphers and we have to support it, otherwise it's not possible to perform HTTPS healthcheck from an AWS LB.

### Solution details

* Make TLS cipher suites configurable (per component) with the new API.
* Use lists of TLS cipher suites from the CIS benchmark as defaults.

kron4eg:
Fixes #3079

**What type of PR is this?**
/kind feature

```release-note
New API to configure TLS cipher suites for kube-apiserver, etcd and kubelet
```

**Documentation**:

```documentation
NONE
```


pkg/templates/kubernetesconfigs/kubelet.go
```diff
@@ -27,32 +27,45 @@ import (
...
-		tls.CipherSuiteName(tls.TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256),
-		tls.CipherSuiteName(tls.TLS_RSA_WITH_3DES_EDE_CBC_SHA),
-		tls.CipherSuiteName(tls.TLS_RSA_WITH_AES_128_CBC_SHA),
-		tls.CipherSuiteName(tls.TLS_RSA_WITH_AES_128_GCM_SHA256),
-		tls.CipherSuiteName(tls.TLS_RSA_WITH_AES_256_CBC_SHA),
-		tls.CipherSuiteName(tls.TLS_RSA_WITH_AES_256_GCM_SHA384),
+// This list is produces acording to CIS 1.8 / 1.2.30
+//
+// See more: https://github.com/aquasecurity/kube-bench/blob/v0.7.2/cfg/cis-1.8/master.yaml#L768-L788
+func APIServerDefaultTLSCipherSuites() []*tls.CipherSuite {
```
xmudrii, line 33: Why we don't just keep string values instead of `tls.CipherSuite`?

pkg/apis/kubeone/v1beta2/defaults.go
```diff
@@ -285,6 +288,20 @@ func SetDefaults_Features(obj *KubeOneCluster) {
...
+		obj.TLSCipherSuites.APIServer = kubernetesconfigs.TLSCipherSuites(kubernetesconfigs.APIServerDefaultTLSCipherSuites())
+	}
+
+	if obj.TLSCipherSuites.Etcd == nil {
+		obj.TLSCipherSuites.Etcd = kubernetesconfigs.TLSCipherSuites(tls.CipherSuites())
+	}
+
+	if obj.TLSCipherSuites.Kubelet == nil {
+		obj.TLSCipherSuites.Kubelet = kubernetesconfigs.TLSCipherSuites(tls.CipherSuites())
+	}
```
xmudrii, line 302: Can we reduce this to ciphers demanded by CIS? See https://github.com/aquasecurity/kube-bench/blob/abfa7d9613f0f5f9e628a3ec87fea3443fe57805/cfg/cis-1.8/node.yaml#L420-L442

xmudrii (changes requested):

xmudrii:
@kron4eg API looks good, just two minor nits

pkg/templates/kubernetesconfigs/kubelet.go
kron4eg: I actually started from []string, but after some play with API this looked more natural to make it so that `TLSCipherSuites()` can take as input both, this function's return value as well as `tls.CipherSuites()` without additional conversions. You can see it in `SetDefaults_TLSCipherSuites()`.

pkg/apis/kubeone/v1beta2/defaults.go
kron4eg: BTW: this list includes 2 insecure cipher suites: `TLS_RSA_WITH_AES_256_GCM_SHA384`, `TLS_RSA_WITH_AES_128_GCM_SHA256`.

pkg/cmd/config.go
xmudrii, line 1: This file needs to be updated again after the latest changes

xmudrii (changes requested):
One nit, otherwise LGTM

pkg/cmd/config.go
kron4eg: :+1: done

xmudrii (approved):
/lgtm
/approve

PR #3149, opened 2024-04-16
Move Azure terraform example to Standard SKU for IP addresses

kron4eg:
This fixes broken LoadBalancer functionality in azure CCM (it refuses to add Basic IPs into the LB)

Fixes #

**What type of PR is this?**
/kind regression

```release-note
[ACTION REQUIRED] The example Terraform configs for Azure have been migrated to use the Standard SKU for IP addresses. This is a breaking change for existing setups; in which case you should continue using your current SKU
```

**Documentation**:

```documentation
NONE
```


examples/terraform/azure/variables.tf
```diff
@@ -106,11 +106,6 @@ variable "disable_kubeapi_loadbalancer" {
 
 # Provider specific settings
 
-variable "ip_sku" {
-  default     = "Basic"
-  description = "SKU to use for IP addresses"
-}
```
xmudrii, line 112: We should leave this variable but default it to `Standard` so users with existing setups can eventually migrate to the new Terraform configs if they want.

examples/terraform/azure/main.tf
```diff
@@ -132,20 +118,39 @@ resource "azurerm_network_security_group" "sg" {
...
+
 resource "azurerm_public_ip" "lbip" {
   count = local.loadbalancer_count
 
   name                = "${var.cluster_name}-lbip"
   location            = var.location
   resource_group_name = azurerm_resource_group.rg.name
-  allocation_method   = "Dynamic"
-  sku                 = var.ip_sku
+  allocation_method   = "Static"
```
xmudrii, line 152: Can we also add a variable for this?

kron4eg: Problem is, it doesn't work with Basic anymore. So does it make sense to parametrize it?

Issue #3143, opened 2024-04-12
Support configuration of Service Account Issuer Discovery


JanKoppe:
### Description of the feature you would like to add / User story

As a Kubernetes cluster administrator using KubeOne
I would like to be able to configure the service account issuer discovery parameters of kube-apiserver 
in order to allow workloads to make use of AWS IRSA to authenticate against AWS APIs.

### Solution details

- Implement ServiceAccountSettings in the KubeOneCluster spec, similar to https://docs.kubermatic.com/kubermatic/v2.25/references/crds/#serviceaccountsettings
- Kubeone will hand over the ServiceAccountSettings to kubeadm with the proper arguments
- kubeadm will configure the kube-apiserver with the proper arguments (e.g. at least `--service-account-issuer`)

In total, users need to be able to adjust that `iss` value to a user-specificed value.

### Alternative approaches

I've considered modifying the kube-apiserver manifest after the cluster is up and running. This won't work though as it is a persistent manifest on the disk of the controllers. It will also be overwritten by KubeOne/kubeadm.

I've considered providing a method to pass on free-form arguments to kubeadm, but it was decided against offering something like that in the past (See https://github.com/kubermatic/kubeone/pull/3074#issuecomment-1983176830)

### Use cases

AWS provides [IAM roles for service accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) as a defined mechanic to allow workloads in EKS to fetch temporary AWS API credentials with the specific permissions of an IAM role using their service account tokens.

While AWS markets this feature for their own EKS of course, this is actually not specific to EKS. The feature allows in general any verifiable web (openid) identity to assume an IAM role by providing a valid token.

A verifiably web identity in this case can be any OpenID connect identity provider that provides at least the `.well-known/openid-configuration` endpoint and a JWKS (JSON Web Key Set).

AWS describes the method to setup such an OpenID connect identity provider in this documentation: https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_create_oidc.html

The complication that we're running into when using KubeOne is that AWS requires that the value of the issuer `iss` claim in the provided tokens must be of the same hostname that provides the `.well-known/openid-configuration` endpoint.

Kubernetes allows the configuration of that `iss` value through the `--service-account-issuer` flag for kube-apiserver (https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/).

[...]

Issue #2431, opened 2022-10-29
Support for PodTolerationRestriction


hnaderi:
### Description of the feature you would like to add / User story

As a cluster manager
I would like to dedicate nodes for some namespaces
in order to reduce blast ratio and allow fine grained resource management

### Solution details

- Use `PodNodeSelector` to automatically add `nodeSelector` to pods (Supported directly in KubeOne API)
- User `PodTolerationRestriction` to automatically add tolerations to pods (Not supported directly or indirectly, this can be done indirectly if https://github.com/kubermatic/kubeone/issues/2275 is supported)

### Alternative approaches

- Write a custom admission controller

### Use cases

Allow easy and effortless resource allocation for simple use cases

xmudrii:
This can now be done with the custom API server flags feature (#2275)
/close

PR #3162, opened 2024-04-29
Support for passing additional args to API server, controller manager, and scheduler

### LINKED ISSUE #2275: Ability to override arbitary apiserver args

ahmedwaleedmalik:
This PR adds support in KubeOne to configure additional flags and feature gates for the following components:
1. API server
2. Controller Manager
3. Scheduler

KubeOne also [configures flags and crucial feature gates](https://github.com/kubermatic/kubeone/blob/release/v1.7/pkg/templates/kubeadm/v1beta3/kubeadm.go#L80-L390) that are essential for the working of the underlying Kubernetes cluster. Due to this these flags and feature gates are always merged with what KubeOne configures. Since it's easy to misconfigure feature gates, we have prohibited users from specifying them with `flags` and instead created a dedicated `featureGates` field.

Fixes #2275 #2431
xref #3143

**What type of PR is this?**

/kind feature

Resurrection of https://github.com/kubermatic/kubeone/pull/3074

```release-note
Support for passing additional args to the API server, controller manager, and scheduler
```

**Documentation**:

```documentation
TBD
```

/assign @kron4eg


cnvergence:
I have tested it with the AWS cluster, this configuration seems like it covers most of the options and they are being applied properly.
I am not sure how I feel about allowing all of them while enabling support in these components seems to be non-breaking, fiddling with configuration files, for instance providing the wrong path for TLS cert, breaking cluster with manual cluster repair required, if you try to rollback it with KubeOne.

cnvergence (commented):
I have tested it with the AWS cluster, this configuration seems like it covers most of the options and they are being applied properly.

ahmedwaleedmalik:
> I am not sure how I feel about allowing all of them while enabling support in these components seems to be non-breaking, fiddling with configuration files, for instance providing the wrong path for TLS cert, breaking cluster with manual cluster repair required, if you try to rollback it with KubeOne.

Hey @cnvergence, this has been quite thoroughly discussed internally. We are aware of the risks, as mentioned in the documentation of the structs/fields as well. However, we don't want to block users from overriding these values since some times it's essential for their workflows.

cnvergence:
Understood, for that extensibility I am all for it, as it is a strong feature request
Just would love to have this process well explained and easier to mitigate

pkg/apis/kubeone/config/config.go
```diff
@@ -42,7 +42,8 @@ import (
 
 const (
 	// KubeOneClusterKind is kind of the KubeOneCluster object
-	KubeOneClusterKind = "KubeOneCluster"
+	KubeOneClusterKind                  = "KubeOneCluster"
+	flagsAndFeatureGateOverridesWarning = "%s only covers %s. For some features, you may also need additional configuration for other components. Use this at your own risk!!"
```
xmudrii, line 46: I think we could have more explicit message, let me see if I can come up with something.

pkg/templates/kubeadm/v1beta3/kubeadm.go
```diff
@@ -366,6 +366,7 @@ func NewConfig(s *state.State, host kubeoneapi.HostConfig) ([]runtime.Object, er
 	}
 
 	addControllerManagerNetworkArgs(clusterConfig.ControllerManager.ExtraArgs, cluster.ClusterNetwork)
+	addControlPlaneComponentsAdditionalArgs(cluster, clusterConfig)
```
xmudrii, line 369: This might be too early, I have to double check, but later functions might override some flags and feature gates.

xmudrii (commented):
This looks good in majority, I left some comments, I'll take a look at them later today and try to get this PR merged.

xmudrii (approved):
/lgtm
/approve

PR #3210, opened 2024-05-31
Initial version of v1beta3 KubeOneCluster

kron4eg:
This is a base code-infrastructure to further improve and apply out plans for v1beta3 API. It's identical to v1beta2 except docker container runtime.

**What type of PR is this?**
/kind feature
/kind chore

```release-note
Initial version of v1beta3 KubeOneCluster
```

**Documentation**:

```documentation
NONE
```


pkg/apis/kubeone/v1beta3/register.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2020 The KubeOne Authors.
+Copyright 2021 The KubeOne Authors.
```
xmudrii, line 2: We should change license headers to 2024 for v1beta3 (affects other files in this package too).

pkg/apis/kubeone/config/migrate_test.go
xmudrii, line 1: Why are tests for the migrate command removed? I think those would still be useful.

pkg/cmd/config-images.go
```diff
@@ -120,23 +120,29 @@ func listImages(opts *listImagesOpts) error {
...
-			return err
-		}
-
-		overRegGetter := images.WithOverwriteRegistryGetter(func() string {
-			if rc := conf.RegistryConfiguration; rc != nil {
-				return rc.OverwriteRegistry
+		func() {
+			var conf kubeonev1beta3.KubeOneCluster
+			if err := yaml.Unmarshal(configBuf, &conf); err != nil {
+				// we don't care if there was an error, just proceed like there was no manifest file given at all
```
xmudrii, line 126: Why is this change needed?

kron4eg: Initial idea was that "since we don't care if file exists, maybe we don't care about the unmarshal error ether". But yeah, those are different cases. Returned back as it was.

pkg/apis/kubeone/v1beta3/register.go
kron4eg: Fixed

xmudrii (approved):
/lgtm
/approve

Issue #3206, opened 2024-05-23
Update images to support Kubernetes 1.30


xmudrii:
Reference to the previous issue for updating images: 

Action items:

- [ ] Update the issue template to add/remove [images](https://github.com/kubermatic/kubeone/blob/main/pkg/templates/images/images.go) as appropriate

The following components/images should be updated:

### General

- [x] [Canal CNI](https://github.com/projectcalico/calico) (#3216)
- [ ] [Calico VXLAN CNI](https://github.com/projectcalico/calico) 
- [x] [Cilium CNI](https://github.com/cilium/cilium) (#3214)
- [x] [NodeLocalDNS](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml) (#3214)
- [x] [metrics-server](https://github.com/kubernetes-sigs/metrics-server) (#3209)
- [x] [Cluster Autoscaler](https://github.com/kubernetes/autoscaler) (#3214)

### Cloud provider components

- [x] [AWS CCM](https://github.com/kubernetes/cloud-provider-aws) (#3214)
- [x] [AWS CSI](https://github.com/kubernetes-sigs/aws-ebs-csi-driver) (#3214)
- [x] [Azure CCM](https://github.com/kubernetes-sigs/cloud-provider-azure) (#3214)
- [x] [AzureDisk CSI](https://github.com/kubernetes-sigs/azuredisk-csi-driver) (#3214)
- [x] [AzureFile CSI](https://github.com/kubernetes-sigs/azurefile-csi-driver) (#3214)
- [x] [DigitalOcean CCM](https://github.com/digitalocean/digitalocean-cloud-controller-manager) (#3214)
- [x] [DigitalOcean CSI](https://github.com/digitalocean/csi-digitalocean) (#3214)
- [x] [EquinixMetal CCM](https://github.com/equinix/cloud-provider-equinix-metal) (already the latest)
- [ ] [GCP CCM](https://github.com/kubernetes/cloud-provider-gcp)  
- [x] [GCP Compute Persistent Disk CSI](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver) (#3214)
- [x] [Hetzner CCM](https://github.com/hetznercloud/hcloud-cloud-controller-manager) (already the latest)
- [x] [Hetzner CSI](https://github.com/hetznercloud/csi-driver) (#3214)
[...]

Issue #3205, opened 2024-05-23
Support Kubernetes 1.30


xmudrii:
This is a collector issue for Kubernetes 1.30 support in KubeOne. The following tasks should be taken care of:

* [x] Check the Kubernetes changelog to ensure there are no breaking changes and removals
* [x] Update the `build` image to update Sonobuoy (and other dependencies if needed)
* [x] Update the latest supported Kubernetes version in [the API validation](https://github.com/kubermatic/kubeone/blob/main/pkg/apis/kubeone/validation/validation.go#L40-L41) (#3209)
* [x] Update [default admission controllers](https://github.com/kubermatic/kubeone/blob/main/pkg/kubeflags/data.go) if needed (not needed)
* [x] Update `pause` image version in `pkg/apis/kubeone`  if needed (not needed)
* [x] Update [the stable version marker in Makefile](https://github.com/kubermatic/kubeone/blob/5273f9a372736569c6b09b38f2959019d29e4d6a/Makefile#L24) (#3209)
* [x] Add E2E tests (#3209)
* [x] Update daily periodics to use the latest Kubernetes release
* [x] Update [the Compatibility Matrix](https://docs.kubermatic.com/kubeone/main/architecture/compatibility/supported-versions/) (https://github.com/kubermatic/docs/pull/1697)
* [x] Create an issue to track updating [images](https://github.com/kubermatic/kubeone/blob/main/pkg/templates/images/images.go) (#3206)

xmudrii:
Not all tasks are done
/reopen

PR #3234, opened 2024-06-18
Improve the Kubernetes 1.30 support

xmudrii:

This PR introduces several improvements and bug fixes to our Kubernetes 1.30 support:

- Set `cri-tools` version for Kubernetes 1.30 on Flatcar
- Drop `--certificate-renew=true` and `--yes` flags from `kubeadm upgrade apply` command as it's not possible to combine the `--config` flag with other flags as of `kubeadm` 1.30
  - Dropping the `--certificate-renew=true` flag is a no-op change because the flag is `true` by default
  - Dropping the `--yes` flag means that we run `kubeadm` in the interactive mode and that `kubeadm` will ask user to type `yes` to confirm upgrading the cluster. As KubeOne doesn't support interactive inputs, we mitigate this by piping `yes` to the `kubeadm upgrade apply` command
- Fix Rocky Linux tests failing by using `t3.nano` instead of `t3a.nano` for the bastion instance
  - It appears that the latest Rocky Linux AMI doesn't work on `t3a.nano` any longer, the instance is stuck in initializing and it's not possible to SSH to it

Additionally, this PR does some bookkeeping work:

- Add an entry for setting `cri-tools` version for a new Kubernetes version on Flatcar
- Update the `Update images` template with the latest changes

xref #3205 #3206

**What type of PR is this?**

/kind bug
/kind cleanup

```release-note
KubeOne now runs `kubeadm upgrade apply` without `--certificate-renewal=true` and `--yes` flags. This change should not have any effect to the upgrade process, but if you discover any issue, please create a new issue in the KubeOne repository
```

**Documentation**:
```documentation
NONE
```

/assign @kron4eg @xrstf

### DIFF
--- a/.github/ISSUE_TEMPLATE/update-images.md
+++ b/.github/ISSUE_TEMPLATE/update-images.md
@@ -19,18 +19,23 @@ Reference to the previous issue for updating images: <!-- (issue reference) -->
 
 Action items:
 
-- [ ] Update the issue template to add/remove [images](https://github.com/kubermatic/kubeone/blob/main/pkg/templates/images/images.go) as appropriate
+- [ ] Update the issue template to add/remove [images](https://github.com/kubermatic/kubeone/blob/main/pkg/templates/images/images.go) as appropriate <!-- (PR reference|already the latest) -->
 
 The following components/images should be updated:
 
 ### General
 
+- [ ] [machine-controller](https://github.com/kubermatic/machine-controller) <!-- (PR reference|already the latest) -->
+- [ ] [operating-system-manager](https://github.com/kubermatic/operating-system-manager)
+- [ ] [metrics-server](https://github.com/kubernetes-sigs/metrics-server) <!-- (PR reference|already the latest) -->
+- [ ] [NodeLocalDNS](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml) <!-- (PR reference|already the latest) -->
+- [ ] [Cluster Autoscaler](https://github.com/kubernetes/autoscaler) <!-- (PR reference|already the latest) -->
+
+### CNIs
+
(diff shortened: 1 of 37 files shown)


.github/ISSUE_TEMPLATE/update-images.md
```diff
@@ -19,18 +19,23 @@ Reference to the previous issue for updating images: <!-- (issue reference) -->
...
+- [ ] [machine-controller](https://github.com/kubermatic/machine-controller) <!-- (PR reference|already the latest) -->
+- [ ] [operating-system-manager](https://github.com/kubermatic/operating-system-manager)
+- [ ] [metrics-server](https://github.com/kubernetes-sigs/metrics-server) <!-- (PR reference|already the latest) -->
+- [ ] [NodeLocalDNS](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml) <!-- (PR reference|already the latest) -->
+- [ ] [Cluster Autoscaler](https://github.com/kubernetes/autoscaler) <!-- (PR reference|already the latest) -->
+
+### CNIs
+
 - [ ] [Canal CNI](https://github.com/projectcalico/calico) <!-- (PR reference|already the latest) -->
 - [ ] [Calico VXLAN CNI](https://github.com/projectcalico/calico) <!-- (PR reference|already the latest) -->
```
SimonTheLeg, line 37: Should Calico VXLAN CNI be removed from the list? Iirc we discussed in Slack that this is not required for updates anymore

SimonTheLeg (commented):
small question from my side

.github/ISSUE_TEMPLATE/update-images.md
xmudrii: I left it as of now until we don't remove the addon itself.

PR #3243, opened 2024-06-18
Fix local helm releases deletion issue

### LINKED ISSUE #3211: Support of installing local fs charts
Helmrelease feature should not uninstall existing installed charts while `kubeone apply` again in case of chart are present on local filesystem. This does not happen with charts applied via their repoURL(remote charts).

we could read the chart from the local FS (if it's a file), and check for read name of the chart.

mohamed-rafraf:

`releasesFilterFn` is used to filter out the helm releases that will be uninstalled. When using local helm charts, the helm releases can be deleted without any releases after applying `kubeone apply` again because we are comparing the Chart Name with the path of the Chart (`HelmRelease.Chart` is considered a chart path for local usage).

This PR will allow kubeone to correctly read the chart name for local and remote helm charts!  

Fixes #3211

**What type of PR is this?**
/kind bug

```release-note
Fix local helm chart releases deletion issue
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/localhelm/helm3.go
+++ b/pkg/localhelm/helm3.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path"
 	"sort"
 
 	"github.com/google/go-cmp/cmp"
@@ -37,6 +38,7 @@ import (
 	"helm.sh/helm/v3/pkg/registry"
 	helmrelease "helm.sh/helm/v3/pkg/release"
 	"helm.sh/helm/v3/pkg/storage/driver"
+	"sigs.k8s.io/yaml"
 
 	kubeoneapi "k8c.io/kubeone/pkg/apis/kubeone"
 	"k8c.io/kubeone/pkg/fail"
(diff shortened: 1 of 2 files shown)


pkg/localhelm/helm3.go
```diff
@@ -169,7 +170,7 @@ func Deploy(st *state.State) error {
 func releasesFilterFn(helmReleases []kubeoneapi.HelmRelease, logger logrus.FieldLogger) func(rel *helmrelease.Release) bool {
 	return func(rel *helmrelease.Release) bool {
 		for _, hr := range helmReleases {
-			if rel.Name == hr.ReleaseName && rel.Namespace == hr.Namespace && rel.Chart.Name() == hr.Chart {
+			if rel.Name == hr.ReleaseName && rel.Namespace == hr.Namespace && rel.Chart.Name() == path.Base(hr.Chart) {
```
kron4eg, line 173: Unfortunatelly this is not enough. Directory name can be different from chart name, the actual chart name contained in the Chart.yaml file, which we need to YAML unmarshall in case when chart is specified as a directory path.

kron4eg (changes requested):
Some more logic is needed

PR #3202, opened 2024-05-21
Workloads API

kron4eg:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #3201

**What type of PR is this?**
/kind feature

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change? Then add your Release Note here**:

```release-note
Workloads API
```

**Documentation**:

```documentation
TBD
```

Issue #2708, opened 2023-03-03
UX - deploy default storage classes by default


toschneck:
### Description of the feature you would like to add / User story

As a user
I would like to have mostly functional cluster with the default storage classes
in order to don't lookup docu to get default storage.

### Solution details

Enable by default `default-storage-class` addon similar as in KKP

PR #3248, opened 2024-06-20
New addons v1beta3 API

kron4eg:
This is a replacement for #3202

**What type of PR is this?**
/kind feature
/kind api-change

```release-note
New addons v1beta3 API
```

**Documentation**:

```documentation
TBD
```


pkg/apis/kubeone/validation/validation_test.go
```diff
@@ -2090,37 +2089,18 @@ func TestValidateAddons(t *testing.T) {
...
+			name:          "no path set and no embedded addons specified",
+			addons:        &kubeoneapi.Addons{},
 			expectedError: false,
 		},
 		{
-			name: "valid addons config (disabled)",
-			addons: &kubeoneapi.Addons{
-				Enable: false,
-			},
+			name:          "valid addons config (disabled)",
```
xmudrii, line 2102: This is the same case as `no path set and no embedded addons specified`, we can remove this one.

test/e2e/testdata/containerd_calico_external.yaml
```diff
@@ -15,7 +12,8 @@ cloudProvider:
   external: true
 
 addons:
-  enable: true
   addons:
-  - name: default-storage-class
-  - name: calico-vxlan
+    - addon:
+        name: default-storage-class
```
xmudrii, line 17: We should enable this addon by default in v1beta3.

pkg/apis/kubeone/types.go
```diff
@@ -1003,19 +999,22 @@ type Addon struct {
 	Delete bool `json:"delete,omitempty"`
 }
 
+type AddonRef struct {
+	// KubeOne's internal Addon
```
xmudrii, line 1003: This should be a valid godoc comment, maybe something like `Addon is a built-in addons mechanism`

pkg/apis/kubeone/config/migrate.go
```diff
@@ -17,45 +17,72 @@ limitations under the License.
...
+		return nil, fail.Runtime(err, "testing unmarshal v1beta2 KubeOneCluster")
+	}
+
+	if err = scheme.Scheme.Convert(oldManifest, internalManifest, kubeoneapi.SchemeGroupVersion); err != nil {
+		return nil, fail.Config(err, "converting v1beta2/KubeOneCluster into internal KubeOneCluster")
+	}
+
+	if err = scheme.Scheme.Convert(internalManifest, newManifest, kubeonev1beta3.SchemeGroupVersion); err != nil {
+		return nil, fail.Config(err, "converting internal KubeOneCluster into v1beta3/KubeOneCluster")
+	}
```
xmudrii, line 83: I strongly believe we should use `yamled` here instead. There are many problems with this implementation, the first one being the fact that comments are not preserved, the second one that the file will be alphabetically sorted (instead of leaving sort as it is). This is not acceptable for a user-facing command, I'm strong -1 on this change.

pkg/apis/kubeone/v1beta2/conversion.go
```diff
@@ -30,3 +53,23 @@ func Convert_kubeone_KubeOneCluster_To_v1beta2_KubeOneCluster(in *kubeoneapi.Kub
...
+		DisableTemplating: in.DisableTemplating,
+		Params:            in.Params,
+		Delete:            in.Delete,
+	}
+
+	return nil
+}
+
+func Convert_kubeone_AddonRef_To_v1beta2_Addon(*kubeoneapi.AddonRef, *Addon, conversion.Scope) error {
+	// TODO: do something ?
```
xmudrii, line 69: Do we still need this TODO?

xmudrii (changes requested):

test/e2e/testdata/containerd_calico_external.yaml
kron4eg: yes, in a dedicated PR. There is #2708 for this case.

pkg/apis/kubeone/config/migrate.go
kron4eg: yamled does not preserve comment ether

xmudrii: I see, but at least we get to preserve the key order and the key set, so I believe we still have huge benefits. Thanks for addressing this issue!

pkg/cmd/config.go
```diff
@@ -45,6 +46,9 @@ const (
 	defaultCloudProviderName = "aws"
 )
 
+//go:embed example-manifest.tmpl.yaml
+var exampleManifest string
```
xmudrii, line 50: FYI: `kubeone config print -f` fails with:
```
Error: runtime: testing marshal/unmarshal
error unmarshaling JSON: while decoding JSON: json: unknown field "podSecurityPolicy"
```

pkg/apis/kubeone/config/migrate.go
```diff
@@ -17,45 +17,113 @@ limitations under the License.
...
 	"k8c.io/kubeone/pkg/yamled"
+
+	kyaml "sigs.k8s.io/yaml"
 )
 
-// MigrateOldConfig migrates KubeOneCluster v1beta1 object to v1beta2
-func MigrateOldConfig(clusterFilePath string) (interface{}, error) {
-	oldConfig, err := loadClusterConfig(clusterFilePath)
+// MigrateV1beta2V1beta3 migrates KubeOneCluster v1beta2 object to v1beta3
+func MigrateV1beta2V1beta3(clusterFilePath string) ([]byte, error) {
```
xmudrii, line 37: We should add some unit tests for this command.

xmudrii (changes requested):

pkg/cmd/config.go
kron4eg: Yeah, after rebasing and getting podSecurityPolicy removed we got the error. fixed now.

pkg/apis/kubeone/config/migrate.go
kron4eg: done

xmudrii (approved):
/lgtm
/approve
:shipit:

PR #3251, opened 2024-06-20
Refactor the upgrade process to stop using config on upgrade

xmudrii:

This PR refactors the cluster upgrade process to adhere to recommendations from [the upstream community](https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-reconfigure/). With that, using the `--config` flag along with the `kubeadm upgrade apply` command:

- is strictly discouraged because of possible side effects affecting the cluster stability
- is deprecated and partially removed as of Kubernetes 1.30 and will be completely removed with Kubernetes 1.31

`kubeadm` writes configuration to different ConfigMaps in the `kube-system` namespace, mainly `kubeadm-config`, `kubelet-config`, and `kube-proxy`. The recommended workflow is to update the needed ConfigMap and then run the appropriate `kubeadm` command as described in the document.

Upon initiating the upgrade process, KubeOne will build the Kubernetes clientset, generate new Kubeadm configs and update all the ConfigMaps used by Kubeadm. Kubeadm will reconfigure the cluster as needed based on what's changed (if anything is changed at all).

This however means that some options are now immutable, but most of them should not be user-facing, and I don't think we ever supported changing these options:

- Anything in `InitConfiguration` and `JoinConfiguration` structs. There is `.LocalAPIEndpoint.AdvertiseAddress`, but I don't think we support changing the IP address of existing control plane or worker nodes

xref #3205 

**What type of PR is this?**

/kind feature
/kind design

In the future, we should refactor the force upgrade process to work in a reconciliation manner, i.e. to avoid running the full upgrade process, but only compare what's changed and regenerate only those affected manifests.

```release-note
Refactor the cluster upgrade process to adhere to the [Kubernetes recommendations](https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-reconfigure/) by updating ConfigMaps used by Kubeadm instead of providing the full config to Kubeadm itself. This change should not have any effect to cluster upgrades, but if you encounter any issue, please create an issue in our repository
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/pkg/templates/kubeadm/v1beta3/kubeadm.go
+++ b/pkg/templates/kubeadm/v1beta3/kubeadm.go
@@ -77,8 +77,17 @@ var (
 	fixedEtcd129Constraint = semverutil.MustParseConstraint(">= 1.29")
 )
 
+type Config struct {
+	InitConfiguration    *kubeadmv1beta3.InitConfiguration
+	JoinConfiguration    *kubeadmv1beta3.JoinConfiguration
+	ClusterConfiguration *kubeadmv1beta3.ClusterConfiguration
+
+	KubeletConfiguration   runtime.Object
+	KubeProxyConfiguration runtime.Object
+}
+
 // NewConfig returns all required configs to init a cluster via a set of v1beta3 configs
-func NewConfig(s *state.State, host kubeoneapi.HostConfig) ([]runtime.Object, error) {
+func NewConfig(s *state.State, host kubeoneapi.HostConfig) (*Config, error) {
 	cluster := s.Cluster
 	kubeSemVer, err := semver.NewVersion(cluster.Versions.Kubernetes)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


pkg/templates/kubeadm/v1beta3/kubeadm.go
```diff
@@ -78,17 +78,24 @@ var (
 )
 
 // NewConfig returns all required configs to init a cluster via a set of v1beta3 configs
-func NewConfig(s *state.State, host kubeoneapi.HostConfig) ([]runtime.Object, error) {
+func NewConfig(s *state.State, host kubeoneapi.HostConfig) (
+	initConfig,
+	joinConfig,
+	clusterConfig,
+	kubeletConfig,
+	kubeproxyConfig runtime.Object,
```
xmudrii, line 86: I'm wondering if we should instead use a struct that contains all these objects.

kron4eg: Totally, such a big number of return values reads like a mess.

kron4eg (commented):
All in all, I like the idea.

pkg/templates/kubeadm/v1beta3/kubeadm.go
xmudrii: @kron4eg Refactored into structs.

PR #3271, opened 2024-06-26
Fix openstack CCM addon

kron4eg:
on cluster upgrade we check if openstack ccm is present and kubeone want to find it using k8s-app label selector.

* add proper label that kubeone expects CCM to have in order to locate it
* automate some routinte like image and annotations

**What type of PR is this?**
/kind bug

```release-note
Fix openstack CCM addon
```

**Documentation**:

```documentation
NONE
```


xmudrii (approved):
/lgtm
/approve

Issue #3228, opened 2024-06-14
Test Release KubeOne 1.8.1 - Test Cluster Provisioning on VMware Cloud Director


xmudrii:
This requires manual testing

kron4eg:
Cluster provisioning VMware Cloud Director is successful

Issue #2057, opened 2022-05-26
Investigate and add External CCM for VMware Cloud Director


ahmedwaleedmalik:

kron4eg:
At last we can close this, as CCM is now part of KubeOne v1.11.

PR #3277, opened 2024-06-27
VCD: Add bastion host to terraform configs

ahmedwaleedmalik:
This PR adds a bastion host for the VCD terraform provider. 

Currently, we have a limitation "that there is only 1 public IP in our VDC and that cannot be used for e2e tests since it's attached to the edge gateway" as stated in https://github.com/kubermatic/kubeone/issues/2056. We have a single edge gateway that is shared within our organization on VCD. Hence, using this public IP to further expose resources like bastion VMs, kubeapi server, etc., using NAT is not ideal and can result in conflicts. 

With that said, I'm not aware of the current state of our VCD infra as reportedly NSX-T has been enabled for our organization there. Having NSX-T should fix the limitation of using edge gateway's public IP and there should be some pool of public IPs that we can then use. Anyways, due to a lack of knowledge, I'll leave this up to the KubeOne maintainers, or whoever ends up working on this ticket to figure this out.

This should unblock KubeOne maintainers from continuing with https://github.com/kubermatic/kubeone/issues/3228, https://github.com/kubermatic/kubeone/issues/2056, and https://github.com/kubermatic/kubeone/issues/2057, etc., without being dependent on me.

I've also added a [sample](https://github.com/kubermatic/kubeone/blob/bc7e9290a5866493b0dbba5efc9046dc4a28adef/test/e2e/testdata/vmware-cloud-director.tfvars) `tfvars` file that documents the values that can be used for this provider and can be later used in the E2E tests as well.

Fixes #

**What type of PR is this?**

/kind feature

```release-note
VCD: Add bastion host support to terraform configs
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/examples/terraform/vmware-cloud-director/output.tf
+++ b/examples/terraform/vmware-cloud-director/output.tf
@@ -23,6 +23,10 @@ output "kubeone_api" {
   }
 }
 
+output "ssh_commands" {
+  value = var.enable_bastion_host ? formatlist("ssh -J ${var.ssh_username}@${local.external_network_ip}:${var.bastion_host_ssh_port} ${var.ssh_username}@%s", vcd_vapp_vm.control_plane.*.network.0.ip) : formatlist("ssh ${var.ssh_username}@%s", vcd_vapp_vm.control_plane.*.network.0.ip)
+}
+
 output "kubeone_hosts" {
   description = "Control plane endpoints to SSH to"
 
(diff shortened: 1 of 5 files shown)


examples/terraform/vmware-cloud-director/output.tf
```diff
@@ -23,6 +23,10 @@ output "kubeone_api" {
   }
 }
 
+output "ssh_commands" {
+  value = var.enable_bastion_host ? formatlist("ssh -J ${var.ssh_username}@${local.external_network_ip}:${var.bastion_host_ssh_port} ${var.ssh_username}@%s", vcd_vapp_vm.control_plane.*.network.0.ip) : formatlist("ssh ubuntu@%s", vcd_vapp_vm.control_plane.*.network.0.ip)
```
kron4eg, line 27: Instead of ubuntu it should be var.ssh_username

ahmedwaleedmalik: Good point, fixed 👍

PR #3301, opened 2024-07-09
Upgrade followers using individual tasks

kron4eg:
For a precise retry if a task fails.

This helps to avoid followers upgrade procedure from the very beginning in case when it fails on the second follower.

Fixes #

**What type of PR is this?**
/kind flake

```release-note
Upgrade followers using individual tasks
```

**Documentation**:

```documentation
NONE
```


xmudrii (approved):
/lgtm
/approve

PR #3305, opened 2024-07-16
Allow upstream cluster autoscaler flags configuration

archups:
This PR is to allow the configuration of the upstream cluster-autoscaler supported flags via KubeOne provided cluster-autoscaler addon to address https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#my-cluster-is-below-minimum--above-maximum-number-of-nodes-but-ca-did-not-fix-that-why and https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#im-running-cluster-with-nodes-in-multiple-zones-for-ha-purposes-is-that-supported-by-cluster-autoscaler issue as faced at the one of the customer end. 

Fixes #

**What type of PR is this?**
/kind feature

```release-note
Allow the configuration of the upstream cluster-autoscaler flags  `--enforce-node-group-min-size` and `--balance-similar-node-groups`
```

**Documentation**:

```documentation
TBD
```

### DIFF
--- a/addons/cluster-autoscaler/README.md
+++ b/addons/cluster-autoscaler/README.md
@@ -117,14 +117,20 @@ You need to replace the following values with the actual ones:
 
 * `CLUSTER_AUTOSCALER_IMAGE` can be used to replace the Cluster Autoscaler image
   * The minor versions of Cluster Autoscaler and Kubernetes cluster should
-    match, as per
-    [Cluster Autoscaler recommendations][recommended-autoscaler-versions]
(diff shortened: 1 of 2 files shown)


kron4eg:
@archups it's not a bug it's a feature :joy: 

/approve
/lgtm

archups:
> @archups it's not a bug it's a feature 😂
> 
> /approve /lgtm

Oops, I was confused whether to add as feature or bug..  Thanks @kron4eg  for the clarification and review. :)

Issue #2675, opened 2023-02-20
Replace DynamicAuditLog with the webhook backend


xmudrii:
### Description of the feature you would like to add / User story

The DynamicAuditLog feature got removed in Kubernetes 1.19, but we still offer it via our API: https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates-removed/

We should consider replacing it with a webhook backend which is considered a replacement for the old feature: https://v1-23.docs.kubernetes.io/docs/tasks/debug/debug-cluster/audit/#webhook-backend

### Solution details

- DynamicAuditLog feature is disallowed in the API
- We support the webhook backend as an alternative

PR #3307, opened 2024-07-19
Add support for kube-apiserver WebhookAuditLog

SimonTheLeg:

Adds support for using the k8s api-server webhook backend for auditing https://kubernetes.io/docs/tasks/debug/debug-cluster/audit/#webhook-backend

Targets first part of #2675. Note: This PR does not remove the old DynamicWebHook yet, which is required for the issue to fully close, but it makes sense to have these in two separate PRs.

**What type of PR is this?**

/kind feature

* it is possible to both use StaticAuditLog (via file) and WebHookLog at the same time, so I have omitted any check in validation that would only pass if a singular one is being used
* ~there are a lot of very specific flags (see TODO marker in types.go). Do we want to make them available to end-users?~ => done we decided to implement all
* coding-style question: the variable `auditPolicyFileFlag` is shared between activateubeadmWebHookAuditLogs and activateKubeadmStaticAuditLogs in two different files within the same package. Is re-using it the preferred way or should I create a shared addAuditPolicyFileFlag func in a separate file which is then used by both activate-funcs?

```release-note
Add audit support for k8s api-server webhook backend
```

**Documentation**:

```documentation
TBD
```
(I am thinking for documentation it makes sense to write a little article demonstrating how to do it, especially providing a sample for the config file, since it's explained pretty confusingly inside the official kubernetes documentation)

Has been discussed in sig-custer-mgmt, detailed documentation will follow:
~There are a lot of gotchas with this feature we need to document super thoroughly:~
~1. Because kube-apiserver pod is [not able to resolve cluster internal dns](https://github.com/kubernetes/minikube/issues/3772), you cannot use kubernetes svc dns as a parameter for the webhook target. But of course we cannot dynamically change and reload the webhook config with updated info how to find the webhook receiver. Therefore possibly the only way to expose this is making use of NodePort services with a fixed NodePort to expose the webhook receiver and then set `http://localhost:{aforementioned fixed port}` inside the webhook config for api-server~
~2. The effects described in 1 are especially severe if you set `--audit-webhook-mode=blocking` in the api-server flags. This leads to an interesting chicken-and-egg problem: kube-apiserver won't start because it cannot contact the webhook and eventually stops processing requests, but the webhook receiver can also not start, because the api-server is not ready. As a result I am very sceptical if we should make this feature available~
~3. All of this of course does not apply if you have your webhook receiver outside of the cluster.~


pkg/apis/kubeone/types.go
```diff
@@ -935,6 +938,43 @@ type StaticAuditLogConfig struct {
 	LogMaxSize int `json:"logMaxSize,omitempty"`
 }
 
+type WebHookAuditLog struct {
```
xmudrii, line 941: The API type should be called `WebhookAuditLog` (with lowercase `h`). That's also to make it consistent with different Kubernetes resources, e.g. `MutatingAdmissionWebhook` and `ValidatingAdmissionWebhook`

xmudrii (commented):
I'll review the PR once it's not WIP anymore, but this should be fixed beforehand.

SimonTheLeg:
aright, after quite a lot of testing, I think I finally have it figured out.

Using the following steps, this works:

1. Create machines using the AWS terraform example
2. Create the cluster using the following files

```yaml
# kubeone.yaml
apiVersion: kubeone.k8c.io/v1beta3
kind: KubeOneCluster
versions:
  kubernetes: “1.30.2”

features:
  webhookAuditLog:
    enable: true
    config:
      policyFilePath: policy.yaml
      configFilePath: config.yaml
```

```yaml
# policy.yaml
apiVersion: audit.k8s.io/v1 # This is required.
kind: Policy
# Don’t generate audit events for all requests in RequestReceived stage.
omitStages:
  - “RequestReceived”
rules:
  # Log pod changes at RequestResponse level
  - level: RequestResponse
    resources:
      - group: “”
        # Resource “pods” doesn’t match requests to any subresource of pods,
        # which is consistent with the RBAC policy.
        resources: [“pods”, “deployments”]

  - level: RequestResponse
    resources:
      - group: “rbac.authorization.k8s.io”
        # Resource “pods” doesn’t match requests to any subresource of pods,
        # which is consistent with the RBAC policy.
        resources: [“clusterroles”, “clusterrolebindings”]

  # Log “pods/log”, “pods/status” at Metadata level
  - level: Metadata
    resources:
      - group: “”
        resources: [“pods/log”, “pods/status”]

  # Don’t log requests to a configmap called “controller-leader”
  - level: None
    resources:
      - group: “”
        resources: [“configmaps”]
        resourceNames: [“controller-leader”]

  # Don’t log watch requests by the “system:kube-proxy” on endpoints or services
  - level: None
    users: [“system:kube-proxy”]
    verbs: [“watch”]
    resources:
      - group: “” # core API group
        resources: [“endpoints”, “services”]

  # Don’t log authenticated requests to certain non-resource URL paths.
  - level: None
    userGroups: [“system:authenticated”]
    nonResourceURLs:
      - “/api*” # Wildcard matching.
      - “/version”

  # Log the request body of configmap changes in kube-system.
  - level: Request
    resources:
      - group: “” # core API group
        resources: [“configmaps”]
    # This rule only applies to resources in the “kube-system” namespace.
    # The empty string “” can be used to select non-namespaced resources.
    namespaces: [“kube-system”]

  # Log configmap changes in all other namespaces at the RequestResponse level.
  - level: RequestResponse
    resources:
      - group: “” # core API group
        resources: [“configmaps”]

  # Log secret changes in all other namespaces at the Metadata level.
  - level: Metadata
    resources:
      - group: “” # core API group
        resources: [“secrets”]

  # Log all other resources in core and extensions at the Request level.
  - level: Request
    resources:
      - group: “” # core API group
      - group: “extensions” # Version of group should NOT be included.

  # A catch-all rule to log all other requests at the Metadata level.
  - level: Metadata
    # Long-running requests like watches that fall under this rule will not
    # generate an audit event in RequestReceived.
    omitStages:
      - “RequestReceived”
```

```yaml
# config.yaml
apiVersion: v1
kind: Config
clusters:
- name: falco
  cluster:
    # certificate-authority: /path/to/ca.crt # for https
    server: http://localhost:31000 # this must match later what is configured as a k8s service for the backend
contexts:
- context:
    cluster: falco
    user: “”
  name: default-context
current-context: default-context
preferences: {}
users: []
```

3. To ensure that it works, I installed an echo server as a dummy backend
```bash
k create deploy debug-echo-server --image=bfosberry/http-logger
```

And then apply the following service
```yaml
# service.yaml
apiVersion: v1
kind: Service
metadata:
  labels:
    app: debug-echo-server
  name: debug-echo-server
  namespace: debug
spec:
  ports:
    - nodePort: 31000 # this needs to be fixed port and match what's defined in the webhook config on apiserver
      port: 80
      protocol: TCP
      targetPort: 8090
  selector:
    app: debug-echo-server
  type: NodePort
```

4. We can then simulate an event using
```shell
kubectl create cm  myconfigmap —from-literal=username=admin —from-literal=password=123456
```

And see a big json sent to the echo-server eventually. Note: kube-apiserver uses buffering and batching, so you might need to repeat the step or wait for other cluster audit events to occur, before it shows up as a logline in the echo-server,

Here is a shortened and json-prettified output.
```log
          {
            "metadata": {
              "name": "some-configmap",
              "namespace": "default",
              "uid": "2ef588f4-5802-460a-a8cf-b24596f96410",
              "resourceVersion": "315636",
              "creationTimestamp": "2024-07-22T20:06:33Z",
              "managedFields": [
                {
                  "manager": "kubectl-create",
                  "operation": "Update",
                  "apiVersion": "v1",
                  "time": "2024-07-22T20:06:33Z",
                  "fieldsType": "FieldsV1",
                  "fieldsV1": {
                    "f:data": {
                      ".": {},
                      "f:password": {},
                      "f:username": {}
                    }
                  }
                }
              ]
            },
            "data": {
              "password": "123456",
              "username": "admin"
            }
          },
```

xmudrii (changes requested):

SimonTheLeg:
@xmudrii 
I have added godoc. Regarding the enum, I was not 100% sure, since go does not have first-class enums, but I think this is kind of the way to do it for this case? ptal.

Also one more small thing. coding-style question: the variable auditPolicyFileFlag is shared between activateubeadmWebHookAuditLogs and activateKubeadmStaticAuditLogs in two different files within the same package. Is re-using it the preferred way or should I create a shared addAuditPolicyFileFlag func in a separate file which is then used by both activate-funcs?

docs/api_reference/v1beta3.en.md
```diff
@@ -829,3 +835,67 @@ WeaveNetSpec defines the WeaveNet CNI plugin
...
+
+[Back to Group](#v1beta3)
+
+### WebHookAuditLogThrottleConfig
+
+
+
+| Field | Description | Scheme | Required |
+| ----- | ----------- | ------ | -------- |
+| disable | Disable disables webhook throttling. Defaults to false, which corresponds to kube-apiservers default of enabling throttling. | bool | false |
```
xmudrii, line 858: @kron4eg For a follow up, we have in one struct `disable`, but in other `enable`. I think we should standardize this, we could throw in bool pointers, and use it to properly determine if the value is set or not. Slight inconvenience for us code wise, but makes it easier for users.

pkg/apis/kubeone/types.go
```diff
@@ -923,6 +926,95 @@ type StaticAuditLogConfig struct {
...
+type WebhookAuditLog struct {
+	// Enable
+	// Default value is false.
+	Enable bool `json:"enable,omitempty"`
+
+	// Config
+	Config WebhookAuditLogConfig `json:"config"`
+}
+
+type WebhookMode string
```
xmudrii, line 938: For a follow up, maybe rename this to `WebhookAuditLogMode`

pkg/apis/kubeone/types.go
```diff
@@ -923,6 +926,95 @@ type StaticAuditLogConfig struct {
...
+
+const (
+	Batch          WebhookMode = "batch"
+	Blocking       WebhookMode = "blocking"
+	BlockingStrict WebhookMode = "blocking-strict"
```
xmudrii, line 943: For a follow up:
```suggestion
	BatchWebhookAuditLogMode          WebhookAuditLogMode = "batch"
	BlockingWebhookAuditLogMode       WebhookAuditLogMode = "blocking"
	BlockingStrict WebhookAuditLogMode WebhookAuditLogMode = "blocking-strict"
```
or something like that

pkg/apis/kubeone/types.go
```diff
@@ -923,6 +926,95 @@ type StaticAuditLogConfig struct {
...
+
+	// ConfigFilePath is a path on local file system to a kubeconfig formatted file that
+	// defines how kube-apiserver can connect to the audit webhook.
+	// ConfigFilePath is a required field.
+	// More info: https://kubernetes.io/docs/tasks/debug/debug-cluster/audit/#webhook-backend
+	ConfigFilePath string `json:"configFilePath"`
+
+	// InitialBackOff defines the amount of time to wait before retrying the first failed request.
+	// Defaults to 10s.
+	InitialBackOff metav1.Duration `json:"initialBackOff,omitempty"`
```
xmudrii, line 961: @kron4eg For a follow up: I'm not the biggest fan of zero value resulting in a default value, e.g. if you don't provide anything, it's defaulted to 10s. Also, we don't default this in code, we just don't provide the flag value. This is prone to some problems like the API server changing default values, then our documentation is not going to match the reality.

We should either:
- Default to proper values in our defaulting code
- Use pointers, if nil, don't set the flag, otherwise set it (I like this one)

xmudrii (approved):
We have agreed to merge this PR and iterate later on. I left some comments for @kron4eg to fix later on.

@SimonTheLeg You can eventually create an issue to follow up on the API refactor.

/lgtm
/approve

docs/api_reference/v1beta3.en.md
SimonTheLeg: @kron4eg it's return of the pointers :D 
(for reference Artiom and I had a talk about this exact thing two days ago)

PR #3319, opened 2024-07-29
Add kubeconfig generate subcommand

### LINKED ISSUE #3082: Decide how do we want to handle `super-admin.conf` kubeconfig

There's a new kubeconfig file at `/etc/kubernetes/super-admin.conf` starting with Kubernetes 1.29 that's part of `system:masters` group. The original kubeocnfig at `/etc/kubernetes/admin.conf` is now bound by RBAC instead of using `system:masters` group.

We have to figure out how do we want to handle `super-admin.conf`:

- do we want to delete it?
- do we want to generate it on our own, but short-lived?

### Solution details

- TBD

### Alternative approaches

TBD

kron4eg:
Fixes #3082

**What type of PR is this?**
/kind feature

```release-note
* Add new `kubeone kubeconfig generate` command with an option to generate kubeconfig file with the custom permissions (for example: the super-admin kubeconfig). 
* Super-admin kubeconfig (`/etc/kubernetes/super-admin.conf`) automatically generated by kubeadm is now removed from the nodes.
```

**Documentation**:

```documentation
TBD
```

### DIFF
--- a/pkg/certificate/pki.go
+++ b/pkg/certificate/pki.go
@@ -87,17 +87,20 @@ func encodePrivateKeyPEM(key *rsa.PrivateKey) []byte {
 	return pem.EncodeToMemory(&block)
 }
 
-// newPrivateKey creates an RSA private key
-func newPrivateKey() (*rsa.PrivateKey, error) {
-	return rsa.GenerateKey(rand.Reader, rsaKeySize)
+// NewPrivateKey creates an RSA private key
+func NewPrivateKey() (*rsa.PrivateKey, error) {
+	key, err := rsa.GenerateKey(rand.Reader, rsaKeySize)
+
+	return key, fail.Runtime(err, "generating RSA private key")
 }
 
-// newSignedCert creates a signed certificate using the given CA certificate and key
-func newSignedCert(cfg *certutil.Config, key crypto.Signer, caCert *x509.Certificate, caKey crypto.Signer) (*x509.Certificate, error) {
+// NewSignedCert creates a signed certificate using the given CA certificate and key
+func NewSignedCert(cfg *certutil.Config, key crypto.Signer, caCert *x509.Certificate, caKey crypto.Signer, notAfter time.Time) (*x509.Certificate, error) {
 	serial, err := rand.Int(rand.Reader, new(big.Int).SetInt64(math.MaxInt64))
 	if err != nil {
-		return nil, err
+		return nil, fail.Runtime(err, "generating random int")
 	}
+
 	if len(cfg.CommonName) == 0 {
 		return nil, fail.RuntimeError{
 			Op:  "checking requested CommonName",
@@ -112,6 +115,11 @@ func newSignedCert(cfg *certutil.Config, key crypto.Signer, caCert *x509.Certifi
 		}
 	}
 
+	notBefore := cfg.NotBefore.UTC()
+	if notBefore.IsZero() {
+		notBefore = time.Now().UTC()
+	}
+
 	certTmpl := x509.Certificate{
 		Subject: pkix.Name{
 			CommonName:   cfg.CommonName,
@@ -120,8 +128,8 @@ func newSignedCert(cfg *certutil.Config, key crypto.Signer, caCert *x509.Certifi
 		DNSNames:     cfg.AltNames.DNSNames,
 		IPAddresses:  cfg.AltNames.IPs,
 		SerialNumber: serial,
-		NotBefore:    caCert.NotBefore,
-		NotAfter:     time.Now().Add(duration365d).UTC(),
+		NotBefore:    notBefore,
+		NotAfter:     notAfter.UTC(),
 		KeyUsage:     x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature,
 		ExtKeyUsage:  cfg.Usages,
 	}
--- a/pkg/cmd/kubeconfig.go
+++ b/pkg/cmd/kubeconfig.go
@@ -73,7 +73,7 @@ type kubeconfigGenerateOpts struct {
 	globalOptions
 	CommonName        string        `longflag:"cn"`
 	OrganizationNames []string      `longflag:"on"`
-	ShortSuperAdmin   bool          `longflag:"super-admin" shortflag:"s"`
+	SuperAdmin        bool          `longflag:"super-admin" shortflag:"s"`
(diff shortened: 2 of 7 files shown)


pkg/cmd/kubeconfig.go
```diff
@@ -45,21 +53,38 @@ func kubeconfigCmd(rootFlags *pflag.FlagSet) *cobra.Command {
...
-	}
+func runKubeconfig(st *state.State, generateSuperAdmin bool) error {
+	var (
+		konfig []byte
+		err    error
+	)
 
-	konfig, err := kubeconfig.Download(s)
+	if generateSuperAdmin {
+		konfig, err = kubeconfig.GenerateSuperAdmin(st)
```
xmudrii, line 84: I think we should extend this so the users can specify how long the kubeconfig should be valid for.

pkg/cmd/kubeconfig.go
```diff
@@ -45,21 +53,38 @@ func kubeconfigCmd(rootFlags *pflag.FlagSet) *cobra.Command {
...
+	} else {
+		konfig, err = kubeconfig.Download(st)
```
xmudrii, line 86: I'm wondering if we should replace this with generating our own kubeconfig where users can specify the group and for how long the kubeconfig should be valid for.

pkg/kubeconfig/kubeconfig.go
```diff
@@ -17,22 +17,124 @@ limitations under the License.
...
+	if err != nil {
+		return nil, err
+	}
+
+	caCerts, err := certutil.ParseCertsPEM(caCertsPEM)
+	if err != nil {
+		return nil, err
+	}
+	if len(caCerts) == 0 {
+		return nil, fmt.Errorf("no certificates found in %s", pkiCAcrt)
```
xmudrii, line 72: Nit: why not `fail.NewRuntimeError` here instead?

pkg/kubeconfig/kubeconfig.go
```diff
@@ -17,22 +17,124 @@ limitations under the License.
...
+		return nil, fail.NewRuntimeError("type asserting rsa.PrivateKey type", "private key is not a RSA private key")
+	}
+
+	superAdminUserKey, err := certificate.NewPrivateKey()
+	if err != nil {
+		return nil, err
+	}
+
+	certCfg := certutil.Config{
+		CommonName:   "short-lived-super-admin",
```
xmudrii, line 96: I think we should make this customizable. AFAIK, this name is used for different purposes including the audit logging, so it might be important to our users. We can default to `short-lived-super-admin`, but there should be a flag where you can set this.

xmudrii:
@kron4eg Can you please extend the release note? I imagine something like this would work fine:
```
Extend the `kubeone kubeconfig` command with an option to generate a short-lived kubeconfig file with the `system:masters` permissions (aka the super-admin kubeconfig). The long-lived super-admin kubeconfig (`/etc/kubernetes/super-admin.conf`) automatically generated by kubeconfig is removed from the nodes by default.
```
Also, please make sure that you set docs to TBD, as we need to document this feature for sure.

pkg/certificate/pki.go
```diff
@@ -112,6 +115,11 @@ func newSignedCert(cfg *certutil.Config, key crypto.Signer, caCert *x509.Certifi
 		}
 	}
 
+	notBefore := cfg.NotBefore.UTC()
+	if notBefore.IsZero() {
+		notBefore = time.Now().UTC()
```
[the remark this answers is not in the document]
kron4eg: caCert.NotBefore has nothing to do with client cert notBefore. Those are completely different notBefores. Short lived cert = small fixed time frame to live. From now + 1h for example.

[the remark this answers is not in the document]
xmudrii: This is what we had before:
```
NotBefore: caCert.NotBefore,
```
Hence my question, I was wondering if we should keep it or not, but I think this makes more sense anyways, so +1 to this change.

pkg/kubeconfig/kubeconfig.go
kron4eg: Missed it probably after experimentation phase.

pkg/cmd/kubeconfig.go
kron4eg: added as `kubeone kubeconfig generate --ttl` flag

pkg/certificate/pki.go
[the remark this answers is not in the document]
kron4eg: IDK why it was caCert.NotBefore before :man_shrugging:. I assume it was some mistake or something.

pkg/cmd/kubeconfig.go
```diff
@@ -45,21 +49,108 @@ func kubeconfigCmd(rootFlags *pflag.FlagSet) *cobra.Command {
...
+		"CommonName (CN) for the generated client certificate.")
+
+	cmd.Flags().StringArrayVar(
+		&opts.OrganizationNames,
+		longFlagName(opts, "OrganizationNames"),
+		[]string{},
+		"OrganizationName (ON) for the generated client certificate.")
+
+	cmd.Flags().BoolVarP(
+		&opts.ShortSuperAdmin,
```
xmudrii, line 117: We should have validation to reject `--super-admin` if ON and CN are provided.

pkg/kubeconfig/kubeconfig.go
```diff
@@ -17,22 +17,124 @@ limitations under the License.
...
+	"k8s.io/client-go/util/keyutil"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	pkiCAcrt = "/etc/kubernetes/pki/ca.crt"
+	pkiCAkey = "/etc/kubernetes/pki/ca.key"
+)
+
+func GenerateSuperAdmin(st *state.State, cn string, on []string, ttl time.Duration) ([]byte, error) {
```
xmudrii, line 51: This function can generate any kubeconfig now, so we should rename it to something else. Maybe `GenerateKubeconfig` or just `Generate`?

xmudrii (changes requested):

pkg/cmd/kubeconfig.go
kron4eg: ON is getting (silently) overwritten and that might be a problem, but what's the problem with CN?

xmudrii (approved):
/lgtm
/approve

PR #3332, opened 2024-08-06
Deploy default-storage-class addon by default

### LINKED ISSUE #2708: UX - deploy default storage classes by default

kron4eg:
Unless it's already configured and there is no another default StorageClass

Fixes #2708

**What type of PR is this?**
/kind feature

```release-note
Deploy default-storage-class addon by default
```

**Documentation**:

```documentation
TBD
```

### DIFF
--- a/pkg/addons/ensure.go
+++ b/pkg/addons/ensure.go
@@ -19,21 +19,22 @@ package addons
 import (
 	"fmt"
 	"io/fs"
+	"slices"
 
 	"github.com/pkg/errors"
 
+	kubeoneapi "k8c.io/kubeone/pkg/apis/kubeone"
 	"k8c.io/kubeone/pkg/fail"
 	"k8c.io/kubeone/pkg/state"
 	"k8c.io/kubeone/pkg/templates/resources"
 	"k8c.io/kubeone/pkg/templates/weave"
+
+	storagev1 "k8s.io/api/storage/v1"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


pkg/addons/ensure.go
```diff
@@ -140,6 +141,32 @@ func collectAddons(s *state.State) []addonAction {
...
+				foundDefaultStorageClass = true
+			}
+		}
+	}
+
+	if !foundDefaultStorageClass {
+		addonsToDeploy = append(addonsToDeploy, addonAction{
+			name: resources.AddonDefaultStorageClass,
+		})
+	}
```
xmudrii, line 168: We should have a way to disable deployment of this addon. Do we already have that in place?

xmudrii (changes requested):

pkg/addons/ensure.go
kron4eg: no. there was no such way.

kron4eg: Well, with delete: true it will be deleted everytime (which after the first time will have no affect any more). And once the used will have own default SC, this addon will be stopped from adding.

xmudrii (approved):
/lgtm
/approve
/hold

PR #3333, opened 2024-08-06
Stop defaulting kube-api endpoint

### LINKED ISSUE #1848: Consider removing defaulting for the API endpoint
**What happened**:

Some providers don't provide the IP address of the API server LB on the first `terraform apply` run. If there's no API server LB in the Terraform state, and the user runs `kubeone apply`, KubeOne will default the API endpoint to the first node (even if there's LB created). Such a cluster is not an HA cluster, and the user should be aware of that.

The API endpoint should not be defaulted. KubeOne should throw an error if the API endpoint is missing. If users want to use the first (or any other) node as the Load Balancer, they should explicitly provide its IP address as a way of confirming their intentions.

See #1847

Changing the API endpoint once the cluster is created is not an option, because it's considered to be risky, as it might break the cluster.

kron4eg:
Fixes #1848

**What type of PR is this?**
/kind api-change

```release-note
Stop defaulting kube-api endpoint
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/apis/kubeone/config/config.go
+++ b/pkg/apis/kubeone/config/config.go
@@ -72,7 +72,27 @@ func LoadKubeOneCluster(clusterCfgPath, tfOutputPath, credentialsFilePath string
 		return nil, fail.Runtime(err, "reading cluster configuration")
 	}
 
-	var tfOutput []byte
+	tfOutput, err := TFOutput(tfOutputPath)
+	if err != nil {
+		return nil, err
+	}
+
+	var credentialsFile []byte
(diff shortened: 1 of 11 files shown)


xmudrii (changes requested):
We should have a migration for this change, i.e. `kubeone config migrate` should apply this migration.

kron4eg:
blocked by #3248 to get the migration framework

xmudrii (approved):
/lgtm
/approve

PR #3334, opened 2024-08-06
Validate that IP addresses for ControlPlane and StaticWorkers are different

### LINKED ISSUE #2632: Disallow using same machines for both control plane and static workers

KubeOne considers the following config as "valid":

```yaml
...
controlPlane:
  hosts:
    - publicAddress: '192.168.1.2'
      privateAddress: '192.168.1.2'

staticWorkers:
  hosts:
    - publicAddress: '192.168.1.2'
      privateAddress: '192.168.1.2'
```

Such a config cannot produce a working cluster, as you can't reuse the same machine for both a control plane and a static worker. Instead, you can untaint the control plane node and schedule all the workload there.

### Solution details

- KubeOne validation should reject such a config

kron4eg:
Fixes #2632

**What type of PR is this?**
/kind api-change

```release-note
Validate that IP addresses for ControlPlane and StaticWorkers are different
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/apis/kubeone/validation/validation.go
+++ b/pkg/apis/kubeone/validation/validation.go
@@ -544,12 +544,34 @@ func ValidateStaticWorkersConfig(staticWorkers kubeoneapi.StaticWorkersConfig, c
 
 	for idx, worker := range staticWorkers.Hosts {
 		for _, cp := range controlPlane.Hosts {
+			if cp.Hostname != "" && worker.Hostname != "" && cp.Hostname == worker.Hostname {
+				allErrs = append(allErrs,
+					field.Invalid(
+						fldPath.Child("hosts").Index(idx),
+						"Hostname",
+						fmt.Sprintf("hostname %q already used for control-plane node", worker.Hostname),
+					),
+				)
+			}
+
 			if cp.PrivateAddress == worker.PrivateAddress {
-				allErrs = append(allErrs, field.Invalid(fldPath.Child("hosts").Index(idx), "PrivateAddress", "can not be the same as control plane"))
+				allErrs = append(allErrs,
+					field.Invalid(
+						fldPath.Child("hosts").Index(idx),
+						"PrivateAddress",
+						fmt.Sprintf("private IP address %q already used for control-plane node", worker.PrivateAddress),
+					),
+				)
 			}
 
 			if cp.PublicAddress == worker.PublicAddress {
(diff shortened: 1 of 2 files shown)


xmudrii:
@kron4eg What if we add public and private addresses, and hostnames, of CP and static worker nodes to two different sets, and then compare sets? That seems like a great way to prevent any possible combination of what can go wrong, as long as we find any intersection, we error out.

kron4eg:
> @kron4eg What if we add public and private addresses, and hostnames, of CP and static worker nodes to two different sets, and then compare sets? That seems like a great way to prevent any possible combination of what can go wrong, as long as we find any intersection, we error out.

Hostnames are optional and may be absent at the time when validation is running. That's why I didn't included them in the check. Given nested for loop almost exactly the same as sets but without fancy APIs/functions, except one thing. It point to the exact index of the problematic static worker with with .Index() function. Which is not possible with sets.

xmudrii:
> Hostnames are optional and may be absent at the time when validation is running. That's why I didn't included them in the check.

That's true, but if we know them, it's useful to do the validation, and with sets this is very easy.

> Given nested for loop almost exactly the same as sets but without fancy APIs/functions, except one thing.

They don't function the exactly same, for example, you might "hack" around by inverting public and private addresses. This and some other cases are much easier to catch by doing the intersection.

> It point to the exact index of the problematic static worker with with .Index() function. Which is not possible with sets.

This is nice, but IMO it's also enough just to point out to what value is conflicting and you can do that with intersect.

kron4eg:
TBH I don't want to change the thing "just because" with loosing some of the info. There has to be a good reason for that and I don't see sets as one.

xmudrii:
@kron4eg Can we at least include the information what's the conflicting IP address? I find that much more valuable compared to the index.

kron4eg:
> @kron4eg Can we at least include the information what's the conflicting IP address? I find that much more valuable compared to the index.

sure, and as a bonus optional validation on the hostname (excluding empty case).

xmudrii (approved):
/lgtm
/approve

xmudrii:
/release-note-edit
```release-note
[ACTION REQUIRED] KubeOne is now validating that IP addresses and hostnames provided for control plane nodes and static worker nodes are different. In other words, it's not possible to use the same machine both as a control plane node and a static worker node. This behavior has never been supported by KubeOne; if you want a control plane node that can schedule any pod, you can provision it as a control plane node and remove the control plane taint (`node-role.kubernetes.io/control-plane:NoSchedule`).
```

PR #3348, opened 2024-08-14
Add `--prune-images` for upgrade process

### LINKED ISSUE #3231: Prune images after successful Kubernetes upgrade

After ~2 Kubernetes version upgrades, we always receive alerts as disk usage has climbed >80% (of 50GB). We then address this by SSHing into all control plane nodes and executing `sudo crictl rmi --prune` to delete unused images, bringing us down to ~40% usage.

It would be nice if we could tell kubeone, probably with a flag, to prune the images on all control plane nodes after the control plane has been upgraded. It's not super elaborate to do it by hand, but it would be a quality of life improvement. The actual cleanup isn't even the "hard" task, but the alert that fires first is often something unrelated (e.g. node-exporter running out of disk space) and thus it first takes a bit of investigation.

With a `--prune-images` flag, we'd just add that to our pipeline job and wouldn't have to care about it anymore.

### Solution details

Add a flag, such as `--prune-images`, to the `kubeone` CLI that runs `sudo crictl rmi --prune` on each control plane node after the upgrade (either on that node or of the whole cluster - you're probably the better judge of that). I guess ideally it could also be used to prune images even when no upgrade is performed, in case people want to run it separately after the upgrade was done.

### Alternative approaches

SSH into each control plane node and execute the command manually.

mohamed-rafraf:
This PR add `--prune-images` for upgrade process. 

Fixes #3231

**What type of PR is this?**

/kind feature

```release-note
Delete unused container images after upgrading the cluster (Mostly the old control plane images)
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/cmd/apply.go
+++ b/pkg/cmd/apply.go
@@ -138,7 +138,7 @@ func applyCmd(rootFlags *pflag.FlagSet) *cobra.Command {
 		&opts.PruneImages,
 		longFlagName(opts, "PruneImages"),
 		false,
-		"delete unused container images")
+		"delete unused container images on control plane and static worker nodes")
 
 	cmd.Flags().BoolVar(
 		&opts.CreateMachineDeployments,
--- a/pkg/cmd/apply.go
+++ b/pkg/cmd/apply.go
@@ -44,6 +44,7 @@ type applyOpts struct {
 	// Upgrade flags
 	ForceUpgrade              bool `longflag:"force-upgrade"`
 	UpgradeMachineDeployments bool `longflag:"upgrade-machine-deployments"`
+	PruneImages               bool `longflag:"prune-images"`
 	CreateMachineDeployments  bool `longflag:"create-machine-deployments"`
 	RotateEncryptionKey       bool `longflag:"rotate-encryption-key"`
 }
@@ -58,6 +59,7 @@ func (opts *applyOpts) BuildState() (*state.State, error) {
 	s.ForceInstall = opts.ForceInstall
 	s.ForceUpgrade = opts.ForceUpgrade
 	s.UpgradeMachineDeployments = opts.UpgradeMachineDeployments
+	s.PruneImages = opts.PruneImages
 	s.CreateMachineDeployments = opts.CreateMachineDeployments
 
 	return s, initBackup(s.BackupFile)
@@ -131,6 +133,12 @@ func applyCmd(rootFlags *pflag.FlagSet) *cobra.Command {
 		longFlagName(opts, "UpgradeMachineDeployments"),
 		false,
 		"upgrade MachineDeployments objects")
+
+	cmd.Flags().BoolVar(
+		&opts.PruneImages,
+		longFlagName(opts, "PruneImages"),
+		false,
+		"delete unused container images")
 
 	cmd.Flags().BoolVar(
 		&opts.CreateMachineDeployments,
--- a/pkg/cmd/upgrade.go
+++ b/pkg/cmd/upgrade.go
@@ -89,7 +89,7 @@ func upgradeCmd(rootFlags *pflag.FlagSet) *cobra.Command {
 		&opts.PruneImages,
 		longFlagName(opts, "PruneImages"),
 		false,
-		"delete unused container images")
+		"delete unused container images on control plane and static worker nodes")
 
 	return cmd
 }
--- a/pkg/cmd/upgrade.go
+++ b/pkg/cmd/upgrade.go
@@ -29,6 +29,7 @@ type upgradeOpts struct {
 	globalOptions
 	ForceUpgrade              bool `longflag:"force" shortflag:"f"`
 	UpgradeMachineDeployments bool `longflag:"upgrade-machine-deployments"`
+	PruneImages               bool `longflag:"prune-images"`
 }
 
 func (opts *upgradeOpts) BuildState() (*state.State, error) {
@@ -83,6 +84,12 @@ func upgradeCmd(rootFlags *pflag.FlagSet) *cobra.Command {
 		longFlagName(opts, "UpgradeMachineDeployments"),
 		false,
 		"upgrade MachineDeployments objects")
+
+	cmd.Flags().BoolVar(
+		&opts.PruneImages,
+		longFlagName(opts, "PruneImages"),
+		false,
+		"delete unused container images")
 
 	return cmd
 }
--- a/pkg/tasks/nodes.go
+++ b/pkg/tasks/nodes.go
@@ -99,7 +99,7 @@ func ensureRestartKubeAPIServerCrictl(s *state.State) error {
 }
 
 func pruneImagesOnAllNodes(s *state.State) error {
-	s.Logger.Infof("Deleting unused container Images...")
+	s.Logger.Infof("Deleting unused container images...")
 
 	// Prune unused container images on all nodes.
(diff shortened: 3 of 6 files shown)


pkg/cmd/apply.go
```diff
@@ -132,6 +134,12 @@ func applyCmd(rootFlags *pflag.FlagSet) *cobra.Command {
 		false,
 		"upgrade MachineDeployments objects")
 
+	cmd.Flags().BoolVar(
+		&opts.PruneImages,
+		longFlagName(opts, "PruneImages"),
+		false,
+		"delete unused container images")
```
xmudrii, line 141:
```suggestion
		"delete unused container images on control plane and static worker nodes")
```

pkg/cmd/upgrade.go
```diff
@@ -84,6 +85,12 @@ func upgradeCmd(rootFlags *pflag.FlagSet) *cobra.Command {
 		false,
 		"upgrade MachineDeployments objects")
 
+	cmd.Flags().BoolVar(
+		&opts.PruneImages,
+		longFlagName(opts, "PruneImages"),
+		false,
+		"delete unused container images")
```
xmudrii, line 92:
```suggestion
		"delete unused container images on control plane and static worker nodes")
```

pkg/tasks/nodes.go
```diff
@@ -45,6 +45,13 @@ func restartKubeAPIServer(s *state.State) error {
 		return restartKubeAPIServerOnOS(s, *node)
 	}, state.RunSequentially)
 }
+func pruneImages(s *state.State) error {
+	s.Logger.Infoln("Prune container Images...")
```
xmudrii, line 49:
```suggestion
	s.Logger.Infoln("Pruning unused container Images...")
```

pkg/tasks/nodes.go
```diff
@@ -78,6 +85,18 @@ func ensureRestartKubeAPIServerOnOS(s *state.State, node kubeoneapi.HostConfig)
...
+	return runOnOS(s, node.OperatingSystem, map[kubeoneapi.OperatingSystemName]runOnOSFn{
+		kubeoneapi.OperatingSystemNameAmazon:     pruneImagesCrictl,
+		kubeoneapi.OperatingSystemNameCentOS:     pruneImagesCrictl,
+		kubeoneapi.OperatingSystemNameDebian:     pruneImagesCrictl,
+		kubeoneapi.OperatingSystemNameFlatcar:    pruneImagesCrictl,
+		kubeoneapi.OperatingSystemNameRHEL:       pruneImagesCrictl,
+		kubeoneapi.OperatingSystemNameRockyLinux: pruneImagesCrictl,
+		kubeoneapi.OperatingSystemNameUbuntu:     pruneImagesCrictl,
+	})
+}
```
xmudrii, line 98: We don't really need this because the command is the same for all OS-es.

pkg/tasks/nodes.go
```diff
@@ -98,6 +117,13 @@ func ensureRestartKubeAPIServerCrictl(s *state.State) error {
 	return fail.SSH(err, "restarting kubeapi-server pod")
 }
 
+func pruneImagesCrictl(s *state.State) error {
+	cmd := scripts.PruneImages()
+	_, _, err := s.Runner.RunRaw(cmd)
+
+	return fail.SSH(err, "deleting unused container images...")
```
xmudrii, line 124:
```suggestion
	return fail.SSH(err, "deleting unused container images")
```

pkg/tasks/nodes.go
```diff
@@ -45,6 +45,13 @@ func restartKubeAPIServer(s *state.State) error {
 		return restartKubeAPIServerOnOS(s, *node)
 	}, state.RunSequentially)
 }
+func pruneImages(s *state.State) error {
+	s.Logger.Infoln("Prune container Images...")
+
+	return s.RunTaskOnControlPlane(func(s *state.State, node *kubeoneapi.HostConfig, _ executor.Interface) error {
```
xmudrii, line 51: We should run this on all KubeOne-managed nodes IMO.

pkg/tasks/tasks.go
```diff
@@ -352,6 +352,11 @@ func WithUpgrade(t Tasks, followers ...kubeoneapi.HostConfig) Tasks {
 				Description: "upgrade MachineDeployments",
 				Predicate:   func(s *state.State) bool { return s.UpgradeMachineDeployments },
 			},
+			Task{
+				Fn: pruneImages, Operation: "deleting unused container images",
```
xmudrii, line 356:
```suggestion
				Fn: pruneImages, 
				Operation: "pruning unused container images",
```

pkg/tasks/tasks.go
```diff
@@ -352,6 +352,11 @@ func WithUpgrade(t Tasks, followers ...kubeoneapi.HostConfig) Tasks {
...
+				Description: "delete unused container images",
```
xmudrii, line 357:
```suggestion
				Description: "prune unused container images",
```

xmudrii (changes requested):
Looks good! Some nits and we can merge this!

pkg/tasks/nodes.go
```diff
@@ -98,6 +98,23 @@ func ensureRestartKubeAPIServerCrictl(s *state.State) error {
 	return fail.SSH(err, "restarting kubeapi-server pod")
 }
 
+func pruneImagesOnAllNodes(s *state.State) error {
+	s.Logger.Infof("Prune unused container Images...")
```
xmudrii, line 102:
```suggestion
	s.Logger.Infof("Deleting unused container images...")
```

pkg/tasks/tasks.go
```diff
@@ -352,6 +352,11 @@ func WithUpgrade(t Tasks, followers ...kubeoneapi.HostConfig) Tasks {
 				Description: "upgrade MachineDeployments",
 				Predicate:   func(s *state.State) bool { return s.UpgradeMachineDeployments },
 			},
+			Task{
+				Fn: pruneImagesOnAllNodes, Operation: "pruning unused container images",
+				Description: "prune unused container images",
+				Predicate:   func(s *state.State) bool { return s.PruneImages },
```
xmudrii, line 358: Nit: `Fn` and `Operation` should be on two lines, also let's replace `pruning` with `deleting` to standardize the naming.

xmudrii (changes requested):
Two very tiny nits. The last one might revert my comment from the previous review, sorry about that! The idea is to have standard naming everywhere, i.e. if the flag is referring to it as `deleting`, we also show that in the logs.

pkg/cmd/upgrade.go
```diff
@@ -29,6 +29,7 @@ type upgradeOpts struct {
 	globalOptions
 	ForceUpgrade              bool `longflag:"force" shortflag:"f"`
 	UpgradeMachineDeployments bool `longflag:"upgrade-machine-deployments"`
+	PruneImages               bool `longflag:"prune-images"`
```
kron4eg, line 32: we don't need to update that deprecated and hidden subcommand. In fact I'm surprised it still exists and we forgot to remove it and install subcommand :man_shrugging: :joy:

kron4eg: OK, so together with @mohamed-rafraf we've decided to go ahead and remove those commands in the follow up PR, so I'm /lgtm /approve this one (to avoid additional retesting).

PR #3352, opened 2024-08-14
Add Pod PID Limit in Kubelet Configuration

### LINKED ISSUE #3351: Add Pod PID Limit in Kubelet Configuration

As a kubeone user, I would like to configure the Pod PID Limit from `kubeone.yaml` to pass the CIS Benchmark test `4.2.13 Ensure that a limit is set on pod PIDs (Manual)`

### Solution details

This is an example of the Kubelet configuration

```yaml
hosts:
- kubelet:
    systemReserved: xxxxx
    kubeReserved: xxxxx
    evictionHard: xxxxx
    podPidsLimit: 10000

```

### Additional information

In case the podPidsLimit is not used. It would be great to use the default value used by kubelet

mohamed-rafraf:
This PR adds Pod PID Limit in Kubelet Configuration. We can Pass the CIS Benchmark `4.2.13 Ensure that a limit is set on pod PIDs (Manual)`

Fixes #3351

**What type of PR is this?**

/kind api-change
/kind feature

```release-note
The Kubeone HostConfig support Pod PID Limits in its Kubelet Config
```

**Documentation**:

```documentation
T.B.D
```


mohamed-rafraf:
I built a Kubeone binary and created an AWS cluster to run CIS Benchmark:

image

pkg/apis/kubeone/v1beta2/types.go
```diff
@@ -353,6 +353,11 @@ type KubeletConfig struct {
 	// If not provided, default value provided by kubelet will be used
 	// (max. 110 pods per node)
 	MaxPods *int32 `json:"maxPods,omitempty"`
+
+	// PodPidsLimit configures the maximum number of processes running in a Pod
+	// If not provided, default value provided by kubelet will be used -1
+	// See more about pid-limiting at: https://kubernetes.io/docs/concepts/policy/pid-limiting/
+	PodPidsLimit *int64 `json:"podPidsLimit,omitempty"`
```
kron4eg, line 360: Please don't change v1beta2. We want to get rid of it in the future.

pkg/apis/kubeone/types.go
```diff
@@ -343,6 +343,11 @@ type KubeletConfig struct {
 	// If not provided, default value provided by kubelet will be used
 	// (max. 110 pods per node)
 	MaxPods *int32 `json:"maxPods,omitempty"`
+
+	// PodPidsLimit configures the maximum number of processes running in a Pod
```
xmudrii, line 347: There should be a dot at the end of this line for better API docs rendering, i.e. `Pod.`

pkg/apis/kubeone/types.go
```diff
@@ -343,6 +343,11 @@ type KubeletConfig struct {
...
+	// If not provided, default value provided by kubelet will be used -1
```
xmudrii, line 348: Same here, we should add a dot at the end of this line.

pkg/templates/kubeadm/v1beta3/kubeadm.go
```diff
@@ -590,6 +590,14 @@ func newNodeRegistration(s *state.State, host kubeoneapi.HostConfig) kubeadmv1be
 		kubeletCLIFlags["max-pods"] = strconv.Itoa(int(*m))
 	}
 
+	if m := host.Kubelet.PodPidsLimit; m != nil {
+		kubeletCLIFlags["pod-max-pids"] = strconv.Itoa(int(*m))
+	} else {
+		// Set default value if PodsPidsLimits is nil
+		// in order to pass the 4.2.13 Check in CIS Benchmark 1.8
+		kubeletCLIFlags["pod-max-pids"] = strconv.Itoa(-1)
+	}
```
xmudrii, line 599: Have we checked if this solves the CIS benchmark issue or they require setting a value that's not -1?

mohamed-rafraf: CIS Benchmark requires to put a value and don't rely on kubelet default value. So in our case if the pod pid limit we will put the default value of kubelet to pass the test

xmudrii: That's great, thanks for following up on this!

PR #3354, opened 2024-08-16
Add server groups to OpenStack example

jan-di:

Add server groups to OpenStack example. Server groups allow controlling anti affinity of control plane nodes. Therefore, you can ensure that control plane nodes run on different hypervisors. Per default, its configured to use "soft-anti-affinity" (try to use different hypervisors, fallback to use the same hypervisors if not possible to spread).

**What type of PR is this?**

```release-note
NONE
```

**Documentation**:

```documentation

```

### DIFF
--- a/examples/terraform/openstack/control_plane.tf
+++ b/examples/terraform/openstack/control_plane.tf
@@ -14,6 +14,11 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+resource "openstack_compute_servergroup_v2" "control_plane" {
+  name     = "${var.cluster_name}-control_plane"
+  policies = [var.control_plane_group_policy]
+}
+
 resource "openstack_networking_port_v2" "control_plane" {
   count = var.control_plane_vm_count
   name  = "${var.cluster_name}-control_plane-${count.index}"
@@ -36,6 +41,10 @@ resource "openstack_compute_instance_v2" "control_plane" {
   key_pair        = openstack_compute_keypair_v2.deployer.name
   security_groups = [openstack_networking_secgroup_v2.securitygroup.name]
   config_drive    = var.config_drive
+
+  scheduler_hints {
+    group = openstack_compute_servergroup_v2.control_plane.id
+  }
 
   network {
     port = element(openstack_networking_port_v2.control_plane[*].id, count.index)
--- a/examples/terraform/openstack/variables.tf
+++ b/examples/terraform/openstack/variables.tf
@@ -88,6 +88,12 @@ variable "control_plane_flavor" {
   type        = string
 }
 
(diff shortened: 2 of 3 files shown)


examples/terraform/openstack/control_plane.tf
```diff
@@ -37,6 +42,10 @@ resource "openstack_compute_instance_v2" "control_plane" {
   security_groups = [openstack_networking_secgroup_v2.securitygroup.name]
   config_drive    = var.config_drive
 
+  scheduler_hints {
+    group = openstack_compute_servergroup_v2.control_plane.id
+  }
```
xmudrii, line 47: Can we provide an easy way to disable this? For example, if we can control this with a variable.

examples/terraform/openstack/variables.tf
```diff
@@ -88,6 +88,12 @@ variable "control_plane_flavor" {
   type        = string
 }
 
+variable "control_plane_group_policy" {
+  description = "OpenStack instance group policy for the control plane nodes. Cannot be changed without recreating VMs!"
```
xmudrii, line 92:
```suggestion
  description = "OpenStack instance group policy for the control plane nodes"
```

We'll document the other part in the changelog.

examples/terraform/openstack/variables.tf
```diff
@@ -88,6 +88,12 @@ variable "control_plane_flavor" {
...
+  default     = "soft-anti-affinity"
```
xmudrii, line 93: Can we somewhere document supported values and why we recommend `soft-anti-affinity`?

examples/terraform/openstack/control_plane.tf
```diff
@@ -14,6 +14,11 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+resource "openstack_compute_servergroup_v2" "control_plane" {
+  name     = "${var.cluster_name}-control_plane"
+  policies = [var.control_plane_group_policy]
+}
```
xmudrii, line 20: Are policies supported by default on all OpenStack installations or does it require policies to be explicitly enabled and created?

examples/terraform/openstack/variables.tf
sschmittsva: There are four policies supported.

From https://docs.openstack.org/nova/2024.2/user/server-groups.html:

* `affinity`
* `anti-affinity`
* `soft-affinity`
* `soft-anti-affinity`

`soft-anti-affinity` looks like a good default:

> Attempts to restrict instances belonging to the server group to separate hosts. Where it is not possible to schedule all instances to separate hosts, they will be scheduled on as many separate hosts as possible.

Issue #3346, opened 2024-08-14
Support Kubernetes 1.31


xmudrii:
This is a collector issue for Kubernetes 1.31 support in KubeOne. The following tasks should be taken care of:

* [x] Check the Kubernetes changelog to ensure there are no breaking changes and removals
* [x] Update the `build` image to update Sonobuoy (and other dependencies if needed) (#3358)
* [x] Update the latest supported Kubernetes version in [the API validation](https://github.com/kubermatic/kubeone/blob/main/pkg/apis/kubeone/validation/validation.go#L40-L41) (#3358)
* [x] Update [default admission controllers](https://github.com/kubermatic/kubeone/blob/main/pkg/kubeflags/data.go) if needed (not needed)
* [x] Update `pause` image version in `pkg/apis/kubeone/helpers.go` if needed (#3358)
* [x] Add the appropriate `cri-tools` version in `pkg/scripts/os_flatcar.go` (#3358)
* [x] Update [the stable version marker in Makefile](https://github.com/kubermatic/kubeone/blob/5273f9a372736569c6b09b38f2959019d29e4d6a/Makefile#L24) (#3358)
* [x] Add E2E tests inside `tests.yml` (#3358)
* [x] Update daily periodics to use the latest Kubernetes release
* [x] Update [the Compatibility Matrix](https://docs.kubermatic.com/kubeone/main/architecture/compatibility/supported-versions/) (https://github.com/kubermatic/docs/pull/1737)
* [x] Create an issue to track updating [images](https://github.com/kubermatic/kubeone/blob/main/pkg/templates/images/images.go) (#3347)

Tips & Tricks:

* Sometimes the image tag does no match any official GitHub Release name. In these cases `crane ls <url of the image>` can be used to check if a new version of an image exists. Nonetheless, changelog still needs to be checked if the new version introduces any breaking changes
* Default Admission Controllers can be obtained via `docker run --rm registry.k8s.io/kube-apiserver:v${KUBERNETES_VERSION} kube-apiserver -h | grep "admission-control"`

**Action items:**

```[tasklist]
### Tasks
- [x] https://github.com/kubermatic/kubeone/pull/3360
- [x] https://github.com/kubermatic/kubeone/pull/3361
- [ ] https://github.com/kubermatic/kubeone/pull/3387
```

PR #3358, opened 2024-08-22
Introduce initial support for Kubernetes v1.31, introduce support for kubeadm v1beta4 API

xmudrii:

This PR introduces two significant changes:

- Support for Kubernetes v1.31
- Support for kubeadm v1beta4 API

I decided to go a step further and do some cleanups now that we're introducing a new kubeadm API that's used only for Kubernetes v1.31+ clusters:

- Code related to the in-tree cloud providers and CCM/CSI migration has been removed from the kubeadm v1beta4 API
- The code has been split into multiple functions to avoid linter complaining due to complexity
- Warnings reported by kubeadm are fixed, mainly the one about what manifests are ignored
  - We are now uploading two files per node, one containing the full configuration (ClusterConfiguration, InitConfiguration, JoinConfiguration, KubeletConfiguration, and KubeProxyConfiguration) and one containing only the JoinConfiguration manifest. The latter is used with the `kubeadm join` command, while the former is used with some other kubeadm commands. In the future, the former manifest/configuration will be phased out
  - The file has been renamed from `master_{{ .NODE_ID }}.yaml` to `control_plane_{{ .NODE_ID }}.yaml` to match the official phraseology change

xref #3346

**What type of PR is this?**

/kind feature

The v1.31 implementation has been tested locally on AWS, including provisioning and upgrading. Some further refactoring will be done in a follow-up PR.

```release-note
- Add initial support for Kubernetes v1.31
- Add support for kubeadm v1beta4 API, used for Kubernetes clusters running v1.31 and newer
```

**Documentation**:
```documentation
TBD
```

/assign @xrstf


pkg/apis/kubeadm/bootstraptoken/v1/utils.go
```diff
@@ -44,15 +56,14 @@ func (bts *BootstrapTokenString) UnmarshalJSON(b []byte) error {
 	}
 
 	// Remove unnecessary " characters coming from the JSON parser
-	token := strings.ReplaceAll(string(b), `"`, ``)
+	token := strings.Replace(string(b), `"`, ``, -1)
```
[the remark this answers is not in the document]
xmudrii: https://github.com/kubermatic/kubeone/blob/e284cacfc6320b8d3caaf84099f572a43d6a16a9/.golangci.yml#L73

😄

pkg/apis/kubeadm/bootstraptoken/v1/utils.go
```diff
@@ -71,8 +81,7 @@ func (bts BootstrapTokenString) String() string {
 // is of the right format
 func NewBootstrapTokenString(token string) (*BootstrapTokenString, error) {
 	substrs := bootstraputil.BootstrapTokenRegexp.FindStringSubmatch(token)
-	// TODO: Add a constant for the 3 value here, and explain better why it's needed (other than because how the regexp parsing works)
-	if len(substrs) != 3 {
+	if len(substrs) != validatedSubstringsSize {
```
xrstf, line 84: You can simply `if substrs == nil` to see if the regex matched.

pkg/apis/kubeadm/bootstraptoken/v1/utils.go
```diff
@@ -24,13 +24,25 @@ import (
 
 	"github.com/pkg/errors"
 
-	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
xrstf, line 27: Why is this being changed back to a worse alias?

pkg/apis/kubeadm/bootstraptoken/v1/utils.go
xmudrii: This file is a copy from k/k, I'd rather leave it as it is to make keeping it up to date easier.

pkg/apis/kubeadm/bootstraptoken/v1/utils.go
xmudrii: https://github.com/kubermatic/kubeone/pull/3358#discussion_r1727404220

pkg/apis/kubeadm/bootstraptoken/v1/utils.go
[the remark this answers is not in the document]
xrstf: My question remains: Why did you change ReplaceAll() with Replace()?

pkg/apis/kubeone/helpers.go
```diff
@@ -35,11 +35,7 @@ const (
 	credentialSecretName = "kube-system/kubeone-registry-credentials" //nolint:gosec
 )
 
-var (
-	v124Constraint         = semverutil.MustParseConstraint(">= 1.24.0, < 1.25.0")
-	v125Constraint         = semverutil.MustParseConstraint(">= 1.25.0, < 1.26.0")
-	v126AndNewerConstraint = semverutil.MustParseConstraint(">= 1.26.0")
-)
+var olderThanV131Constraint = semverutil.MustParseConstraint("< 1.31")
```
xrstf, line 38: This might just be me, but my brain read "older than" as ">" at first.

Maybe rename to `preV131Constraint`?

pkg/apis/kubeone/validation/validation.go
```diff
@@ -39,9 +39,9 @@ import (
 
 const (
 	// lowerVersionConstraint defines a semver constraint that validates Kubernetes versions against a lower bound
-	lowerVersionConstraint = ">= 1.27"
+	lowerVersionConstraint = ">= 1.28"
```
xrstf, line 42: This is a nice comment.

It doesn't help me at all to understand what the constraint _means_. This constant isn't offering any abstraction in itself.

Maybe `minSupported` is a better name or the comment could explain `// ... defines the lowest Kubernetes version supported by KubeOne.`.

pkg/features/activate.go
```diff
@@ -41,7 +41,7 @@ func Activate(s *state.State) error {
 
 // UpdateKubeadmClusterConfiguration update additional config options in the kubeadm's
```
xrstf, line 42: Comment has not been updated, and "update" should be "updates"

pkg/tasks/controlplane.go
```diff
@@ -19,15 +19,20 @@ package tasks
...
 	"k8c.io/kubeone/pkg/executor"
 	"k8c.io/kubeone/pkg/fail"
 	"k8c.io/kubeone/pkg/scripts"
+	"k8c.io/kubeone/pkg/semverutil"
 	"k8c.io/kubeone/pkg/state"
 )
 
 const kubeadmPhaseKubeProxy = "addon/kube-proxy"
 
+var olderThanV131Constraint = semverutil.MustParseConstraint("< 1.31")
```
xrstf, line 34: same optional nit regarding "olderThan"

pkg/tasks/kubeadm_config.go
```diff
@@ -73,7 +76,16 @@ func generateKubeadm(s *state.State) error {
 			return err
 		}
 
-		s.Configuration.AddFile(fmt.Sprintf("cfg/master_%d.yaml", node.ID), kubeadmConf.FullConfiguration)
+		// This needs further refactoring. The follower control plane nodes should not need the full configuration
+		// any circumstances. However, the PR where this is being introduced is already huge, and I don't want
```
xrstf, line 80: There's a word or two missing at the linebreak

xrstf: Can you leave this as a TODO as well?

pkg/templates/kubeadm/kubeadm.go
```diff
@@ -17,20 +17,25 @@ limitations under the License.
...
 	kubeoneapi "k8c.io/kubeone/pkg/apis/kubeone"
+	"k8c.io/kubeone/pkg/semverutil"
 	"k8c.io/kubeone/pkg/state"
 )
 
 const (
 	kubeadmUpgradeNodeCommand = "kubeadm upgrade node"
 )
 
+var olderThanV131Constraint = semverutil.MustParseConstraint("< 1.31")
```
xrstf, line 31: same optional nitpick

pkg/templates/kubeadm/kubeadm.go
```diff
@@ -43,6 +48,15 @@ type Kubedm interface {
 }
 
 // New constructor
```
xrstf, line 50: I wonder how the old constructor looked like.

pkg/templates/kubeadm/kubeadmv1beta4.go
```diff
@@ -0,0 +1,106 @@
+/*
+Copyright 2019 The KubeOne Authors.
```
xrstf, line 2: 2024

pkg/templates/kubeadm/v1beta4/kubeadm.go
```diff
@@ -0,0 +1,621 @@
+/*
+Copyright 2019 The KubeOne Authors.
```
xrstf, line 2: 2024

pkg/templates/kubeadm/v1beta4/kubeadm.go
```diff
@@ -0,0 +1,621 @@
...
+	}
+
+	var advertiseAddress string
+	if cluster.ClusterNetwork.IPFamily.IsIPv6Primary() {
+		advertiseAddress = host.IPv6Addresses[0]
+	} else {
+		advertiseAddress = newNodeIP(host)
+	}
+
+	controlPlaneEndpoint := fmt.Sprintf("%s:%d", cluster.APIEndpoint.Host, cluster.APIEndpoint.Port)
```
xrstf, line 77: net.JoinHostPort()

pkg/templates/kubeadm/v1beta4/kubeadm.go
```diff
@@ -0,0 +1,621 @@
...
+					Name:  "flex-volume-plugin-dir",
+					Value: "/var/lib/kubelet/volumeplugins",
+				},
+				{
+					Name:  "profiling",
+					Value: "false",
+				},
+				{
+					Name:  "terminated-pod-gc-threshold",
+					Value: "1000",
```
xrstf, line 145: If you wanna make me happy, leave a comment or even make it a constant and explain why 1000 and not 999, or even better, 997.

pkg/templates/kubeadm/v1beta4/kubeadm.go
```diff
@@ -0,0 +1,621 @@
...
+				cluster.ClusterNetwork.IPFamily,
+				cluster.ClusterNetwork.ServiceSubnet,
+				cluster.ClusterNetwork.ServiceSubnetIPv6,
+			),
+			DNSDomain: cluster.ClusterNetwork.ServiceDomainName,
+		},
+	}
+
+	if cluster.ClusterNetwork.KubeProxy != nil && cluster.ClusterNetwork.KubeProxy.SkipInstallation {
+		clusterConfig.DNS.Disabled = cluster.ClusterNetwork.KubeProxy.SkipInstallation
```
xrstf, line 189: Can be shortened to `= true`

xrstf: Alternatively, if setting to false would also be fine, remove the second part of the condition instead.

pkg/templates/kubeadm/v1beta4/kubeadm.go
```diff
@@ -0,0 +1,621 @@
...
+	}
+
+	if fErr := addFeaturesExtraMounts(s, clusterConfig); err != nil {
```
xrstf, line 192: fErr vs err

pkg/templates/kubeadm/v1beta4/kubeadm.go
```diff
@@ -0,0 +1,621 @@
...
+		return nil, fErr
+	}
+	addControllerManagerNetworkArgs(clusterConfig, cluster.ClusterNetwork)
+
+	args := kubeadmargs.NewFrom(argsToStringStringMap(clusterConfig.APIServer.ExtraArgs))
```
xrstf, line 197: personally I think `argsToMap` would be plenty specific here.

pkg/templates/kubeadm/v1beta4/kubeadm.go
```diff
@@ -0,0 +1,621 @@
...
+			MountPath: "/etc/kubernetes/admission",
+			ReadOnly:  true,
+			PathType:  corev1.HostPathDirectoryOrCreate,
+		}
+		clusterConfig.APIServer.ExtraVolumes = append(clusterConfig.APIServer.ExtraVolumes, admissionVol)
+	}
+
+	// this is not exactly as s.EncryptionEnabled(). We need this to be true during the enable/disable or disable/enable transition.
+	if (cluster.Features.EncryptionProviders != nil && cluster.Features.EncryptionProviders.Enable) ||
+		s.LiveCluster.EncryptionConfiguration.Enable {
```
xrstf, line 299: possible NPE?

pkg/templates/kubeadm/v1beta4/kubeadm.go
```diff
@@ -0,0 +1,621 @@
...
+// NewConfig returns all required configs to init a cluster via a set of v13 configs
+func NewConfigWorker(s *state.State, host kubeoneapi.HostConfig) (*Config, error) {
+	cluster := s.Cluster
+
+	nodeRegistration := newNodeRegistration(s, host)
+	nodeRegistration.IgnorePreflightErrors = []string{
+		"DirAvailable--etc-kubernetes-manifests",
+	}
+
+	controlPlaneEndpoint := fmt.Sprintf("%s:%d", cluster.APIEndpoint.Host, cluster.APIEndpoint.Port)
```
xrstf, line 401: net.JoinHostPort()

pkg/templates/kubeadm/v1beta4/kubeadm.go
```diff
@@ -0,0 +1,621 @@
...
+		kubeletCLIFlags = append(kubeletCLIFlags, kubeadmv1beta4.Arg{
+			Name:  "cloud-provider",
+			Value: "external",
+		})
+	}
+
+	// If external or in-tree CCM is in use we don't need to set --node-ip
+	// as the cloud provider will know what IPs to return.
+	if s.Cluster.ClusterNetwork.IPFamily.IsDualstack() {
+		if !s.Cluster.CloudProvider.External {
```
xrstf, line 493: The comment seemingly doesn't match the code. It looks like if it's not dualstack, we _always_ set --node-ip, and even if it's dualstack, we still set if for in-tree CCMs?

pkg/templates/kubeadm/v1beta4/kubeadm.go
xrstf: Also, this violates the naming scheme from https://github.com/kubermatic/kubeone/blob/e284cacfc6320b8d3caaf84099f572a43d6a16a9/pkg/apis/kubeone/helpers.go#L496. Personally I find "string-string map" much more natural to pronounce than a literal "map string string" reading". So I would still rename this function here to `argsToMap`.

pkg/templates/kubeadm/v1beta4/kubeadm.go
```diff
@@ -0,0 +1,621 @@
...
+	case kubeoneapi.IPFamilyIPv4:
+		return ipv4Subnet
+	case kubeoneapi.IPFamilyIPv6:
+		return ipv6Subnet
+	case kubeoneapi.IPFamilyIPv4IPv6:
+		return strings.Join([]string{ipv4Subnet, ipv6Subnet}, ",")
+	case kubeoneapi.IPFamilyIPv6IPv4:
+		return strings.Join([]string{ipv6Subnet, ipv4Subnet}, ",")
+	default:
+		return "unknown IP family"
```
xrstf, line 581: This should be a panic(), no?

pkg/templates/kubeadm/v1beta4/kubeadm.go
```diff
@@ -0,0 +1,621 @@
...
+	return m
+}
+
+func newNodeIP(host kubeoneapi.HostConfig) string {
+	nodeIP := host.PrivateAddress
+	if nodeIP == "" {
+		nodeIP = host.PublicAddress
+	}
+
+	return nodeIP
```
xrstf, line 567: `return defaults(host.PrivateAddress, host.PublicAddress)`

pkg/templates/kubeadm/v1beta4/kubeadm_test.go
```diff
@@ -0,0 +1,314 @@
+/*
+Copyright 2022 The KubeOne Authors.
```
xrstf, line 2: 2024

pkg/templates/kubeadm/v1beta4/kubeadm_test.go
```diff
@@ -0,0 +1,314 @@
...
+// areFeatureGatesEqual compares two feature gates strings irrespective of the order of the feature gates.
+func areFeatureGatesEqual(result, expected string) bool {
+	resultMap := make(map[string]bool)
+	expectedMap := make(map[string]bool)
+
+	// Parse the result string into a map
+	for _, fg := range strings.Split(result, ",") {
+		kv := strings.Split(fg, "=")
+		if len(kv) == 2 {
+			resultMap[kv[0]] = kv[1] == "true"
```
xrstf, line 274: This parse logic for a feature gate string is different from the logic in the kubeadm.go.

Maybe the code for merging FGs should be split into 3 functions, one to parse, one to concat and one to merge (that uses parse and concat internally). Then you can re-use the parsing code here in the tests as well.

pkg/templates/kubeadm/v1beta4/kubeadm_test.go
```diff
@@ -0,0 +1,314 @@
...
+		}
+	}
+
+	// Compare the maps
+	if len(resultMap) != len(expectedMap) {
+		return false
+	}
+
+	for k, v := range resultMap {
+		if expectedMap[k] != v {
```
xrstf, line 292: Could this fail because of the implicit zero value? What if `resultMap = {"foo":false}` and `expectedMap = {"bla":true]`, then it would fail for k = `foo`, no?

test/e2e/helpers.go
```diff
@@ -58,7 +58,7 @@ import (
 
 const (
 	labelControlPlaneNode = "node-role.kubernetes.io/control-plane"
-	prowImage             = "quay.io/kubermatic/build:go-1.23-node-20-0"
+	prowImage             = "quay.io/kubermatic/build:go-1.23-node-20-1"
```
xrstf, line 61: This image does not contain Prow.

`prowjobImage` would maybe be better.

xrstf (changes requested):
Consider yourself reviewed.

pkg/apis/kubeadm/bootstraptoken/v1/utils.go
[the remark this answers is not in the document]
xmudrii: https://github.com/kubermatic/kubeone/pull/3358#discussion_r1727404220

pkg/apis/kubeone/helpers.go
xmudrii: I'm struggling to come up with a better name, let's go with `preV131Constraint` as of now. Hopefully we can completely remove this variable at some point soon.

pkg/templates/kubeadm/kubeadm.go
xmudrii: Uh... good question 😅

pkg/templates/kubeadm/v1beta4/kubeadm.go
xmudrii: Neat, changed.

pkg/templates/kubeadm/v1beta4/kubeadm.go
xmudrii: I have zero idea to be honest.

pkg/templates/kubeadm/v1beta4/kubeadm.go
xmudrii: Linter complains because of shadowing, so `fErr` it is or we don't use in-line if/err

pkg/templates/kubeadm/v1beta4/kubeadm.go
xmudrii: `argsToMap` it is. 😄

pkg/templates/kubeadm/v1beta4/kubeadm.go
xmudrii: True, fixed.

pkg/templates/kubeadm/v1beta4/kubeadm.go
xmudrii: Exactly, the comment is now updated.

pkg/templates/kubeadm/v1beta4/kubeadm.go
xmudrii: No, it should return an error, panic doesn't fail nicely and we have pretty neat error handling for this. Shouldn't even be possible though, more like fail-safe mechanism.

xrstf: So now instead of doing either of the two correct things, we continue to just return a wrong value and continue on the happy path?

I understand that the IPFamily type should never ever have a value outside of the 4 that are listed. Which is exactly why the default case should be a panic, because there is something deeply wrong with the code if we hit that default case.

xmudrii: https://github.com/kubermatic/kubeone/pull/3360

PR #3360, opened 2024-08-28
Add panic in joinSubnets function

xmudrii:

Follow up on https://github.com/kubermatic/kubeone/pull/3358#discussion_r1727487494

The `join` function (which has been renamed to `joinSubnets` in this PR) has a case that's impossible to hit unless there's a bug in the KubeOne code. As discussed with @xrstf, such cases should be handled with `panic`.

I also added a comment explaining where we got value for the `terminated-pod-gc-threshold` flag.

xref #3346 

**What type of PR is this?**

/kind cleanup

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```

/assign @xrstf


xrstf (approved):
/approve

PR #3364, opened 2024-08-30
Support Kubernetes 1.31: Update Cluster Autoscaler, AWS CCM, AWS CSI

mohamed-rafraf:
This PR updated the images and the addons for cluster autoscaler, aws ccm, and aws csi in order to support Kubernetes 1.31 

Fixes #

**What type of PR is this?**

/kind feature

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


addons/ccm-aws/ccm-aws.yaml
```diff
@@ -152,9 +152,6 @@ spec:
             - --configure-cloud-routes=false
             - --cloud-config=/etc/kubernetes/cloud/cloud-config
             - --v=2
-            {{- with .Params.CCM_CONCURRENT_SERVICE_SYNCS }}
-            - --concurrent-service-syncs={{ . }}
-            {{- end }}
```
xmudrii, line 157: We need this. Unfortunately we didn't find a good way to keep it across updates, so putting it back manually is needed 😅

addons/csi-aws-ebs/aws-csi.yaml
```diff
@@ -430,7 +460,7 @@ spec:
         runAsUser: 0
       containers:
         - name: ebs-plugin
-          image: "{{ .InternalImages.Get "AwsEbsCSI" }}"
+          image: {{ .InternalImages.Get "..." }}
```
xmudrii, line 463: We need to put actual values here (and in other similar places in this file). We should just leave the value that was there before.

xmudrii (changes requested):

xmudrii (approved):
/lgtm
/approve

PR #3370, opened 2024-09-13
Replace deprecated Hetzner (in fact, restricted) default cx21 instance type with cx22

kron4eg:
Hetzner has deprecated (does not allow to create new) cx21 instance types, so the default goes to the (same size) cx22.

**What type of PR is this?**
/kind regression

```release-note
[ACTION REQUIRED] Example Terraform configs for Hetzner are now using `cx22` instead of `cx21` instance type by default. If you use the new Terraform configs with an existing cluster, make sure to override the instance type as needed, otherwise your instances/cluster will be destroyed
```

**Documentation**:

```documentation
NONE
```


xmudrii:
/lgtm
/approve
/release-note-edit
```release-note
[ACTION REQUIRED] Example Terraform configs for Hetzner are now using `cx22` instead of `cx21` instance type by default. If you use the new Terraform configs with an existing cluster, make sure to override the instance type as needed, otherwise your instances/cluster will be destroyed
```

PR #3382, opened 2024-09-18
Fetch the latest Kubernetes version in tests

xmudrii:

We're hard coding the full Kubernetes version including the patch release in our tests at the moment. This means that once a new patch release is out, we have to manually update the patch release in tests and in all other places (e.g. periodics). This is error prone and requires manual action from maintainers.

To ensure that we test with the latest Kubernetes patch release available, this PR:

- changes test definitions to include only major and minor (i.e. no patch version)
- fetch the latest patch release using the stable version marker (`https://dl.k8s.io/release/stable-x.y.txt`) upon running the test

Given this, we're also dropping the patch version from the job and test names.

**What type of PR is this?**
/kind chore

There's a slim chance that we might not be able to run tests while the new patch release is being published. That's between the version marker being updated and the packages being published. However, if this happens, it would only take a few minutes. The worst case, we can try fixing this on the upstream side if it appears to be a big problem.

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/test/e2e/helpers.go
+++ b/test/e2e/helpers.go
@@ -489,6 +489,34 @@ func dynamicClientRetriable(t *testing.T, k1 *kubeoneBin) ctrlruntimeclient.Clie
 	return client
 }
 
+func latestUpstreamVersion(majorMinorVersion string) (string, error) {
+	majorMinorSemver := semver.MustParse(majorMinorVersion)
+
+	const urlTemplate = "https://dl.k8s.io/release/stable-%d.%d.txt"
+	downloadURL := fmt.Sprintf(urlTemplate, majorMinorSemver.Major(), majorMinorSemver.Minor())
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+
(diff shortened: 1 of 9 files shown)


test/e2e/helpers.go
```diff
@@ -489,6 +489,23 @@ func dynamicClientRetriable(t *testing.T, k1 *kubeoneBin) ctrlruntimeclient.Clie
...
+	markerUrl := fmt.Sprintf("https://dl.k8s.io/release/stable-%d.%d.txt", majorMinorSemver.Major(), majorMinorSemver.Minor())
+	resp, err := http.Get(markerUrl)
+	if err != nil {
+		return "", fmt.Errorf("fetching kubernetes version maker: %v", err)
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return "", fmt.Errorf("reading version marker http response: %v", err)
+	}
```
kron4eg, line 504: missing `defer resp.Body.Close()`

kron4eg (commented):
Very cool! It's so tedious to always update patch versions.

P.S.
Small comment on conde

PR #3383, opened 2024-09-19
Rerun failed Kubernetes conformance tests

### LINKED ISSUE #3287: Restart failed conformance tests in E2E tests

We often face issue where a couple of conformance tests fail due to the unstable infrastructure. This fails the whole E2E job and we have to restart it which takes a lot of time and resources, and especially slows down the testing phase.

### Solution details

If there are failed conformance tests in the E2E tests, those tests should be restarted once or twice to rule out potential infrastructure instability issues. A very basic implementation of this has been done in #3261

### Alternative approaches

```
/retest
```

### Use cases

Significantly speeds up the KubeOne testing phase (potentially talking about hours). 🚀

xmudrii:

Turns out, Sonobuoy can very easily restart failed conformance tests, at least in theory. Let's put that into the practice!

With this PR, failed conformance tests will be restarted 3 times. This way, we don't fail the whole job if some tests flakes, instead we try to run them again. Those flakes cannot be caused by KubeOne, those flakes can be only the result of the unstable infrastructure or a bug in the test itself.

This should save us quite a lot of time when testing PRs and releases. Although, we might need to increase timeouts, but let's first see how this works.

Fixes #3287 

**What type of PR is this?**
/kind failing-test

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```

/assign @kron4eg

### DIFF
--- a/test/e2e/helpers.go
+++ b/test/e2e/helpers.go
@@ -594,6 +594,10 @@ func waitKubeOneNodesReady(ctx context.Context, t *testing.T, k1 *kubeoneBin) {
 }
 
 func sonobuoyRun(ctx context.Context, t *testing.T, k1 *kubeoneBin, mode sonobuoyMode, proxyURL string) {
+	sonobuoyRunWithRunCount(ctx, t, k1, mode, proxyURL, 0)
+}
+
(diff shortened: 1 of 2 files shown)


kron4eg (approved):
/approve

PR #3392, opened 2024-09-24
Add device_ownership_from_security_context containerd CRI config option

### LINKED ISSUE #3395: Add device_ownership_from_security_context containerd CRI config option

 The goal is to add support for the `device_ownership_from_security_context` flag in the containerd config option

### Solution details

This feature is important for multi-tenant Kubernetes clusters or workloads that rely on fine-grained security policies. It provides an additional layer of isolation between containers by enforcing ownership rules directly at the device level, which is critical for controlling access to sensitive devices (e.g., GPUs, volumes, etc.).

kron4eg:

Fixes #3395

**What type of PR is this?**
/kind feature

```release-note
Add `.containerRuntime.containerd.deviceOwnershipFromSecurityContext` option to the KubeOneCluster API used to enable/disable `device_ownership_from_security_context` option in the containerd configuration. This option is at the moment only applied to the control plane nodes and the static worker nodes. This field is set to false by default in the KubeOneCluster v1beta2 API, and to true by default in the KubeOneCluster v1beta3 API. If you don't want to use this option, we recommend explicitly disabling it to avoid any potential issues after migrating to the v1beta3 API
```

**Documentation**:

```documentation
TBD
```


pkg/apis/kubeone/types.go
```diff
@@ -194,6 +194,8 @@ type ContainerRuntimeConfig struct {
 type ContainerRuntimeContainerd struct {
 	// A map of registries to use to render configs and mirrors for containerd registries
 	Registries map[string]ContainerdRegistry `json:"registries,omitempty"`
+	// Enable or disable device_ownership_from_security_context containerd CRI config. Default to true.
+	DeviceOwnershipFromSecurityContext *bool `json:"device_ownership_from_security_context,omitempty"`
```
xmudrii, line 198: Can we please use camelCase instead because we use it everywhere else?

xmudrii (changes requested):

pkg/apis/kubeone/types.go
kron4eg: oh yeah, it was a copypaste from toml :shrug:

xmudrii:
/docs tbd
/release-note-edit
```release-note
Add `.containerRuntime.containerd.deviceOwnershipFromSecurityContext` option to the KubeOneCluster API used to enable/disable `device_ownership_from_security_context` option in the containerd configuration. This option is at the moment only applied to the control plane nodes and the static worker nodes. This field is set to false by default in the KubeOneCluster v1beta2 API, and to true by default in the KubeOneCluster v1beta3 API. If you don't want to use this option, we recommend explicitly disabling it to avoid any potential issues after migrating to the v1beta3 API
```

xmudrii (approved):
:shipit: 
/lgtm
/approve

Issue #3366, opened 2024-09-03
Validation permits having cloudProvider.none and machineController.deploy=true


kron4eg:
### What happened?
Sometimes people can be confused and try to set unsupported config combination that lead to very obscure error messages, for example in this case:
```
Task failed, error was: runtime: unmarshalling manifest "deployment-controller.yaml"
error converting YAML to JSON: yaml: line 64: could not find expected ':'
```

### Expected behavior

KubeOne should error out early, clearly indicating that such a config is not supported.

### What KubeOne version are you using?
Any of them.

### Provide your KubeOneCluster manifest here (if applicable)

<details>

```yaml
apiVersion: kubeone.k8c.io/v1beta2
kind: KubeOneCluster
name: kubeone
versions:
  kubernetes: 1.29.7

cloudProvider:
  none: {}

controlPlane:
  hosts:
    - privateAddress: 10.xxx
    - privateAddress: 10.xxx
    - privateAddress: 10.xxx

apiEndpoint:
  host: 10.xxx
  alternativeNames:
    - seed.xxx
  port: 6443

machineController:
  deploy: true
operatingSystemManager:
  deploy: true

```

</details>

xmudrii:
> KubeOne should error out early, clearly indicating that such a config is not supported.

This config should be supported, but we shouldn't deploy any credentials.

kron4eg:
> > KubeOne should error out early, clearly indicating that such a config is not supported.
> 
> This config should be supported, but we shouldn't deploy any credentials.

How so? What's the purpose of crenetial-less cloud-less machine-controller and osm?

xmudrii:
@kron4eg machine-controller has more providers than what we support in KubeOne. The most notable example is Kubevirt, I could create a KubeOne cluster, deploy machine-controller and OSM, deploy credentials, and then use the machine-controller to create Kubevirt-based nodes.

kron4eg:
In that case we should start supporting kubevirt.

PR #3416, opened 2024-10-25
Initial Kubevirt provider implementation

xmudrii:

This PR implements the initial and very minimal Kubevirt provider for KubeOne. The purposes of this implementation is to mitigate the issue outlined in #3366, i.e. to allow using machine-controller and operating-system-manager for Kubevirt clusters.

I initially though about reading supported environment variables and putting them in the secret if cloud provider is none, but it has two very big problems:

- We would start deploying arbitrary environment variables to the cluster, that are very likely not even related to that cluster
- We don't have a way to validate what environment variables are needed, e.g. there's no way to know that we need `KUBEVIRT_KUBECONFIG` for the said cluster that uses cloud provider none
  - This can make a huge trouble if someone runs `kubeone apply` without setting the needed environment variables, in which case the value would be removed from the cluster and machine-controller would start failing

Because of the given reasons, I thought that it's much safer to implement a very minimalistic provider for Kubevirt that doesn't have Terraform integration and support for CCM/CSI, but that can be used as a way to deploy machine-controller and operating-system-manager in a safe and reliable way.

xref #3366

**What type of PR is this?**
/kind feature

This is very initial Kubevirt provider implementation, we'll consider and discuss about extending it in future KubeOne releases.

```release-note
Initial implementation of Kubevirt provider for KubeOne. At the moment, the provider only supports deploying machine-controller and operating-system-manager for Kubevirt-based clusters. Terraform integration, CCM, CSI, and other cloud provider specific components are not supported and/or deployed automatically.
```

**Documentation**:
```documentation
TBD
```

/assign @kron4eg @moadqassem 
/hold for discussion and lazy consensus


moadqassem (approved):
/approve
/lgtm

xmudrii:
/hold cancel
:shipit:

PR #3458, opened 2024-11-12
Fix some of bugs discovered in the KubeOne 1.9 testing phase

xmudrii:

- Properly handle AppArmor policies for Cilium
  - The latest Cilium manifests and Helm chart are using the `appArmorProfile` field in `securityContext` to set the needed AppArmor profile. However, this field is only available starting with Kubernetes v1.30. Prior versions shall still use annotations, which is ensured by this change
- Initially deploy CSI Snapshotter webhook with `failurePolicy: Ignore`
  - This is a tricky one. First of all, it's recommended by [the CSI Snapshotter maintainers](https://github.com/kubernetes-csi/external-snapshotter/blob/d5c03dbde7ca6664707c05870eae3de4255766fb/deploy/kubernetes/webhook-example/admission-configuration-template#L21) to do it that way, i.e. deploy with `failurePolicy: Ignore` then change to `failurePolicy: Fail`. Second, we have a problem in our flow, because we first deploy this webhook, then we deploy the built-in addons (including `default-storage-class`), and then at the end we deploy user-provided addons. This is a problem when using an external CNI. That's because KubeOne will deploy the CSI Snapshotter webhook, which will wait for CNI. When trying to deploy `default-storage-class` addon, it would fail to do so, because the webhook is not running. CNI is not yet deployed, it's supposed to be deployed at the very end. To avoid this situation, we allow the webhook to run in the Ignore mode, and then change it to the Fail mode on subsequent `kubeone apply` run
- Fix GCP Ubuntu 24.04 image
- Increase context timeout in tests to 120m
  - This was the cause of flakes we were observing recently. The context had a timeout out 1 hour - 5 minutes, but tests have timeout of 2 hours by default. Because the context would timeout, we weren't able to continue running tests, even though the test itself wasn't supposed to timeout.

**What type of PR is this?**
/kind bug

All these changes are already thoroughly tested within testing PRs.

There still some bugs affecting the example Terraform configs for Azure, those are yet to be fixed in a follow up PR.

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```

/assign @kron4eg

### DIFF
--- a/addons/cni-cilium/cilium.yaml
+++ b/addons/cni-cilium/cilium.yaml
@@ -1,3 +1,4 @@
+{{ $kubernetesVersion := semver .Config.Versions.Kubernetes }}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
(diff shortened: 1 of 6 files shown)


xmudrii:
@kron4eg There was a failing test because I forgot to run `update-codegen`, should be good now.

PR #3476, opened 2024-12-05
return empty list instead of nil to not break the ui

soer3n:

If a kubeone cluster is only configured with static workers the ui is not rendered properly because the machinedeployment struct used for templating is nil. Instead an empty list should be returned by the function which fetches the machinedeployment list.

Fixes #

**What type of PR is this?**

/kind bug

```release-note
bug: fixed ui rendering when only static workers configured
```

**Documentation**:

```documentation
NONE
```


xmudrii:
/release-note-edit
```release-note
Fix an error message appearing in the KubeOne UI for clusters that don't have any Machine/MachineDeployment
```

PR #3524, opened 2025-01-08
Fix APT/RPM PGP key expiration issue for old kubernetes versiosn

### LINKED ISSUE #3460: Issue upgrading Kubernetes cluster from 1.26 to 1.27 due to expired APT keys
I'm experiencing an issue while upgrading my Kubernetes cluster from version 1.26 to 1.27 with KubeOne. The old APT keys required for the upgrade have expired, blocking the process. I see that a fix is available using a new key, as discussed in [kubernetes/release issue #3818](https://github.com/kubernetes/release/issues/3818), but this update hasn’t been integrated into KubeOne yet. Could you please update the key in KubeOne or suggest a workaround?

Thanks for your help!

mohamed-rafraf:
This PR addresses the issue of expired GPG keys for Kubernetes apt/rpm repositories by dynamically fetching the GPG key corresponding to the latest stable Kubernetes version.

According to this comment `https://github.com/kubernetes/release/issues/3818#issuecomment-2455160940` 
We can use the gpg key from the latest kubernetes repository.
 
Fixes #3460

**What type of PR is this?**

/kind bug

```release-note
Use the GPG key from the latest Kubernetes package repository to fix failures to install older versions of Kubernetes packages
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/scripts/os_amzn.go
+++ b/pkg/scripts/os_amzn.go
@@ -44,7 +44,7 @@ echo -n "${yum_proxy}" >> /tmp/yum.conf
 sudo mv /tmp/yum.conf /etc/yum.conf
 
 {{ if .CONFIGURE_REPOSITORIES }}
-LATEST_STABLE=$(curl -s https://cdn.dl.k8s.io/release/stable.txt | sed 's/\.[0-9]*$//')
+LATEST_STABLE=$(curl -sL https://dl.k8s.io/release/stable.txt | sed 's/\.[0-9]*$//')
 cat <<EOF | sudo tee /etc/yum.repos.d/kubernetes.repo
 [kubernetes]
 name=Kubernetes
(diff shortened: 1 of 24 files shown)


pkg/scripts/os_amzn.go
```diff
@@ -44,13 +44,14 @@ echo -n "${yum_proxy}" >> /tmp/yum.conf
 sudo mv /tmp/yum.conf /etc/yum.conf
 
 {{ if .CONFIGURE_REPOSITORIES }}
+LATEST_STABLE=$(curl -s https://cdn.dl.k8s.io/release/stable.txt | sed 's/\.[0-9]*$//')
```
xmudrii, line 47: Let's not use `cdn.dl.k8s.io` directly, but instead use just `dl.k8s.io`, to be on the safe side.

xmudrii (changes requested):
You also need to update fixtures by running `./hack/update-fixtures.sh`

xmudrii (approved):
Thank you so much for working on this one!
/lgtm
/approve

xmudrii:
/release-note-edit
```release-note
Use the GPG key from the latest Kubernetes package repository to fix failures to install older versions of Kubernetes packages
```

PR #2873, opened 2023-08-28
Migrate to the community-hosted package repositories (`pkgs.k8s.io`) and to `dl.k8s.io`

xmudrii:
**What this PR does / why we need it**:

There are three changes introduced in this PR:

- Migrate to the community-hosted package repositories (`pkgs.k8s.io`) from the deprecated Google-hosted package repository (`apt.kubernetes.io` and `yum.kubernetes.io`)
  - The Kubernetes project strongly encourages and recommends migrating to the new package repositories as soon as possible
  - Given that the Google-hosted repository is deprecated, I decided to do unconditional migration, i.e. without giving an option to opt-out from this change. This is going to make it easier for us to maintain and support this release branch in the long term. Moreover, most users are not going to notice this change at all. This change might only affect users allowlisting traffic based on IP addresses or URLs, in which case they need to update their allowlists (this is explained in the release note)
-  Migrate to `dl.k8s.io` instead of pulling binaries from the release bucket directly
  - The `kubernetes-release` bucket might get deprecated and frozen in the future as part of efforts to migrate to the community-owned infrastructure. The Kubernetes project strongly recommends using `dl.k8s.io` going forward
  - Same as the previous change, this change shouldn't affect many users. Additionally, this only affects Flatcar-based clusters. Users using allowlisting might need to update their configuration and this is explained in the release note
- cri-tools is updated to v1.27.1 for clusters running Kubernetes 1.27. Kubernetes 1.28 clusters will use cri-tools v1.28.0

Additionally, tests are updated to use KubeOne 1.6 as the stable version for initially provisioning the cluster. 

**Which issue(s) this PR fixes**:
Fixes #2870 

**What type of PR is this?**

/kind feature

**Does this PR introduce a user-facing change? Then add your Release Note here**:
```release-note
- [ACTION REQUIRED] Migrate from the Google-hosted package repositories (`apt.kubernetes.io` and `yum.kubernetes.io`) to the community-hosted package repositories. The Google-hosted package repositories are deprecated as of August 15, 2023 and using the community-hosted package repositories is strongly encouraged. For most users, this change shouldn't make a difference. However, if you do IP-based or URL-based allowlisting, you need to allow outgoing traffic to `pkgs.k8s.io` and `prod-cdn.packages.k8s.io`. For more details, see the official announcement: https://kubernetes.io/blog/2023/08/15/pkgs-k8s-io-introduction/
[...]

PR #3536, opened 2025-01-16
Always  unlock system components versions during our changes.

### LINKED ISSUE #1578: Tolerate held packages during cluster repair
**What happened**: After one of our control planes in the staging cluster failed, I followed the [cluster repair guide](https://docs.kubermatic.com/kubeone/v1.3/guides/manual_cluster_repair/) in order to set up a new node:

- `kubectl exec`'d into a working node
- removed the `NotReady` node from etcd ring
- ran `terraform apply` so that a new server was created
- ran `kubeone apply` in order to install kubernetes on it and let it join the node into the cluster

But then, on the existing control planes, the following error occured (error msg unescaped and shortened for better readability):

```
failed to install kubeadm: + export PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/sbin:/usr/local/bin:/opt/bin
+ PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/sbin:/usr/local/bin:/opt/bin
[...]
+ sudo add-apt-repository 'deb https://download.docker.com/linux/ubuntu focal stable'
+ sudo apt-get install -y 'containerd.io=1.4.*'
E: Held packages were changed and -y was used without --allow-change-held-packages.
: Process exited with status 100
```
This seems to happen because there are several packages held back by default:
```
# apt-mark showhold
containerd.io
kubeadm
kubectl
kubelet
kubernetes-cni
```
When I "unheld" them by running `apt-mark unhold containerd.io kubeadm kubectl kubelet kubernetes-cni`, everything worked as expected (and described in the guide).

You should either add this step to the guide or implement toleration of held packages (e.g. use `--allow-change-held-packages`).

`containerd.io` was `1.4.9` before, which is matched by `'containerd.io=1.4.*'`. Maybe that caused the problem in the first place?

KubeOne version (`kubeone version`):
```
{
  "kubeone": {
    "major": "1",
    "minor": "3",
    "gitVersion": "1.3.0",
    "gitCommit": "bfe6683334acdbb1a1d9cbbb2d5d5095f6f0111e",
    "gitTreeState": "",
    "buildDate": "2021-09-15T06:03:30Z",
    "goVersion": "go1.16.7",
    "compiler": "gc",
    "platform": "linux/amd64"
  },
  "machine_controller": {
    "major": "1",
    "minor": "35",
    "gitVersion": "v1.35.2",
    "gitCommit": "",
    "gitTreeState": "",
    "buildDate": "",
    "goVersion": "",
    "compiler": "",
    "platform": "linux/amd64"
  }
}
```
Operating system: Alpine (Docker image: `hashicorp/terraform:1.0.8` + kubeone installed with `wget`)

kron4eg:
Since we moved to the [versioned repositories](https://github.com/kubermatic/kubeone/pull/2873) we don't need those locks to be only removed on upgrades (or force).

Fixes #1578

**What type of PR is this?**
/kind bug

```release-note
Always  unlock system components versions during our changes.
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/scripts/os_amzn.go
+++ b/pkg/scripts/os_amzn.go
@@ -142,9 +142,7 @@ rm /tmp/k8s-binaries/kubectl
 {{- end }}
 
 {{ if .USE_KUBERNETES_REPO }}
-{{- if or .FORCE .UPGRADE }}
-sudo yum versionlock delete kubelet kubeadm kubectl kubernetes-cni cri-tools || true
-{{- end }}
+sudo yum versionlock delete kubelet kubeadm kubectl kubernetes-cni cri-tools
 
 # Amazon Linux 2 repos include the cri-tools package. These AL2 repos have higher
 # priority over the Kubernetes repos, so it's not possible to install cri-tools
@@ -165,8 +163,9 @@ sudo yum install -y --disableplugin=priorities \
 {{- end }}
 	kubernetes-cni \
 	cri-tools
-sudo yum versionlock add kubelet kubeadm kubectl kubernetes-cni cri-tools
 {{- end }}
+
+sudo yum versionlock add kubelet kubeadm kubectl kubernetes-cni cri-tools
 
 sudo systemctl daemon-reload
 sudo systemctl enable --now kubelet
@@ -179,7 +178,7 @@ sudo systemctl restart kubelet
 	removeBinariesAmazonLinuxScriptTemplate = `
 sudo systemctl stop kubelet || true
 
-sudo yum versionlock delete kubelet kubeadm kubectl kubernetes-cni cri-tools || true
+sudo yum versionlock delete kubelet kubeadm kubectl kubernetes-cni cri-tools
 sudo yum remove -y \
 	kubelet \
 	kubeadm \
--- a/pkg/scripts/os_centos.go
+++ b/pkg/scripts/os_centos.go
@@ -89,9 +89,7 @@ sudo systemctl enable --now iscsid
 {{ template "yum-containerd" . }}
 {{ end }}
 
-{{- if or .FORCE .UPGRADE }}
-sudo yum versionlock delete kubelet kubeadm kubectl kubernetes-cni cri-tools || true
-{{- end }}
+sudo yum versionlock delete kubelet kubeadm kubectl kubernetes-cni cri-tools
 
 sudo yum install -y \
 {{- if .KUBELET }}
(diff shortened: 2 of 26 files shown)


pkg/scripts/os_amzn.go
```diff
@@ -165,9 +163,10 @@ sudo yum install -y --disableplugin=priorities \
 {{- end }}
 	kubernetes-cni \
 	cri-tools
-sudo yum versionlock add kubelet kubeadm kubectl kubernetes-cni cri-tools
 {{- end }}
 
+sudo yum versionlock add kubelet kubeadm kubectl kubernetes-cni cri-tools
```
xmudrii, line 168: Why is this line moved? If we remove the lock only if the Kubernetes repo are used, why do we add the lock even if the Kubernetes repo are not used?

_note: this variable is always `true` at the moment, but this might change very soon._

pkg/scripts/os_amzn.go
```diff
@@ -179,7 +178,7 @@ sudo systemctl restart kubelet
 	removeBinariesAmazonLinuxScriptTemplate = `
 sudo systemctl stop kubelet || true
 
-sudo yum versionlock delete kubelet kubeadm kubectl kubernetes-cni cri-tools || true
```
xmudrii, line 182: The reason why we have this `|| true` is that `yum versionlock delete` fails if there's no versionlock set on packages. I think it makes sense to leave it in place. Is there any concrete reason why are we removing this?

xmudrii (changes requested):
I'm not really in favor of this PR, I find it to be a hacky solution for the referenced problem. This is a really nice safeguard to avoid accidentally bumping up the Kubernetes version. I think we should still leave it in place, but try to figure out what was the original problem and fix it in some other way.

pkg/scripts/os_amzn.go
kron4eg: So that it always adds version lock back, regardless of conditions. it's like a defer.

pkg/scripts/os_amzn.go
kron4eg: Turns out, `yum versionlock delete` do not fail, even if packages do not exist :shrug:.

image

PR #3537, opened 2025-01-16
Make sure no slash at the end of vcenterPrefix

### LINKED ISSUE #3473: vSphere credentials Secret creation fails if `VSPHERE_SERVER` contains a trailing slash
### What happened?

If `VSPHERE_SERVER` environment variable contains a trailing slash, creating vSphere credentials Secret fails because `/` is not an allowed character in a key name.

The key names in the vSphere credentials Secret are using the pattern of `{{ .VSPHERE_SERVER }}.{{ .username}}` and `{{ .VSPHERE_SERVER }}.{{ .password}}`. If `VSPHERE_SERVER` has a trailing slash, the generated key name would be `<server>/.username` which is not allowed by Kubernetes.

https://github.com/kubermatic/kubeone/blob/e43259aaec109a313288928ad3c0569a3dfda68a/pkg/credentials/secret.go#L213-L218

### Expected behavior

There are two possible options:

- Reject `VSPHERE_SERVER` values that have a trailing slash
- Drop a trailing slash in the linked function

- Set `VSPHERE_SERVER` environment variable to a value that contains a trailing slash
- Try to provision or upgrade a cluster

### What cloud provider are you running on?

vSphere

### Additional information

This code might have another bug. We drop `https://` prefix, but not `http://`, which might be a valid prefix too (we should verify that).

kron4eg:
Drops the slash at the `VSPHERE_SERVER`.

Fixes #3473

**What type of PR is this?**
/kind bug

```release-note
Make sure no slash at the end of VSPHERE_SERVER
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/credentials/secret.go
+++ b/pkg/credentials/secret.go
@@ -211,6 +211,7 @@ func vsphereSecret(credentials map[string]string) *corev1.Secret {
 	vscreds := map[string]string{}
 
 	vcenterPrefix := strings.ReplaceAll(credentials[VSphereAddressMC], "https://", "")
+	vcenterPrefix, _ = strings.CutSuffix(vcenterPrefix, "/")
 	// Save credentials in Secret and configure vSphere cloud controller
 	// manager to read it, in replace of storing those in /etc/kubernates/cloud-config
 	// see more: https://vmware.github.io/vsphere-storage-for-kubernetes/documentation/k8s-secret.html


xmudrii (approved):
/lgtm
/approve

xmudrii:
/release-note-edit
```release-note
Drop trailing slash from the `VSPHERE_SERVER` variable to ensure compatibility with machine-controller and vSphere CCM and CSI
```

Issue #3559, opened 2025-02-11
Support Kubernetes 1.32


kron4eg:
This is a collector issue for Kubernetes 1.3x support in KubeOne. The following tasks should be taken care of:

* [x] Check the Kubernetes changelog to ensure there are no breaking changes and removals
* [x] Update the `build` image to update Sonobuoy (and other dependencies if needed) 
* [x] Update the latest supported Kubernetes version in [the API validation](https://github.com/kubermatic/kubeone/blob/main/pkg/apis/kubeone/validation/validation.go#L40-L45) 
* [x] Update [default admission controllers](https://github.com/kubermatic/kubeone/blob/main/pkg/kubeflags/data.go) if needed 
* [x] Update `pause` image version in `pkg/apis/kubeone/helpers.go` if needed 
* [x] Add the appropriate `cri-tools` version in `pkg/scripts/os_flatcar.go` 
* [x] Update [the stable version marker in Makefile](https://github.com/kubermatic/kubeone/blob/5273f9a372736569c6b09b38f2959019d29e4d6a/Makefile#L24) 
* [x] Add E2E tests inside `tests.yml` 
* [x] Update daily periodics to use the latest Kubernetes release
* [x] Update [the Compatibility Matrix](https://docs.kubermatic.com/kubeone/main/architecture/compatibility/supported-versions/) 
* [x] Create an issue to track updating [images](https://github.com/kubermatic/kubeone/blob/main/pkg/templates/images/images.go) 

**Action items:**

```[tasklist]
### Tasks
- [ ] Update this and update images templates to reflect any revisions
```

PR #3565, opened 2025-02-24
Initial Support for Kubernetes 1.32

kron4eg:
Initial support

Xref #3559 

**What type of PR is this?**
/kind feature

```release-note
Initial Support for Kubernetes 1.32
```

**Documentation**:

```documentation
NONE
```


kron4eg:
> [ERROR SystemVerification]: kernel release 4.14.355-275.582.amzn2.x86_64 is unsupported. Recommended LTS version from the 4.x series is 4.19. Any 5.x or 6.x versions are also supported. For cgroups v2 support, the minimal version is 4.15 and the recommended version is 5.8+\n

@xmudrii check this out. Amzn2 looks dead at this point.

pkg/apis/kubeone/validation/validation.go
```diff
@@ -42,7 +42,7 @@ var (
 	// minVersionConstraint defines the minimum Kubernetes version supported by KubeOne
 	minVersionConstraint = semverutil.MustParseConstraint(">= 1.29")
```
xmudrii, line 43: Let's bump this to 1.30, this can be done in a follow up as well.

pkg/scripts/os_flatcar.go
```diff
@@ -251,6 +251,8 @@ func criToolsVersion(cluster *kubeoneapi.KubeOneCluster) string {
 		return "1.30.1"
 	case 31:
 		return "1.31.1"
+	case 32:
+		return "1.31.1"
```
xmudrii, line 255: This should be [1.32.0](https://github.com/kubernetes-sigs/cri-tools/releases/tag/v1.32.0).

pkg/kubeflags/data.go
```diff
@@ -30,7 +30,6 @@ var defaultAdmissionControllers = []string{
 	"RuntimeClass",
 	"CertificateApproval",
 	"CertificateSigning",
-	"ClusterTrustBundleAttest",
```
xmudrii, line 33: Why is this admission controller removed? I still see it as enabled by default in the API server.

xmudrii (changes requested):

pkg/apis/kubeone/validation/validation.go
kron4eg: :heavy_check_mark:

pkg/scripts/os_flatcar.go
kron4eg: :heavy_check_mark:

pkg/kubeflags/data.go
kron4eg: In some doc on kubernetes.io regarding admission controllers it was missing from the list. But since it's present in CLI reference I guess that list is the correct list. Returned it back: :heavy_check_mark:

PR #3574, opened 2025-03-04
Use mirror.gcr.io as a default mirror for docker.io

kron4eg:
Docker registry is about to introduce new limitation: https://www.docker.com/blog/revisiting-docker-hub-policies-prioritizing-developer-experience/. 

**What type of PR is this?**
/kind feature

```release-note
Use mirror.gcr.io as a default mirror for docker.io
```

**Documentation**:

```documentation
NONE
```


xmudrii:
I have two concerns about this change:
- I can't find anything about `mirror.gcr.io`, is it a service provided by Google or by some user? What are terms of usage for non-GKE/non-Google users, can we expect any problems there? Are all Docker Hub images available there?
- This is kind of a breaking change, so an action required release note shall be used

kron4eg:
We will replace this with the documentation.
/close

PR #3627, opened 2025-04-04
Fix flatcar script

kron4eg:

Fixes #

**What type of PR is this?**
/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


pkg/scripts/os_flatcar.go
```diff
@@ -207,37 +155,6 @@ func RemoveBinariesFlatcar() (string, error) {
...
-	if err := containerruntime.UpdateDataMap(cluster, data); err != nil {
-		return "", err
-	}
-
-	result, err := Render(upgradeKubeletAndKubectlFlatcarScriptTemplate, data)
-
-	return result, fail.Runtime(err, "rendering upgradeKubeletAndKubectlFlatcarScriptTemplate script")
-}
-
 const defaultKubernetesCNIVersion = "1.3.0"
```
xmudrii, line 158: That looks old, we should revisit it at some point soon.

xmudrii (approved):
/lgtm
/approve
with a nit for a follow up

pkg/scripts/os_flatcar.go
kron4eg: It's still recommended version in the [upstream docs](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#installing-kubeadm-kubelet-and-kubectl)

xmudrii: Docs are wrong, it is supposed to be 1.6.0 with v1.32: https://github.com/kubernetes/kubernetes/blob/cdc807a9e849b651fb48c962cc18e25d39ec5edf/build/dependencies.yaml#L19-L30

You can check this file on each branch for the recommended version, this is the actual version that Kubernetes has been tested with.

kron4eg: Oh... well, shit. We should fix this.

PR #3634, opened 2025-04-10
Fix digitalocean terraform dpkg lock file issue

kron4eg:

* When DO droplets are started they automatically try to `apt update`, at the same time kubeone also doing just they effectively causing race-condition. This PR introduces a timeout to wait before kubeone kicks in.
* Downgrade digitalocean CCM to v0.1.56
Since v0.1.57 this CCM requires kubelet flag --provider-id=digitalocean://<droplet ID>
See more: https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/v0.1.59/docs/getting-started.md#--provider-iddigitaloceandroplet-id

**What type of PR is this?**
/kind bug

```release-note
* Fix digitalocean terraform dpkg lock file issue
* Downgrade digitalocean CCM to v0.1.56
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/addons/ccm-digitalocean/Kustomization
+++ b/addons/ccm-digitalocean/Kustomization
@@ -3,7 +3,7 @@ kind: Kustomization
 namespace: kube-system
 
 resources:
-  - https://raw.githubusercontent.com/digitalocean/digitalocean-cloud-controller-manager/master/releases/digitalocean-cloud-controller-manager/v0.1.59.yml
+  - https://raw.githubusercontent.com/digitalocean/digitalocean-cloud-controller-manager/master/releases/digitalocean-cloud-controller-manager/v0.1.56.yml
 
 patches:
   - patch: |-
--- a/examples/terraform/digitalocean/README.md
+++ b/examples/terraform/digitalocean/README.md
@@ -19,6 +19,7 @@ use the configs and how to provision a Kubernetes cluster using KubeOne.
 | Name | Version |
 |------|---------|
 | <a name="provider_digitalocean"></a> [digitalocean](#provider\_digitalocean) | ~> 2.9.0 |
+| <a name="provider_time"></a> [time](#provider\_time) | n/a |
 
 ## Modules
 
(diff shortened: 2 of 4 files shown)


xmudrii:
> When DO droplets are started they automatically try to apt update, at the same time kubeone also doing just they effectively causing race-condition. This PR introduces a timeout to wait before kubeone kicks in.

Why we don't check for the existence of the lock file in KubeOne or just let it fail? If we check for the lock file existence, we have a more reliable, future-proof solution that works for every provider. If we just let it fail, wouldn't KubeOne retry a few times and eventually pick up from there automatically? Especially given we're talking about waiting 60 seconds, that should be very well within the KubeOne wait/retry mechanism and defaults.

> Downgrade digitalocean CCM to v0.1.56
Since v0.1.57 this CCM requires kubelet flag --provider-id=digitalocean://
See more: https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/v0.1.59/docs/getting-started.md#--provider-iddigitaloceandroplet-id

My understanding of the provided link is that it's optional, if you don't provide it, the CCM will still query the API but it might take a while if you have a lot of Droplets. I don't see them reverting this decision and we can't stay at the old version forever, so I don't think we should revert.

kron4eg:
> My understanding of the provided link is that it's optional, if you don't provide it, the CCM will still query the API but it might take a while if you have a lot of Droplets.

Your understanding is incorrect. In v0.1.57 they migrated to InstanceV2 API, which requires a droplet ID, and now they require `--provider-id=digitalocean://` kubelet flag. This is a breaking change for us. We will not stay on the outdated version forever, of course. This is just to get the release out of the gates, later we could think about how to pass this ID from Terraform state to KubeOne.



> that should be very well within the KubeOne wait/retry mechanism and defaults.

We retry of course, it's still fails way too often :man_shrugging:

xmudrii (approved):
/lgtm
/approve

PR #3632, opened 2025-04-07
CNI Version on Flatcar now depends on the kubernetes version

kron4eg:
**What type of PR is this?**
/kind bug

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change? Then add your Release Note here**:

```release-note
CNI Version on Flatcar now depends on the kubernetes version
```

**Documentation**:

```documentation
NONE
```

PR #3636, opened 2025-04-14
Fix flatcar 1.31.x CNI plugins directory permissions

kron4eg:
When we upgraded to using CNI per kubernetes version on Flatcar in #3632, turns out that CNI plugins tar archive for 1.31 (v1.5.1) contained incorrect owner permissions on the root directory, which caused cilium init container `mount-cgroup` to fail with `permission denied` error message.

The same already been fixed in https://github.com/kubermatic/operating-system-manager/pull/379

**What type of PR is this?**
/kind bug

```release-note
Fix flatcar 1.31.x CNI plugins directory permissions
```

**Documentation**:

```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #3648, opened 2025-04-25
Make kubeone go installa-ble again

kron4eg:
Latest KubeOne release fails to `go install` with the following 

```
$ go install -v k8c.io/kubeone@latest
go: downloading k8c.io/kubeone v1.10.0
go: k8c.io/kubeone@latest (in k8c.io/kubeone@v1.10.0):
        The go.mod file for the module providing named packages contains one or
        more replace directives. It must not contain directives that would cause
        it to be interpreted differently than if it were the main module.
```

There is no need in this `replace` anymore, seems like it was fixed upstream. Now all the codegen works perfectly fine. This change enables KubeOne to be `go install`-able again.

**What type of PR is this?**
/kind cleanup

```release-note
Make kubeone go installa-ble again
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -184,10 +184,3 @@ require (
 	sigs.k8s.io/kustomize/kyaml v0.18.1 // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.2 // indirect
 )
-
-// This is required because the Kubernetes code generator tools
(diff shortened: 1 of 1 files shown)


xrstf (approved):
/approve

PR #3661, opened 2025-05-21
Support KubeVirt CCM Deployment 

moadqassem:
This PR supports the deployment and the configurations for the KubeVirt CCM. 

Fixes #
/kind feature
**What type of PR is this?**

```release-note
Support KubeVirt CCM Deployment
```

**Documentation**:

```documentation
None
```


kron4eg (changes requested):
This PR needs some more:

* generate and update the Secret with CCM config (based on the provided kubeone manifest)
* enable the addon if kubevirt is a selected cloud provider
* preferably write a Kustomization file to generate ccm YAML automatically and add any customizations if necessary or at least have a README.md that describes "what, where and when" WRT to YAML manifest obtaining

addons/csi-kubevirt/2-kubevirt-csi-controller.yaml
```diff
@@ -72,21 +72,6 @@ roleRef:
...
-kind: Secret
-metadata:
-  labels:
-    app: kubevirt-csi-driver
-  name: infra-kubeconfig
-  namespace: kube-system
-type: Opaque
----
-{{ end }}
-{{ end }}
```
kron4eg, line 89: This secret it shared between CSI and CCM, so naturally it's not managed by the addons but rather by kubeone itself.

addons/ccm-kubevirt/config-map.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: cloud-config
+  namespace: kube-system
+data:
+  cloud-config: |
+    kubeconfig: /var/run/secrets/infracluster/KUBEVIRT_KUBECONFIG
+    namespace: {{ .Config.CloudProvider.KubeVirt.InfraNamespace }}
+    instancesV2:
+      zoneAndRegionEnabled: {{ .Config.CloudProvider.KubeVirt.ZoneAndRegionEnabled }}
+    loadBalancer:
+      enabled: {{ .Config.CloudProvider.KubeVirt.CCMLoadBalancerEnabled }}
```
xmudrii, line 13: This is a bit "NPE-unsafe" code, I'm not sure if we can trigger NPE somehow, but I think it's not a bad idea to guard it just in case, who knows.

pkg/apis/kubeone/types.go
```diff
@@ -463,6 +463,10 @@ type KubevirtSpec struct {
 	// InfraClusterKubeconfig is a base64-encoded kubeconfig file that points to a KubeVirt infra cluster (a cluster where KubeVirt is installed).
 	// This Kubeconfig will be used to create and manage KubeVirt specific resources such as DataVolumes.
 	InfraClusterKubeconfig string `json:"infraClusterKubeconfig,omitempty"`
+	// ZoneAndRegionEnabled indicates if need to get Region and zone labels from the cloud provider
+	ZoneAndRegionEnabled bool `yaml:"zoneAndRegionEnabled,omitempty"`
+	// CCMLoadBalancerEnabled indicates if the ccm should create and manage the clusters load balancers.
+	CCMLoadBalancerEnabled bool `json:"ccmLoadBalancerEnabled,omitempty"`
```
xmudrii, line 469: I'm wondering if we really need this `Enabled` suffix. The variable is already a bool, we can comment that setting true/false is enabling/disabling, but it should be already obvious.

addons/ccm-kubevirt/config-map.yaml
kron4eg: :heavy_check_mark: 
also added validation for required fields

pkg/apis/kubeone/types.go
kron4eg: The `zoneAndRegionEnabled` represents cloud-config `zoneAndRegionEnabled` flag, ubiquitous, domain specific language. 

But as of `ccmLoadBalancerEnabled`, I think we can drop the `ccm` prefix, as it's not in the cloud-config and it's obviously that LBs are managed by the CCM.

PR #3690, opened 2025-06-25
Fix the installation of kubeadm, kubelet and kubectl packages

archups:
This PR is to fix the k8s packages installation error encountered during k8s upgrade from lower version when tried to do using `kubeone apply` on the customer environment having Rocky Linux as OS. 

```
[10.17.187.121] All matches were filtered out by exclude filtering for argument: kubelet-1.30.13
[10.17.187.121] All matches were filtered out by exclude filtering for argument: kubeadm-1.30.13
[10.17.187.121] All matches were filtered out by exclude filtering for argument: kubectl-1.30.13
[10.17.187.121] Package kubernetes-cni-1.4.0-150500.1.1.x86_64 is already installed.
[10.17.187.121] Package cri-tools-1.30.1-150500.1.1.x86_64 is already installed.
[10.17.187.121] Error: Unable to find a match: kubelet-1.30.13 kubeadm-1.30.13 kubectl-1.30.13
ERRO[09:45:04 UTC] ssh: installing kubeadm and kubectl and kubelet
```

It is mentioned [here ](https://v1-30.docs.kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#k8s-install-1) in the official doc to define `exclude` parameter in the repository definition. 
`Add the Kubernetes yum repository. The exclude parameter in the repository definition ensures that the packages related to Kubernetes are not upgraded upon running yum update as there's a special procedure that must be followed for upgrading Kubernetes. Please note that this repository have packages only for Kubernetes 1.30; for other Kubernetes minor versions, you need to change the Kubernetes minor version in the URL to match your desired minor version (you should also check that you are reading the documentation for the version of Kubernetes that you plan to install).`

Fixes #

**What type of PR is this?**
/kind bug

```release-note
Fix the Kubernetes repository configuration on RHEL-like systems (RockyLinux, AmazonLinux) to allow package upgrades
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/scripts/os_amzn.go
+++ b/pkg/scripts/os_amzn.go
@@ -52,6 +52,7 @@ baseurl=https://pkgs.k8s.io/core:/stable:/{{ .KUBERNETES_MAJOR_MINOR }}/rpm/
 enabled=1
 gpgcheck=1
 gpgkey=https://pkgs.k8s.io/core:/stable:/${LATEST_STABLE}/rpm/repodata/repomd.xml.key
+exclude=kubelet kubeadm kubectl cri-tools kubernetes-cni
 EOF
 
 # We must clean 'yum' cache upon changing the package repository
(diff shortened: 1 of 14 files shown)


embik:
Just a minor comment, the release note isn't helpful to users and should be rephrased a bit.

kron4eg (changes requested):
Using `exclude` without `dnf --disableexcludes=kubernetes` will not work.

xmudrii:
/hold
for discussion

I don't see this working

dharapvj:
I just want to say that I also see this in RHEL installations. I end up running `yum versionlock clear
` on the machines where kubeone is stuck re-trying in another shell window to get past this problem but having this fixed in kubeone itself would be awesome!

PR #3773, opened 2025-08-01
kubeone config certificates renew

### LINKED ISSUE #2577: Add a subcommand to renew all the certificates

Add new subcommand, to forcefully regenerate the certificates even if they are still fresh (e.g. more then 90 days to expire).

### Solution details

- New subcommand to run `sudo kubeadm certs review all`
- No kubeapi requests should be attempted, only ssh, to avoid hitting the API with potentially expired certificates

### Use cases

- Regenerate non-expiring (means more then 90 until expiration) certificates when SAN is changed.
- Regenerate already expired certificates, basically reanimating cluster.

kron4eg:

Fixes #2577

**What type of PR is this?**
/kind feature

```release-note
Automate certificates renew
```

**Documentation**:

```documentation
TBD
```

### DIFF
--- a/pkg/certificate/certificate.go
+++ b/pkg/certificate/certificate.go
@@ -27,8 +27,13 @@ import (
 
 	"github.com/pkg/errors"
 
+	kubeoneapi "k8c.io/kubeone/pkg/apis/kubeone"
 	"k8c.io/kubeone/pkg/configupload"
+	"k8c.io/kubeone/pkg/executor"
 	"k8c.io/kubeone/pkg/fail"
+	"k8c.io/kubeone/pkg/runner"
+	"k8c.io/kubeone/pkg/scripts"
+	"k8c.io/kubeone/pkg/state"
 	"k8c.io/kubeone/pkg/templates/resources"
 
 	certutil "k8s.io/client-go/util/cert"
@@ -125,4 +130,35 @@ func GetCertificateSANs(host string, alternativeNames []string) []string {
 	}
 
 	return certSANS
+}
+
+func RenewAll(st *state.State) error {
+	return st.RunTaskOnControlPlane(func(ctx *state.State, node *kubeoneapi.HostConfig, _ executor.Interface) error {
+		logger := ctx.Logger.WithField("node", node.PublicAddress)
+		logger.Infoln("Renew certificates...")
+
+		_, _, err := ctx.Runner.RunRaw("sudo kubeadm certs renew all")
+		if err != nil {
+			return fail.SSH(err, "renewing certificates")
+		}
+
+		pods := []string{
+			"etcd",
+			"kube-apiserver",
+			"kube-controller-manager",
+			"kube-scheduler",
+		}
+
+		for _, pod := range pods {
+			logger.Infof("Restarting %s pod...", pod)
+			_, _, err := ctx.Runner.Run(scripts.RestartPodCrictlTemplate, runner.TemplateVariables{
+				"NAME": pod,
+			})
+			if err != nil {
+				return fail.SSH(err, "renewing certificates")
+			}
+		}
+
+		return nil
+	}, state.RunParallel)
 }
--- a/pkg/cmd/config-certificates.go
+++ b/pkg/cmd/config-certificates.go
@@ -1,0 +1,56 @@
+/*
+Copyright 2025 The KubeOne Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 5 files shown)


pkg/cmd/config-certificates.go
```diff
@@ -0,0 +1,56 @@
...
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	"k8c.io/kubeone/pkg/certificate"
+)
+
+func configCertificatesCmd(rootFlags *pflag.FlagSet) *cobra.Command {
```
xmudrii: `config` seems like a bad root command for this command. To me, `config` has always been about manipulating the KubeOne configuration (`kubeone.yaml`) and other relative configurations (MachineDeployments, list of images...).

What do you think about having this as a dedicated root command, e.g. `kubeone certs renew`? Or maybe even `kubeone apply renew-certs`?

pkg/cmd/config-certificates.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	cmd.AddCommand(configCertificatesRenewCmd(rootFlags))
+
+	return cmd
+}
+
+func configCertificatesRenewCmd(rootFlags *pflag.FlagSet) *cobra.Command {
+	return &cobra.Command{
+		Use:     "renew",
+		Short:   "renew all the ceritificates of the control-plane",
```
xmudrii:
```suggestion
		Short:   "renew all the certificates of the Kubernetes control plane",
```

pkg/certificate/certificate.go
```diff
@@ -126,3 +131,34 @@ func GetCertificateSANs(host string, alternativeNames []string) []string {
...
+			_, _, err := ctx.Runner.Run(scripts.RestartPodCrictlTemplate, runner.TemplateVariables{
+				"NAME": pod,
+			})
+			if err != nil {
+				return fail.SSH(err, "renewing certificates")
+			}
+		}
+
+		return nil
+	}, state.RunParallel)
```
xmudrii, line 163: I'd run this sequentially because it involves restarting pods, so restarting in parallel might affect the control plane availability.

pkg/certificate/certificate.go
```diff
@@ -126,3 +131,34 @@ func GetCertificateSANs(host string, alternativeNames []string) []string {
...
+			"kube-scheduler",
+		}
+
+		for _, pod := range pods {
+			logger.Infof("Restarting %s pod...", pod)
+			_, _, err := ctx.Runner.Run(scripts.RestartPodCrictlTemplate, runner.TemplateVariables{
+				"NAME": pod,
+			})
+			if err != nil {
+				return fail.SSH(err, "renewing certificates")
```
xmudrii:
```suggestion
				return fail.SSH(err, fmt.Sprintf("restarting pod %q after renewing certificates", pod))
```

xmudrii (changes requested):
A few nits otherwise LGTM!

pkg/certificate/certificate.go
kron4eg: Usually this command is needed when the control-plane is already dead, so it doesn't matter. And I've been testing this, practically no effect on the api availability as there are pauses 30 seconds between restarts.

pkg/cmd/config-certificates.go
kron4eg: Then `kubeone certs renew` it is!

pkg/cmd/certificates.go
```diff
@@ -0,0 +1,62 @@
...
+
+	return cmd
+}
+
+func certificatesRenewCmd(rootFlags *pflag.FlagSet) *cobra.Command {
+	return &cobra.Command{
+		Use:   "renew",
+		Short: "renew all the certificates of the Kubernetes control plane",
+		Long: heredoc.Doc(`
+			This command will run "kubeadm certs renew all" across control-plane VMs and restart control-plane pods after that.
```
xmudrii: Nit: it's `control plane`
```suggestion
			This command will run "kubeadm certs renew all" across control plane VMs and restart control plane pods after that.
```

ahmedwaleedmalik (approved):
/approve

Looks really solid, just one small nit; we should add a bit more details to the release note. 

Something like:
```
`kubeone certificates renew` command can be used to renew control plane certificates in a KubeOne cluster
```

xmudrii (commented):
Belated:
/lgtm
/approve
Thank you so much for adding this command! :shipit:

PR #3779, opened 2025-08-08
Make machine-controller -join-cluster-timeout configurable

### LINKED ISSUE #3777: Please make machine-controller's join-cluster-timeout configurable

As someone who has to run K1-provisioned k8s clusters on some IaaS stacks that have low I/O and sometimes very outdated OS images, I'd like to be able to configure the machine-controller's join-cluster-timeout, i.e. the following line through a parameter in the `KubeOneCluster` manifest:
https://github.com/kubermatic/kubeone/blob/main/addons/machinecontroller/deployment-controller.yaml#L55

Yes, 15 minutes usually enough, but honestly it's sometimes just barely enough (and might even take more than one attempt) on our worst environments - we just can't afford to keep every test environment on the same standard as our production ones. So yea, "upgrade on first boot" sometimes takes well over 10 minutes and essentially it's just a matter of time until the 15m timeout will cause us pain (and you know how it goes - it always happens in the worst case, e.g. when testing a hotfix for an incident, so I'd like to be prepared).

### Solution details

- Add a parameter to make machine-controller's join-cluster-timeout configurable
- Use that parameter in the addon on the line linked above

### Alternative approaches

We can just edit the deployment's manifest. But it will get overwritten next time `kubeone apply` is run, which isn't ideal.

### Additional information

Just FYI, I hope to also get this option added and configurable in KKP, see kubermatic/kubermatic#12066

kron4eg:
Some people need this timeout even longed compared to what we had hardcoded.

Fixes #3777

**What type of PR is this?**
/kind feature

```release-note
Make machine-controller -join-cluster-timeout configurable
```

**Documentation**:

```documentation
TBD
```

### DIFF
--- a/addons/machinecontroller/deployment-controller.yaml
+++ b/addons/machinecontroller/deployment-controller.yaml
@@ -52,7 +52,7 @@ spec:
             {{ if CABundle -}}
             - -ca-bundle={{ .Resources.CABundleSSLCertFilePath }}
             {{ end -}}
-            - -join-cluster-timeout=15m
+            - -join-cluster-timeout={{ default "15m" .Params.JOIN_TIMEOUT }}
           env:
   {{ with .Config.CloudProvider.Kubevirt -}}
   {{ with .InfraNamespace }}


ahmedwaleedmalik (approved):
/approve

PR #3780, opened 2025-08-12
refreshed cluster-autoscaler addon

dharapvj:
* Refreshed the cluster-autoscaler addon to have more deterministic and easy upgrade path in future
* Added utilization threshold override

Fixes # NA

**What type of PR is this?**

/kind chore

I am not sure how the upgrade would be handled. I have used this same yaml file in a KKP user-cluster and everything worked as expected. But I am not sure if old addon yaml content will be removed first and replaced with what is in new addon.yaml or not.

```release-note
cluster-autoscaler addon now supports new variable CLUSTER_AUTOSCALER_SCALE_DOWN_UTIL_THRESHOLD to control `--scale-down-utilization-threshold` parameter.
```

**Documentation**:

```documentation
None
```


addons/cluster-autoscaler/cluster-autoscaler.yaml
```diff
@@ -1,114 +1,227 @@
...
+# --set-string service.annotations."prometheus\.io/scrape"=true,service.annotations."prometheus\.io/port"=8085 | sed "s/x-k8s/k8s/g" | sed "/managed-by/d"
+
+# Must add/replace these in Deployment in the generated output for upgrades.
+        # - image: '{{ default (.InternalImages.Get "ClusterAutoscaler") .Params.CLUSTER_AUTOSCALER_IMAGE }}'
+        #   args:
+        #     - --skip-nodes-with-local-storage={{ default "true" .Params.CLUSTER_AUTOSCALER_SKIP_LOCAL_STORAGE }}
+        #     - --enforce-node-group-min-size={{ default "false" .Params.CLUSTER_AUTOSCALER_ENFORCE_NODE_GROUP_MIN_SIZE }}
+        #     - --balance-similar-node-groups={{ default "false" .Params.CLUSTER_AUTOSCALER_BALANCE_SIMILAR_NODE_GROUP }}
+        #     - --scale-down-utilization-threshold={{ default "0.5" .Params.CLUSTER_AUTOSCALER_SCALE_DOWN_UTIL_THRESHOLD }}
```
kron4eg: Instead of providing instructions please write a Kustomization file that will automate them. For inspiration PTAL at https://github.com/kubermatic/kubeone/blob/main/addons/ccm-aws/Kustomization

dharapvj: Perfect..  I tried searching if kustomization support helm chart but I could not find any quick example. But this is perfect. I have updated my code to use Kustomization

addons/cluster-autoscaler/cluster-autoscaler.yaml
```diff
@@ -1,212 +1,371 @@
...
+      - machinepools
       - machines
       - machinesets
     verbs:
       - get
       - list
       - update
       - watch
+  - apiGroups:
+      - cluster.x-k8s.io
```
kron4eg, line 203: @dharapvj apparently upstream chart doesn't have that rule. Please add the similar rule (via Kustmizefile) for `cluster.k8s.io` group as otherwise CA will loose ability to scale our machinedeployments.

addons/cluster-autoscaler/cluster-autoscaler.yaml
```diff
@@ -1,212 +1,371 @@
...
             - --cloud-provider=clusterapi
             - --namespace=kube-system
-            - --skip-nodes-with-local-storage={{ default "true" .Params.CLUSTER_AUTOSCALER_SKIP_LOCAL_STORAGE }}
             - --node-group-auto-discovery=clusterapi:namespace=kube-system
-            - --enforce-node-group-min-size={{ default "false" .Params.CLUSTER_AUTOSCALER_ENFORCE_NODE_GROUP_MIN_SIZE }}
-            - --balance-similar-node-groups={{ default "false" .Params.CLUSTER_AUTOSCALER_BALANCE_SIMILAR_NODE_GROUP }}
             - --logtostderr=true
             - --stderrthreshold=info
             - --v=4
+            - - --skip-nodes-with-local-storage={{ default "true" .Params.CLUSTER_AUTOSCALER_SKIP_LOCAL_STORAGE }}
```
kron4eg: no, this is wrong level of array

kron4eg (approved):
/approve

PR #3793, opened 2025-08-21
add option to enable non root device usage in worker nodes by osm

soer3n:

This pr introduces the option to enable non root device usage in container runtime by configuring a new introduced flag of the operating system manager.
ref: https://github.com/kubermatic/kubermatic/issues/14352

Fixes #

**What type of PR is this?**

/kind feature

```release-note
Non-root device usage on non-static worker nodes can now be enabled for containerd runtime by setting the value `operatingSystemManager.enableNonRootDeviceOwnership` to `true` when osm is enabled.

```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/apis/kubeone/v1beta3/types.go
+++ b/pkg/apis/kubeone/v1beta3/types.go
@@ -745,7 +745,7 @@ type MachineControllerConfig struct {
 type OperatingSystemManagerConfig struct {
 	// Deploy
 	Deploy bool `json:"deploy,omitempty"`
-	//  # Optional: EnableNonRootDeviceOwnership enables the non-root device ownership feature in the container runtime.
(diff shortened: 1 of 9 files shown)


pkg/apis/kubeone/v1beta3/types.go
```diff
@@ -745,6 +745,8 @@ type MachineControllerConfig struct {
 type OperatingSystemManagerConfig struct {
 	// Deploy
 	Deploy bool `json:"deploy,omitempty"`
+	//  # Optional: EnableNonRootDeviceOwnership enables the non-root device ownership feature in the container runtime.
```
kron4eg:
```suggestion
	// EnableNonRootDeviceOwnership enables the non-root device ownership feature in the container runtime.
```

API generator will output optionality for fields with omitempty. No need for additional "Optionals".

PR #3814, opened 2025-09-09
Add bastion SSH private key file setting in host config

### LINKED ISSUE #3153: No sshPrivateKeyFile option for bastion host
### Add bastionHostPrivateKeyFile field for host

As a user
I would like to be able to specify an private key file to use with other bastion host options.

### Solution details

- Add a field named bastionHostPrivateKeyFile in order to specify a ssh private key to used when connecting to the bastion host

### Use cases

Having a bastion host configured to accept only priv key authentication the only other possibility is to configure the .ssh/config file. That option isn't exactly the best one when dealing with automatic process.

kron4eg:

Fixes #3153

**What type of PR is this?**
/kind feature

```release-note
Add bastion SSH private key file setting in host config
```

**Documentation**:

```documentation
API doc
```

### DIFF
--- a/docs/api_reference/v1beta2.en.md
+++ b/docs/api_reference/v1beta2.en.md
@@ -1,6 +1,6 @@
 +++
 title = "v1beta2 API Reference"
-date = 2025-08-21T16:58:51+02:00
+date = 2025-09-09T17:07:06+03:00
 weight = 11
 +++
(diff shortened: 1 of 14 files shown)


ahmedwaleedmalik (approved):
/approve

PR #3815, opened 2025-09-09
Add certificateValidityPeriod and caCertificateValidityPeriod API

### LINKED ISSUE #3582: Configuring certificate validity period

As a KubeOne user, I'd like the ability to configure the Client Certificate expiration duration for the certificates that are generated by Kubeadm. 

Currently, KubeOne provides no way to update those certificates in an automated way, and `Kubeone apply` should be run regularly within a window of a specific time frame for those certificates to be rotated. Otherwise, manual rotation is required, which isn't ideal.

https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/#choosing-cert-validity-period

## Alternative

Automate the cert renewal process and let some agent sitting on the control plane nodes or within the cluster(?) rotate the certificates a month before the expiry. Since KubeOne fetches the KubeConfig from the CP nodes, after rotation users can use KubeOne to fetch the updated Kubeconfig.

However, even with the alternative, I still believe that the ability to configure `certificateValidityPeriod` could still be very beneficial.

kron4eg:
Now users could specify the time certificates generated by the kubeadm are valid.

```yaml
kind: KubeOneCluster

versions:
  kubernetes: "1.2.3"


certificateAuthority:
  certificateValidityPeriod: 87600h # 10 years for kubeapi serving certificate
  caCertificateValidityPeriod: 876000h # 100 years for PKI CA certificate
```

See https://pkg.go.dev/time#ParseDuration for format rules.

Fixes #3582

**What type of PR is this?**
/kind feature

```release-note
Add certificateValidityPeriod and caCertificateValidityPeriod API
```

**Documentation**:

```documentation
API docs
```


ahmedwaleedmalik (approved):
/approve

PR #3823, opened 2025-09-16
Add support for k8s version 1.34

### LINKED ISSUE #3785: Support Kubernetes 1.34
This is a collector issue for Kubernetes 1.34 support in KubeOne. The following tasks should be taken care of:

* [ ] Check the Kubernetes changelog to ensure there are no breaking changes and removals
* [ ] Update the `build` image to update Sonobuoy (and other dependencies if needed) 
* [ ] Update the latest supported Kubernetes version in [the API validation](https://github.com/kubermatic/kubeone/blob/main/pkg/apis/kubeone/validation/validation.go#L40-L45) 
* [ ] Update [default admission controllers](https://github.com/kubermatic/kubeone/blob/main/pkg/kubeflags/data.go) if needed 
* [ ] Update `pause` image version in `pkg/apis/kubeone/helpers.go` if needed 
* [ ] Add the appropriate `cri-tools` version in `pkg/scripts/os_flatcar.go` 
* [ ] Update [the stable version marker in Makefile](https://github.com/kubermatic/kubeone/blob/5273f9a372736569c6b09b38f2959019d29e4d6a/Makefile#L24) 
* [ ] Add E2E tests inside `tests.yml` 
* [ ] Update daily periodics to use the latest Kubernetes release
* [ ] Update [the Compatibility Matrix](https://docs.kubermatic.com/kubeone/main/architecture/compatibility/supported-versions/) 
* [ ] Create an issue to track updating [images](https://github.com/kubermatic/kubeone/blob/main/pkg/templates/images/images.go) 

Update default admission controllers if needed:
To find out what admission controllers are enabled by default, you can run
kube-apiserver --help and search for the --enable-admission-plugins flag.
The easiest way to run kube-apiserver is using Docker such as:
`docker run --rm registry.k8s.io/kube-apiserver:v1.3x.0 kube-apiserver -h`.

Update `pause` image version in `pkg/apis/kubeone` if needed:
The pause image version used by kubeadm can be found here
(search for PauseVersion):
https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/constants/constants.go.

archups:
This PR  is to add support for k8s version 1.34. 
Also include Go build image bump to 1.25.1 and Go dependencies version bump for controller-runtime and k8s.io packages.

Fixes #3785

**What type of PR is this?**
/kind feature

```release-note
Add support for k8s version 1.34
```

**Documentation**:

```documentation
https://github.com/kubermatic/docs/pull/1981
```


pkg/templates/kubeadm/kubeadm.go
```diff
@@ -49,14 +44,5 @@ type Kubedm interface {
...
-	sver, err := semver.NewVersion(ver)
-	if err != nil {
-		return nil, err
-	}
-
-	if preV131Constraint.Check(sver) {
-		return &kubeadmv1beta3{version: ver}, nil
-	}
-
 	return &kubeadmv1beta4{version: ver}, nil
```
kron4eg, line 61: please don't remove this

archups: okay

pkg/kubeflags/data.go
```diff
@@ -33,6 +33,9 @@ var defaultAdmissionControllers = []string{
 	"ClusterTrustBundleAttest",
 	"CertificateSubjectRestriction",
 	"DefaultIngressClass",
+	// TODO: Support for `PodTopologyLabels` is added since 1.33. Uncomment once the k8s 1.32 support is drop
+	// "PodTopologyLabels",
```
kron4eg: We used to have different lists for different versions, but we stopped since the lists were identical, looks like we need to start doing this again.

pkg/templates/kubeadm/v1beta4/kubeadm.go
```diff
@@ -245,10 +245,6 @@ func etcdVersionCorruptCheckExtraArgs(cipherSuites []string) []kubeadmv1beta4.Ar
 			Name:  "experimental-compact-hash-check-enabled",
 			Value: "true",
 		},
-		{
-			Name:  "experimental-initial-corrupt-check",
-			Value: "true",
-		},
 		{
```
kron4eg, line 254: what happend to the flag? it's still available in the etcd.

archups: For k8s 1.34 which includes etcd 3.6.x, we faced e2e prow job `pull-kubeone-e2e-aws-default-install-containerd-external-v1.34` failure due to etcd pod exited with following logs: 
`{"level":"warn","ts":"2025-09-24T21:17:53.052517Z","caller":"etcdmain/etcd.go:66","msg":"failed to verify flags","error":"cannot specify both flags: --experimental-initial-corrupt-check=true and --feature-gates=InitialCorruptCheck=true at the same time, please just use --feature-gates=InitialCorruptCheck=true"}`

Hence it is removed to fix the issue. However, if there is any further thoughts, let me know. Thanks!

pkg/kubeflags/data.go
archups: Do we want to start maintaining separate lists for different versions from now? IMO, if there is no hard requirement for adding  `PodTopologyLabels` which was actually introduced in k8s 1.33 then we can wait for next k8s minor release be out and see if it introduce new flags which are not supported by previous. Also by then 1.32 should reach EOL.

archups: @kron4eg  PTAL, I have pushed change to maintain different lists for different versions, like we had in the past. Thanks!

PR #3840, opened 2025-09-30
Flip apt and repository operations

ttuellmann:

The installer currently tries to install `curl` and `gnupg` after setting up the k8s repository. Preparing the signing key requires both `curl` and `gnupg`. 

This PR flips the order by installing all OS packages first, then downloading the pkgs.k8s.io key and setting up hte sources.list file. Once that is complete `apt-get update` is called again to make `apt` aware of the new repo.

**What type of PR is this?**

/kind cleanup

```release-note
Improve OS package handling on deb systems
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/scripts/os_debian.go
+++ b/pkg/scripts/os_debian.go
@@ -43,19 +43,6 @@ Acquire::http::Proxy "{{ .HTTP_PROXY }}";
 {{- end }}
 EOF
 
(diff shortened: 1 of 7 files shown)


kron4eg (approved):
looks like we got lucky our images always already had gpg and curl

/approve

ttuellmann:
> looks like we got lucky our images always already had gpg and curl

Most normal installations usually have it but we are running on _very_ minimal Debian images. 

Thanks for merging this!

PR #3849, opened 2025-10-28
Fix upgrading OCI releases and uninstalling them without reason

mohamed-rafraf:

This PR avoids upgrading OCI helm releases and queuing them to be uninstalled without any reason! 
* We upgrade if the chart or values are changed! 
* We uninstall the release if it's not defined in the kubeone configuration 

Fixes #

**What type of PR is this?**

/kind bug

```release-note
Fix upgrading OCI helm releases and uninstalling them without reason
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/localhelm/helm3.go
+++ b/pkg/localhelm/helm3.go
@@ -249,24 +249,17 @@ func cleanManifest(manifest string) string {
 		return ""
 	}
 
-	var buf bytes.Buffer
-	lines := bytes.Split([]byte(manifest), []byte("\n"))
+	var buf strings.Builder
 
-	for _, line := range lines {
-		if bytes.HasPrefix(line, []byte("#")) {
+	for line := range strings.Lines(manifest) {
+		if strings.HasPrefix(line, "#") {
 			continue
 		}
-		buf.Write(line)
+		buf.WriteString(line)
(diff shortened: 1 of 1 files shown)


pkg/localhelm/helm3.go
```diff
@@ -231,6 +232,43 @@ func newRestClientGetter(kubeConfigFileName, namespace string, st *state.State)
...
+	}
+
+	// Remove the trailing newline added after the last line
+	result := buf.Bytes()
+	if len(result) > 0 && result[len(result)-1] == '\n' {
+		result = result[:len(result)-1]
+	}
+
+	return string(result)
+}
```
kron4eg, line 263:
```suggestion
func cleanManifest(manifest string) string {
	if manifest == "" {
		return ""
	}

	var buf strings.Builder

	for line := range strings.Lines(manifest) {
		if strings.HasPrefix(line, "#") {
			continue
		}
		buf.WriteString(line)
		buf.WriteByte('\n')
	}

	return strings.TrimSpace(buf.String())
}
```

kron4eg (changes requested):
Some small readability improvements suggested

kron4eg (approved):
/approve

PR #3884, opened 2025-11-09
Fix invalid GCE Persistent Disk CSI Driver image reference from v1.17.3 to v1.17.4 and update associated images.

rajaSahil:
- This PR fixes the invalid reference image for the GCE Persistent Disk CSI Driver by updating it to the correct version `v1.17.4` from registry.k8s.io. The previous image `v1.17.3` was not available in the registry, causing CSI driver pods to fail during cluster creation on GCP.
- Update GCP Compute Persistent Disk CSI image and resources

Fixes #

**What type of PR is this?**

/kind bug

```release-note
Fixed an invalid image reference for the GCE Persistent Disk CSI Driver and update associated images.
```

**Documentation**:

```documentation
NONE
```


kron4eg (approved):
/approve

PR #3888, opened 2025-11-12
Hide helm logs

mohamed-rafraf:

This PR hides the Helm logs from being printed while running kubeone apply (You will see Helm logs when you use OCI Helm charts )

Fixes #

**What type of PR is this?**

/kind chore

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


mohamed-rafraf:
> But how do we know if something happens?

Are they using that writer for Errors, too?? :astonished:

kron4eg:
At least let's choose io.Discard when  there's no verbose flag given

PR #3895, opened 2025-11-17
beta: managed control plane on hetzner

### LINKED ISSUE #3784: Manage Control Plane VMs on Hetzner
Implement https://github.com/kubermatic/kubeone/issues/3783 on Hetzner provider.

kron4eg:
THIS IS BETA, DO NOT USE FOR PRODUCTION! 

At the moment KubeOne requires users to pre-provision control-plane VMs/load-balancer and supply their SSH addresses via a static `controlPlane.hosts` list. This PR implements the foundation for full lifecycle management of control-plane nodes on Hetzner Cloud — KubeOne provisions (or discovers) the VMs and load balancer automatically using machine-controller's cloud-provider drivers.

Users declare a `controlPlane.nodeSets` spec instead of (or in addition to) a static host list. KubeOne then:
* Creates (or finds) a Hetzner Load Balancer fronting port 6443
* Creates (or finds) the control-plane VMs
* Injects the resolved IPs as HostConfig entries before continuing with the normal cluster lifecycle management.

Fixes #3784

**What type of PR is this?**
/kind feature

New API surface

```
cloudProvider:
  hetzner:
    networkID: my-private-network   # required when controlPlane is set
    controlPlane:
      loadBalancer:
        name: my-cluster-kubeapi   # default: "<cluster>-kubeapi"
        type: lb11                  # default
        location: nbg1              # default
        publicIP: true              # default

controlPlane:
  nodeSets:
    - name: cp
      replicas: 3
      operatingSystem: ubuntu
      ssh:
        publicKeys: ["ssh-ed25519 AAAA..."]
        username: ubuntu
      cloudProviderSpec:
        serverType: cx22
        location: nbg1
        image: ubuntu-24.04
        networks:
          - my-private-network
```

```release-note
THIS IS BETA, DO NOT USE FOR PRODUCTION!

KubeOne can now provision and manage Hetzner control-plane VMs and load balancer automatically via `controlPlane.nodeSets`, removing the need for pre-provisioned hosts.
```

**Documentation**:

```documentation
TBD
```

### DIFF
--- a/.prow/verify.yaml
+++ b/.prow/verify.yaml
@@ -90,10 +90,10 @@ presubmits:
             - build
           resources:
             requests:
-              memory: 1Gi
+              memory: 2Gi
               cpu: 1
             limits:
-              memory: 2Gi
+              memory: 3Gi
               cpu: 2
   - name: pull-kubeone-test
     always_run: true
@@ -110,10 +110,10 @@ presubmits:
             - test
           resources:
             requests:
-              memory: 1Gi
+              memory: 2Gi
               cpu: 1
             limits:
-              memory: 2Gi
+              memory: 3Gi
               cpu: 2
   - name: pull-kubeone-build-image
     always_run: true


pkg/provisioner/provisioner.go
```diff
@@ -0,0 +1,214 @@
...
+	"fmt"
+	"regexp"
+	"strings"
+	"text/template"
+	"time"
+
+	"github.com/sirupsen/logrus"
+
+	"k8c.io/kubeone/pkg/fail"
+	"k8c.io/machine-controller/pkg/cloudprovider"
```
adoi, line 31: I think the `pull-kubeone-build` and `pull-kubeone-test` failures/hangs are related to this. `k8c.io/machine-controller/pkg/cloudprovider` is the global provider registry, and its `ForProvider` imports all provider backends, not just Hetzner. For a Hetzner beta feature, that pulls quite a large graph into the main CLI. Can you try importing the Hetzner provider directly instead of the global registry?

pkg/provisioner/provisioner.go
```diff
@@ -0,0 +1,214 @@
...
+			return nil, fail.MachineController(fmt.Errorf("machine %s has not been assigned an IP yet", providerInstance.Name()), "getting instance addresses from provider")
+		}
+
+		instances = append(instances, providerInstance)
+	}
+
+	return getMachineProvisionerOutput(instances), nil
+}
+
+func getProvider(ctx context.Context, machine clusterv1alpha1.Machine) (cloudprovidertypes.Provider, error) {
```
adoi, line 201: This generic provider lookup is also quite expensive, some unrelated SDKs like the AWS one are being dragged into build

pkg/provisioner/provisioner.go
kron4eg: I gave them a bit more room of memory and now they are passing

kron4eg: Unfortunatelly I have to import cloudprovider, as its structures are a part upstream APIs that I have to use in order to get the Machine created. Besides, the generic nature here is to accomodate for the next provoders to come.

adoi: My idea was to import `k8c.io/machine-controller/pkg/cloudprovider/provider/hetzner` directly. That would still use the upstream APIs needed for this feature, and it would avoid pulling unrelated providers into the dep graph until we support them. If increasing CI resources is acceptable it should be fine.
`pull-kubeone-build-image` now takes 40-50mins on average on every PR once it would be merged.

PR #3904, opened 2025-11-21
Use InternalImages for backup restic and unattended upgrades addons

archups:
This PR is to start using InternalImages for backup restic and unattended upgrades addons. This is inline to how the addons images are managed and reference using InternalImages. Currently It is not possible to use overwriteRegistry for Unattended upgrade addons Yum and Apt, hence this change will allow to use overwriteRegistry for them as well. 
Bump alpine image version to 3.22.2

Fixes #

**What type of PR is this?**
/kind chore

```release-note
Bump alpine image version to 3.22.2
```

**Documentation**:

```documentation
NONE
```


kron4eg:
@archups it's been in the release since the beginning

archups:
> @archups it's been in the release since the beginning

@kron4eg yes, I got it. Thanks! :)

PR #3929, opened 2025-12-02
Mirror Images: mirror coredns for containerd mirrors and overwriteReg…

mohamed-rafraf:

KubeOne supports private registries in two ways: by using `registryConfiguration.overwriteRegistry`, where kubeadm rewrites the CoreDNS image to `<registry>/coredns`, and by using containerd registry mirrors, where kubeadm continues using the upstream path and containerd expects CoreDNS at `<registry>/coredns/coredns`. Because both modes require different image paths, CoreDNS must be mirrored under **both** names. The current implementation only mirrors to `<registry>/coredns`, which breaks the mirror-based setup, so this change ensures CoreDNS is mirrored to both `<registry>/coredns` and `<registry>/coredns/coredns`.

Fixes #

**What type of PR is this?**

/kind chore

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


kron4eg (approved):
/approve

Issue #3726, opened 2025-07-01
kubeone installation issues in a hardened environment


dharapvj:
### What happened?

When attempted to use kubeone to install kuberenetes in a security hardened environment, i encountered following issues:

1. /etc/kubeone/proxy-env - Permission denied
2. ssh - administratively provibited.

### Expected behavior

kubeone installation should proceed without any issues and succeed.

####  Issue 1
1. set umask to 0027 in the control plane nodes.
2. set some proxy values in kubeone.yaml
3. run kubeone apply. kubeone uses sudo to create the proxy-env file which gets created fine but the regular login user does not have access to it (because of umask `others` do not have access to /etc/kubeone directory and the file within)

####  Issue 2
1. On the control plane machines, edit /etc/ssh/sshd_config to add `DisableForwarding yes`
2. Restart `sshd service by sudo systemctl restart sshd.service`
3. run kubeone apply. 
4. You will not be able to reach control plane machine during the csr-approval stage with error `administratively prohibited`

### What KubeOne version are you using?

Kubeone v1.10.0

### What cloud provider are you running on?

vsphere

### What operating system are you running in your cluster?

RHEL

dharapvj:
For issue 1 - I think this is the line where we will face issues since the regular ssh user will not have read rights to `proxy-env` file.

https://github.com/kubermatic/kubeone/blob/d148423e3475070a5762e3a055830d69c842f47e/pkg/scripts/proxy.go#L66

Possible fix is to add below line after creation of proxy-env file..

```
# file: pkg/scripts/proxy.go

sudo chmod o=rx -R /etc/kubeone
```

kron4eg:
`DisableForwarding no` is required, as we route all the kube-api requests via the ssh tunnels.

dharapvj:
Can we add this requirement [here](https://docs.kubermatic.com/kubeone/v1.10/guides/ssh/#sshd-requirements-on-instances) then?

Also - I think we should have mentioned these SSHD kubeone requirements in [requirements section](https://docs.kubermatic.com/kubeone/v1.10/architecture/requirements/)

dharapvj:
[PR for updating the docs](https://github.com/kubermatic/docs/pull/1914)

I see that the sshd requirements link is already present in the requirements section.. in last bullet. I had missed it earlier. So no change needed in requirements section

xmudrii:
The docs were updated, I think we can close this issue. If we missed something, please reopen the issue.
/close

PR #3939, opened 2025-12-15
fix: set /etc/kubeone/proxy-env file permissions explicitly 

piotr1212:
On systems with a stricter umask (eg. 0027) the created proxy-env file cannot be sourced later on resulting in the following error:
```
 + source /etc/kubeone/proxy-env
 bash: line 8: /etc/kubeone/proxy-env: Permission denied
```

This sets the permissions explicitly to prevent this. This issue is also described in:
https://github.com/kubermatic/kubeone/issues/3726

Fixes #

**What type of PR is this?**

```release-note
fix: set /etc/kubeone/proxy-env file permissions explicitly 
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/scripts/testdata/TestEnvironmentFile-empty-proxy.golden
+++ b/pkg/scripts/testdata/TestEnvironmentFile-empty-proxy.golden
@@ -1,10 +1,12 @@
 set -xeuo pipefail
 export "PATH=$PATH:/sbin:/usr/local/bin:/opt/bin"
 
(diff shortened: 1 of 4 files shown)


kron4eg:
@piotr1212 please also run `go test ./pkg/scripts/... -update`, and commit the testdata changes too

kron4eg:
@piotr1212 also, please return back release-note block that you've removed from the PR description template.

PR #3971, opened 2026-01-21
feat: OSM resources

### LINKED ISSUE #3965: resources OperatingSystemManager

The OperatingSystemManager's resources, are not sufficient when running 60-ish nodes. It will be OOMKilled. When that is the case, new nodes are not provisioned as the OperatingSystemManager is down.

### Solution details

Please raise the resources to for example:

```
          resources:
            limits:
              cpu: "2"
              memory: 2Gi
            requests:
              cpu: "1"
              memory: 1Gi
```

### Alternative approaches

Use a HPA / VPA of the OperatoringSystemManager does allow such thing. Only useful when there is no leaderElection. This will create a new dependency, but a HPA should be available on vanilla K8s. 

### Use cases

Regular usage of the OperatingSystemManager.

appiepollo14:
The default resources which are given to OSM addon are insufficient with a large amount (>60) of nodes. Please raise them.

Fixes #3965 

**What type of PR is this?**

/kind feature
/kind chore

A HPA would be more appropriate IMHO, but I can't determine wether that is possible. I don't see leader election in OSM, however, as it is a Kubernetes Controller, I would expect the concept to be based on just one manager for the CRD's. Let me know please.

```release-note
Add `OSM_MEMORY_LIMIT` parameter in operating-system-manager addon
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/addons/operating-system-manager/deployment-controller.yaml
+++ b/addons/operating-system-manager/deployment-controller.yaml
@@ -98,7 +98,7 @@ spec:
           resources:
             limits:
               cpu: "2"
-              memory: 2Gi
+              memory: {{ default "512Mi" .Params.OSM_MEMORY_LIMIT }}
             requests:
               cpu: 50m
               memory: 128Mi


addons/operating-system-manager/deployment-controller.yaml
```diff
@@ -98,7 +98,7 @@ spec:
           resources:
             limits:
               cpu: "2"
-              memory: 512Mi
+              memory: 2Gi
```
kron4eg: let's leave it 512Mi by default but make it configurable via the [.Params](https://docs.kubermatic.com/kubeone/v1.12/guides/addons/#parameters)

appiepollo14: That is possible but seems unnecessarily complex for this purpose. I will provide a PR to show this.

kron4eg: I mean, the code could look like this:

```yaml
            limits:
              cpu: "2"
              memory: {{ default "512Mi" .Params.OSM_MEMORY_LIMIT }}
```

doesn't look too much of a trouble for me

appiepollo14: > I mean, the code could look like this:
> 
> ```yaml
>             limits:
>               cpu: "2"
>               memory: {{ default "512Mi" .Params.OSM_MEMORY_LIMIT }}
> ```
> 
> doesn't look too much of a trouble for me

Agreed that this is not much of trouble. The choice to do this is more of a product manager's choice. Meaning, how many of there params would one support until it becomes a HELMChart on it's own.

kron4eg: You can always carry own overriding copy of an addon :man_shrugging:

appiepollo14: > You can always carry own overriding copy of an addon 🤷‍♂️

Agreed, but that defeats the purpose of using KubeOne. That said, I will update the change, it was just a remark that I believe it is a risk of putting too much options in the addons. Therefore I hoped the HPA was an option, as that makes is automatically suitable for each, whichout any config.

appiepollo14:
@kron4eg can you review once more ?

appiepollo14:
@kron4eg how can we get this pr moving?

appiepollo14:
@kron4eg One final question before merging. For example this addon:
'addons/machinecontroller/deployment-controller.yaml' doesn't specify any resources. Wouldn't that be the most simple solution for OSM as well?

appiepollo14:
@kron4eg Which solution should we choose?

kron4eg:
@appiepollo14 you've brought up an interesting topic. I actually completely forgot about MC resources since that manifest virtually never changes... 

* If we have no limits in the MC should we have limits for the OSM?
* Or otherwise, should we introduce configurable limits to the MC?
* Maybe we could simply get away with mere "requests" instead of "limits"?

@appiepollo14 in your observations, what are the resource usage patterns in terms Machines/Nodes count VS memory?

kron4eg:
I'm inclining towards removing limits at all, and have only some "requests". In case if limits must be enforced users could deploy something like kyverno and have control over those fields in the admission phase.

kron4eg:
> I'm inclining towards removing limits at all, and have only some "requests". In case if limits must be enforced users could deploy something like kyverno and have control over those fields in the admission phase.

like this https://github.com/kubermatic/kubeone/pull/3979

appiepollo14:
> I'm inclining towards removing limits at all, and have only some "requests". In case if limits must be enforced users could deploy something like kyverno and have control over those fields in the admission phase.

Hence my discussion about introducing a new parameter in the first place. That would make things more complex when introducing more variables to tune. Removing the limits as mentioned in #3979 is in my opinion the way to proceed. The less complex the better. 

To answer your question about mem usage vs nodes, I can't share that specific details. What I can state is that those controllers, like most of them do, spike in mem usage at start-up, as they then reconcile each object once. As stated earlier, 60 nodes make OSM spike around the 600Mi. One could implement that client-go ratelimiter to be more in control.  

Let's conclude this by closing this PR in favor of #3979 !

PR #3993, opened 2026-02-25
Remove mutual exclusivity between ContainerdRegistry and RegistryConfiguration

### LINKED ISSUE #3988: Support using the RegistryConfiguration API at the same time as the ContainerRegistry API
### Description of the feature you would like to add / Use case

For some reason, KubeOne will refuse using the two mentioned APIs at the same time. I fail to see a technical limitation to combining them. I can kind of understand that combining `overwriteRegistry` with `mirrors`  would be a niche case (although I think I can come up with a valid use case for it). But the ContainerRegistry API can't just be used to configure registry mirrors, it can also be used to configure registry authentication. And that's really something completely different to what the RegistryConfiguration API does.

So in essence, we would like to combine them as follows:
- Use the RegistryConfiguration API to instruct KubeOne to use our mirror for the KubeOne images
- Use the ContainerRegistry API to instruct KubeOne to use credentials for (another) registry - without specifying a mirror. It's simply the easiest way to configure credentials for our private images, rather than adding the credentials to all the pod specs (and many Helm charts still don't support providing credentials anyway, and overwriting it with kustomize isn't fun either)

Maybe to add a rationale for one question in your mind right now: using mirrors isn't an option for us. The problem with configuring registry mirrors is, that it's a transparent configuration (and containerd simply falls back to the original registry if the mirror doesn't have an image iirc - it's a HA solution, while we look at both HA and security). However, we want to configure Kyverno to only allow one registry in future. That's where `overwriteRegistry` is the vastly superior, and preferred option.

### Solution details

- Allow (i.e. stop prohibiting) using the RegistryConfiguration API at the same time as the ContainerRegistry API
- If there's a valid technical reason to prohibit using `overwriteRegistry` and `mirrors` at the same time, prohibit just that

### Alternative approaches

Bribing our information security officer and the ISO certification auditor ... just kidding, please don't. Or do, but make sure nobody learns about it :)

kron4eg:

Fixes #3988

**What type of PR is this?**
/kind feature

```release-note
Remove mutual exclusivity between ContainerdRegistry and RegistryConfiguration
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/apis/kubeone/validation/validation.go
+++ b/pkg/apis/kubeone/validation/validation.go
@@ -82,35 +82,6 @@ func ValidateKubeOneCluster(c kubeoneapi.KubeOneCluster) field.ErrorList {
 	allErrs = append(allErrs, ValidateAddons(c.Addons, field.NewPath("addons"))...)
 	allErrs = append(allErrs, ValidateRegistryConfiguration(c.RegistryConfiguration, field.NewPath("registryConfiguration"))...)
 	allErrs = append(allErrs, ValidateControlPlaneComponents(c.ControlPlaneComponents, field.NewPath("controlPlaneComponents"))...)
-	allErrs = append(allErrs,
-		ValidateContainerRuntimeVSRegistryConfiguration(
-			c.ContainerRuntime,
-			field.NewPath("containerRuntime"),
-			c.RegistryConfiguration,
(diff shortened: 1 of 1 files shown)


judge-red:
Awesome, thanks Artiom!

PR #3998, opened 2026-03-03
feat(etcd): Add etcd configuration options and operational commands

### LINKED ISSUE #3898: Add support for passing etcd arguments in KubeOne

We need the ability to configure and pass additional arguments(like "auto-compaction-retention" and "auto-compaction-mode") to the etcd instances managed by KubeOne. 

### Solution details

- I think we can pass the additional parameters to this field - https://github.com/kubermatic/kubeone/blob/6a1eb946c7586d6b1b7838970e50b385d039f97c/pkg/apis/kubeadm/v1beta4/types.go#L325

### Use cases

There is a support ticket that is a blocker for customers and they need additional arguments to be able to fix this issue: https://github.com/etcd-io/etcd/issues/11947

### LINKED ISSUE #3952: Support for automated ETCD Defragmentation

Currently, when a KubeOne-managed cluster's etcd database reached its storage limit (default 2GiB https://etcd.io/docs/v3.6/dev-guide/limit/), it caused the cluster to become unresponsive, also leading to control plane unavailability with SSH connectivity lost on the cluster control plane nodes due to the resource crisis. Resolution required manual intervention to perform etcdctl defrag across cluster members to reclaim fragmented space.

Hence, I  would like to request an automated etcd defragmentation solution within KubeOne that mirrors the implementation used for Kubermatic Kubernetes Platform (KKP) user clusters.

### Solution details

- Automated Defragmentation: Implement a mechanism to monitor etcd database size and automatically trigger a defragmentation cycle when specific thresholds are reached or when a NOSPACE alarm is detected.
- Storage Quota Customization: Allow users to customize the etcd storage limit via the KubeOneCluster manifest. This should map to the `--quota-backend-bytes` flag in etcd.
- Operational Safety: The defragmentation should be performed in a rolling fashion to maintain cluster availability.
It should also include an automatic "disalarm" (clearing the NOSPACE alarm if raised) after successful defragmentation and verification that all members are below the threshold.

To ensure consistency across the Kubermatic ecosystem, this feature should be implemented in line with the [KKP user cluster etcd maintenance](https://docs.kubermatic.com/kubermatic/v2.29/tutorials-howtos/etcd-backups/) approach, potentially leveraging the EtcdLauncher logic to handle maintenance tasks

### LINKED ISSUE #3953: Set etcd env vars on the etcd pods

To work with etcdctl in the etcd pods, I have to do this:
```
kubectl -n kube-system exec -it etcd-redacted-0 -- etcdctl --endpoints 127.0.0.1:2379   --cert=/etc/kubernetes/pki/etcd/server.crt   --key=/etc/kubernetes/pki/etcd/server.key   --cacert=/etc/kubernetes/pki/etcd/ca.crt   endpoint status
```

... or ideally I specify all three endpoints with their individual IPs.

If I want to do the same on a KKP UC, it's much easier:
```
k -n cluster-redacted exec -it etcd-0 -- etcdctl endpoint status
```

I never remember these damn parameters and it's annoying (particularly since I don't even have tab completion on those paths), please make it as easy as in KKP.

### Solution details

KKP sets these env vars in the pod's container spec:
```
    - name: ETCDCTL_API
      value: "3"
    - name: ETCDCTL_CACERT
      value: /etc/etcd/pki/ca/ca.crt
    - name: ETCDCTL_CERT
      value: /etc/etcd/pki/client/apiserver-etcd-client.crt
    - name: ETCDCTL_KEY
      value: /etc/etcd/pki/client/apiserver-etcd-client.key
    - name: ETCDCTL_ENDPOINTS
      value: https://etcd-0.etcd.cluster-redacted.svc.cluster.local:2379,https://etcd-1.etcd.cluster-redacted.svc.cluster.local:2379,https://etcd-2.etcd.cluster-redacted.svc.cluster.local:2379
```

### Alternative approaches

All others are worse workarounds imho

kron4eg:
This PR introduces a bunch of life improvements for kubeone operator person:

### New kubeone etcd command group
A new etcd top-level command is added with three subcommands, all of which connect to the etcd cluster via the existing SSH tunnel and TLS setup:

`kubeone etcd members` — Lists all etcd cluster members along with their peer/client URLs, learner status, and any active alarms. Supports --output table (default) and --output json.
`kubeone etcd disarm [member-name]` — Disarms active alarms on a specific etcd member (by name) or on all members with --all. The two options are mutually exclusive.
`kubeone etcd defragment <member-name>` — Defragment releases wasted space from internal fragmentation on a given etcd member.
`kubeone etcd snapshot <file>` — Saves a point-in-time snapshot of the etcd cluster data to a local file. Use - as the path to stream to stdout.

### New EtcdConfig API field
A new optional etcd field is added under controlPlaneComponents to expose common etcd tuning parameters:

```
kind: KubeOneCluster
# ... omitted for brevity
controlPlaneComponents:
  etcd:
    # this optional value will translate into `--quota-backend-bytes` flag of etcd
    quotaBackendBytes: 8589934592        # 8 GiB

    # this optional value will translate into `--auto-compaction-retention` flag of etcd
    autoCompactionRetention: 1h

    # this optional value will translate into `--auto-compaction-mode` flag of etcd
    autoCompactionMode: periodic          # periodic | revision
```

### etcd pod ExtraEnvs via kubeadm

The kubeadm v1beta4 configuration now is injected with standard etcdctl environment variables (`ETCDCTL_API`, `ETCDCTL_CACERT`, `ETCDCTL_CERT`, `ETCDCTL_KEY`, `ETCDCTL_ENDPOINTS`) into the etcd pod, making `kubectl exec` into the etcd pod immediately usable without manual flag passing.

Fixes #3952
Fixes #3953
Fixes #3898

**What type of PR is this?**
/kind feature

```release-note
feat: Add etcd configuration options and operational commands
```

**Documentation**:

```documentation
TBD
```

### DIFF
--- a/pkg/apis/kubeone/validation/validation.go
+++ b/pkg/apis/kubeone/validation/validation.go
@@ -864,27 +864,35 @@ func ValidateControlPlaneComponents(c *kubeoneapi.ControlPlaneComponents, fldPat
 		}
 	}
 
-	if c.Etcd != nil {
-		etcdFldPath := fldPath.Child("etcd")
-		if c.Etcd.QuotaBackendBytes != 0 && c.Etcd.QuotaBackendBytes < 2*1024*1024*1024 {
-			allErrs = append(allErrs, field.Invalid(etcdFldPath.Child("quotaBackendBytes"), c.Etcd.QuotaBackendBytes, "quotaBackendBytes must be greater than 2GB"))
-		}
+	allErrs = ValidateEtcdConfig(c.Etcd, fldPath, allErrs)
 
-		switch c.Etcd.AutoCompactionMode {
-		case kubeoneapi.EtcdAutoCompactionModePeriodic, kubeoneapi.EtcdAutoCompactionModeRevision:
-			if c.Etcd.AutoCompactionRetention != "" {
-				_, err := time.ParseDuration(c.Etcd.AutoCompactionRetention) // validate if it's time.Duration format
-				if err != nil {
-					_, err = strconv.Atoi(c.Etcd.AutoCompactionRetention) // validate if it's simply a number
-				}
-				if err != nil {
-					allErrs = append(allErrs, field.Invalid(etcdFldPath.Child("autoCompactionRetention"), c.Etcd.AutoCompactionRetention, fmt.Sprintf("invalid duration format: %v", err)))
-				}
+	return allErrs
+}
+
+func ValidateEtcdConfig(etcdConf *kubeoneapi.EtcdConfig, fldPath *field.Path, allErrs field.ErrorList) field.ErrorList {
+	if etcdConf == nil {
+		return allErrs
+	}
+
+	etcdFldPath := fldPath.Child("etcd")
+	if etcdConf.QuotaBackendBytes != 0 && etcdConf.QuotaBackendBytes < 2*1024*1024*1024 {
+		allErrs = append(allErrs, field.Invalid(etcdFldPath.Child("quotaBackendBytes"), etcdConf.QuotaBackendBytes, "quotaBackendBytes must be greater than 2GB"))
+	}
+
+	switch etcdConf.AutoCompactionMode {
+	case kubeoneapi.EtcdAutoCompactionModePeriodic, kubeoneapi.EtcdAutoCompactionModeRevision:
+		if etcdConf.AutoCompactionRetention != "" {
+			_, err := time.ParseDuration(etcdConf.AutoCompactionRetention) // validate if it's time.Duration format
+			if err != nil {
+				_, err = strconv.Atoi(etcdConf.AutoCompactionRetention) // validate if it's simply a number
 			}
-		case "":
-		default:
-			allErrs = append(allErrs, field.Invalid(etcdFldPath.Child("autoCompactionMode"), c.Etcd.AutoCompactionMode, "invalid autoCompactionMode"))
+			if err != nil {
+				allErrs = append(allErrs, field.Invalid(etcdFldPath.Child("autoCompactionRetention"), etcdConf.AutoCompactionRetention, fmt.Sprintf("invalid duration format: %v", err)))
+			}
 		}
+	case "":
+	default:
+		allErrs = append(allErrs, field.Invalid(etcdFldPath.Child("autoCompactionMode"), etcdConf.AutoCompactionMode, "invalid autoCompactionMode"))
 	}
 
 	return allErrs
--- a/pkg/apis/kubeone/validation/validation_test.go
+++ b/pkg/apis/kubeone/validation/validation_test.go
@@ -930,6 +930,200 @@ func TestValidateVersionConfig(t *testing.T) {
 	}
 }
 
+func TestValidateEtcdConfig(t *testing.T) {
+	twoGB := int64(2 * 1024 * 1024 * 1024)
+
+	tests := []struct {
+		name          string
+		etcdConf      *kubeoneapi.EtcdConfig
+		existingErrs  field.ErrorList
+		expectedError bool
+	}{
+		{
+			name:          "nil etcd config returns existing errors unchanged (no error)",
+			etcdConf:      nil,
+			existingErrs:  field.ErrorList{},
+			expectedError: false,
+		},
+		{
+			name:          "nil etcd config preserves pre-existing errors",
+			etcdConf:      nil,
+			existingErrs:  field.ErrorList{field.Invalid(field.NewPath("test"), "val", "pre-existing error")},
+			expectedError: true,
+		},
+		{
+			name:          "empty etcd config (zero values) is valid",
+			etcdConf:      &kubeoneapi.EtcdConfig{},
+			existingErrs:  field.ErrorList{},
+			expectedError: false,
+		},
+		{
+			name: "quotaBackendBytes = 0 is valid (disabled)",
+			etcdConf: &kubeoneapi.EtcdConfig{
+				QuotaBackendBytes: 0,
+			},
+			existingErrs:  field.ErrorList{},
+			expectedError: false,
+		},
+		{
+			name: "quotaBackendBytes exactly 2GB is invalid (must be greater than 2GB)",
+			etcdConf: &kubeoneapi.EtcdConfig{
+				QuotaBackendBytes: twoGB,
+			},
+			existingErrs:  field.ErrorList{},
+			expectedError: false,
+		},
+		{
+			name: "quotaBackendBytes less than 2GB (1GB) is invalid",
+			etcdConf: &kubeoneapi.EtcdConfig{
+				QuotaBackendBytes: 1 * 1024 * 1024 * 1024,
+			},
+			existingErrs:  field.ErrorList{},
+			expectedError: true,
+		},
+		{
+			name: "quotaBackendBytes greater than 2GB is valid",
+			etcdConf: &kubeoneapi.EtcdConfig{
+				QuotaBackendBytes: twoGB + 1,
+			},
+			existingErrs:  field.ErrorList{},
+			expectedError: false,
+		},
+		{
+			name: "quotaBackendBytes of 8GB is valid",
+			etcdConf: &kubeoneapi.EtcdConfig{
+				QuotaBackendBytes: 8 * 1024 * 1024 * 1024,
+			},
+			existingErrs:  field.ErrorList{},
+			expectedError: false,
+		},
+		{
+			name: "autoCompactionMode empty string is valid",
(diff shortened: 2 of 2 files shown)


pkg/apis/kubeone/v1beta2/types.go
```diff
@@ -128,6 +128,29 @@ type ControlPlaneComponents struct {
...
+// EtcdConfig
+type EtcdConfig struct {
+	// QuotaBackendBytes is the maximum backend size in bytes for etcd. Default is empty (etcd's default).
+	QuotaBackendBytes int `json:"quotaBackendBytes,omitempty"`
+
+	// AutoCompactionRetention is the duration for automatic compaction. Default is empty (etcd's default).
+	AutoCompactionRetention metav1.Duration `json:"autoCompactionRetention,omitempty"`
+
+	// AutoCompactionMode is the mode for automatic compaction (`periodic` or `revision`).  Default is empty (etcd's default).
+	AutoCompactionMode EtcdAutoCompactionMode `json:"autoCompactionMode,omitempty"`
```
ahmedwaleedmalik: > Default is empty (etcd's default)

We should mention the actual value of default here. That'll be really useful for our docs and customers.

pkg/apis/kubeone/v1beta3/types.go
```diff
@@ -124,6 +124,29 @@ type ControlPlaneComponents struct {
...
+// EtcdConfig
+type EtcdConfig struct {
+	// QuotaBackendBytes is the maximum backend size in bytes for etcd. Default is empty (etcd's default).
+	QuotaBackendBytes int `json:"quotaBackendBytes,omitempty"`
+
+	// AutoCompactionRetention is the duration for automatic compaction. Default is empty (etcd's default).
+	AutoCompactionRetention metav1.Duration `json:"autoCompactionRetention,omitempty"`
+
+	// AutoCompactionMode is the mode for automatic compaction (`periodic` or `revision`).  Default is empty (etcd's default).
+	AutoCompactionMode EtcdAutoCompactionMode `json:"autoCompactionMode,omitempty"`
```
ahmedwaleedmalik, line 149: Same as https://github.com/kubermatic/kubeone/pull/3998/changes#r2882742656

pkg/apis/kubeone/types.go
```diff
@@ -127,6 +127,29 @@ type ControlPlaneComponents struct {
...
+// EtcdConfig
+type EtcdConfig struct {
+	// QuotaBackendBytes is the maximum backend size in bytes for etcd. Default is empty (etcd's default).
+	QuotaBackendBytes int `json:"quotaBackendBytes,omitempty"`
+
+	// AutoCompactionRetention is the duration for automatic compaction. Default is empty (etcd's default).
+	AutoCompactionRetention metav1.Duration `json:"autoCompactionRetention,omitempty"`
+
+	// AutoCompactionMode is the mode for automatic compaction (`periodic` or `revision`).  Default is empty (etcd's default).
+	AutoCompactionMode EtcdAutoCompactionMode `json:"autoCompactionMode,omitempty"`
```
ahmedwaleedmalik, line 152: Same as https://github.com/kubermatic/kubeone/pull/3998/changes#r2882742656

pkg/apis/kubeone/v1beta2/types.go
```diff
@@ -128,6 +128,29 @@ type ControlPlaneComponents struct {
...
+
+const (
+	EtcdAutoCompactionModePeriodic EtcdAutoCompactionMode = "periodic"
+	EtcdAutoCompactionModeRevision EtcdAutoCompactionMode = "revision"
+)
+
+// EtcdConfig
+type EtcdConfig struct {
+	// QuotaBackendBytes is the maximum backend size in bytes for etcd. Default is empty (etcd's default).
+	QuotaBackendBytes int `json:"quotaBackendBytes,omitempty"`
```
ahmedwaleedmalik: This should be int64

pkg/templates/kubeadm/v1beta4/kubeadm.go
```diff
@@ -177,7 +180,39 @@ func NewConfig(s *state.State, host kubeoneapi.HostConfig) (*Config, error) {
...
+							Name:  "ETCDCTL_CERT",
+							Value: "/etc/etcd/pki/client/apiserver-etcd-client.crt",
+						},
+					},
+					{
+						EnvVar: corev1.EnvVar{
+							Name:  "ETCDCTL_KEY",
+							Value: "/etc/etcd/pki/client/apiserver-etcd-client.key",
+						},
+					},
```
ahmedwaleedmalik, line 202: The paths are wrong I believe. Should be something like:

```
ETCDCTL_CACERT=/etc/kubernetes/pki/etcd/ca.crt                                                                                                                                                                                                                                          
ETCDCTL_CERT=/etc/kubernetes/pki/apiserver-etcd-client.crt                                                                                                                                                                                                                            
ETCDCTL_KEY=/etc/kubernetes/pki/apiserver-etcd-client.key
```

pkg/apis/kubeone/v1beta2/types.go
```diff
@@ -128,6 +128,29 @@ type ControlPlaneComponents struct {
...
+
+	// AutoCompactionRetention is the duration for automatic compaction. Default is empty (etcd's default).
+	AutoCompactionRetention metav1.Duration `json:"autoCompactionRetention,omitempty"`
```
ahmedwaleedmalik: This will not work, in periodic compaction mode etcd expects a duration like `1h` but in revision mode it expects a number.

Check https://etcd.io/docs/v3.4/op-guide/maintenance/#:~:text=In%20v3.3.3%2C-,%2D%2Dauto%2Dcompaction%2Dmode%3Drevision%20%2D%2Dauto%2Dcompaction%2Dretention%3D1000,-automatically%20Compact%20on

pkg/templates/kubeadm/v1beta4/kubeadm.go
```diff
@@ -240,6 +275,36 @@ func NewConfig(s *state.State, host kubeoneapi.HostConfig) (*Config, error) {
...
+	var args []kubeadmv1beta4.Arg
+
+	if cpc != nil && cpc.Etcd != nil {
+		etcdConfig := cpc.Etcd
+		if etcdConfig.QuotaBackendBytes != 0 {
+			args = append(args, kubeadmv1beta4.Arg{
+				Name:  "quota-backend-bytes",
+				Value: fmt.Sprintf("%d", etcdConfig.QuotaBackendBytes),
+			})
+		}
```
ahmedwaleedmalik, line 282: I remember working on this topic for KKP, we should have validations to stop users from shooting themselves in the foot. 

- Reject negative value for this field
- Set a minimum value that you can't go lower than

pkg/etcdutil/etcd.go
```diff
@@ -68,7 +85,6 @@ func NewClientConfig(s *state.State, host kubeoneapi.HostConfig) (*clientv3.Conf
 		Context:     s.Context,
 		DialTimeout: 5 * time.Second,
 		DialOptions: []grpc.DialOption{
-			grpc.WithBlock(), //nolint:staticcheck
```
ahmedwaleedmalik, line 71: Nit:

Worth it? Client creation would no longer be blocked waiting for connection. Errors will pop up when we execute operations.

pkg/cmd/etcd.go
```diff
@@ -0,0 +1,415 @@
...
+			defer output.Close()
+
+			if outPath != "-" {
+				f, errF := os.Create(outPath)
+				if errF != nil {
+					return fail.Runtime(errF, "creating snapshot file %q", outPath)
+				}
+				output = f
+			}
+
```
ahmedwaleedmalik, line 403:
```suggestion
			var output io.WriteCloser = os.Stdout

			if outPath != "-" {
				f, errF := os.Create(outPath)
				if errF != nil {
					return fail.Runtime(errF, "creating snapshot file %q", outPath)
				}
				output = f
			}
			defer func() { output.Close() }()
```

The defer block wouldn't capture the file created afterwards. So the file handle never gets closed and stdout gets closed instead.

ahmedwaleedmalik (changes requested):

pkg/apis/kubeone/v1beta2/types.go
kron4eg: I wanted to avoid indicating the default value that we don't control and that could be changed without out knowledge.

pkg/etcdutil/etcd.go
kron4eg: it's deprecated and it all works without it :shrug:

Issue #3826, opened 2025-09-19
Fix CIS benchmark issues


ahmedwaleedmalik:
### What happened?

Through https://github.com/kubermatic/docs/pull/1984, we ran the CIS benchmark using KubeOne 1.11.2 on k8s v1.33.5 and documented the results. We need to evaluate all the failing tests and preferably fix them or provide a proper explanation regarding why they are failing.

dharapvj:
Customer is specifically asking for ways to enable/disable and configure [EventRatelimit Admisssion Control plugin](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#eventratelimit) in the kubeone cluster. I will raise a separate issue for tracking this and cross-refer it here.

dharapvj:
opened the issue to track EventRateLimit AdmissionControl plugin https://github.com/kubermatic/kubeone/issues/3841

Also cross-referencing this old ticket that we might want to close if we have run the CIS benchmarks - https://github.com/kubermatic/kubeone/issues/2360

adoi:
This can be closed. After KubeOne 1.13 we will add another section to the docs with 1.13 CIS benchmarks with Kubernetes 1.35. This issue: #4031 will be the parent issue for that.

PR #4027, opened 2026-04-03
Add AlwaysPullImages wiring across kubeadm config and API server argument generation

adoi:
This PR adds support for configuring the `AlwaysPullImages` admission plugin through KubeOne. The setting is exposed in the API, preserved in printed config, and propagated into kubeadm-generated `kube-apiserver` arguments via `enable-admission-plugins`.

This provides end-to-end support for enabling `AlwaysPullImages` in KubeOne-managed clusters, including CIS-aligned environments where the admission plugin is required by the benchmark.

xref https://github.com/kubermatic/kubeone/issues/3826

**What type of PR is this?**

/kind feature
/kind api-change

- AlwaysPullImages is not part of the default upstream admission plugin set in currently supported Kubernetes versions. It is an optional controller typically enabled for specific security or multi-tenant requirements. That's why I have added it as an explicit configuration option as opposed to enabling it by default.

```release-note
Add support for configuring kubeadm `AlwaysPullImages` through KubeOne configuration.
```

**Documentation**:

```documentation
Documentation will be added as part of the CIS conformance results stating this is configurable.
```

### DIFF
--- a/pkg/apis/kubeone/types.go
+++ b/pkg/apis/kubeone/types.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2026 The KubeOne Authors.
+Copyright 2020 The KubeOne Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/pkg/apis/kubeone/v1beta2/types.go
+++ b/pkg/apis/kubeone/v1beta2/types.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2026 The KubeOne Authors.
+Copyright 2021 The KubeOne Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/pkg/apis/kubeone/v1beta3/types.go
+++ b/pkg/apis/kubeone/v1beta3/types.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2026 The KubeOne Authors.
+Copyright 2024 The KubeOne Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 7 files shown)


pkg/apis/kubeone/types.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2020 The KubeOne Authors.
+Copyright 2026 The KubeOne Authors.
```
kron4eg: Please revert this. Copyright yeah is not the current year it's the year since you claim the copyright.

pkg/apis/kubeone/v1beta2/types.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 The KubeOne Authors.
+Copyright 2026 The KubeOne Authors.
```
kron4eg: Same

pkg/apis/kubeone/v1beta3/types.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2024 The KubeOne Authors.
+Copyright 2026 The KubeOne Authors.
```
kron4eg: Same

pkg/cmd/config.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The KubeOne Authors.
+Copyright 2026 The KubeOne Authors.
```
kron4eg: Same

pkg/features/activate.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The KubeOne Authors.
+Copyright 2026 The KubeOne Authors.
```
kron4eg: Same

PR #4037, opened 2026-04-09
Add dnsServiceIP override for cluster DNS settings

### LINKED ISSUE #4033: Allow configuring -cluster-dns in KubeOneCluster and fix missing flag in OSM addon when NodeLocalDNS is disabled
**Problem:**
Currently, in the Operating System Manager (OSM) addon deployment, the `-cluster-dns` argument is only set if `features.nodeLocalDNS.deploy` is `true`. Reference logic https://github.com/kubermatic/kubeone/blob/main/addons/operating-system-manager/deployment-controller.yaml#L49-L51
If nodeLocalDNS is disabled it (set to `false`) in `kubeone.yaml`, the flag is omitted entirely. This prevents nodes from receiving a DNS server configuration via kubelet, causing name resolution failures. Additionally, there is currently no way for a user to manually override or specify a custom DNS server IP within the KubeOneCluster manifest.

**Proposed Changes:**

1. Fix Default Logic: Ensure that if nodeLocalDNS is disabled, the `-cluster-dns` flag defaults to the CoreDNS service IP (typically IP ending with `.10` from the serviceSubnet range defined in KubeOneCluster manifest) instead of being omitted.
2. Make Configurable: Update the KubeOneCluster CRD to allow users to explicitly define the Cluster DNS IP.

Suggested API Addition:
```yaml
kind: KubeOneCluster
...
clusterNetwork:
  serviceSubnet: "10.224.0.0/12"
  dnsServiceIP: "10.224.0.10"  # New optional field to override the default
```

**Expected Behavior:**

- If `dnsServiceIP` is provided in the CRD, OSM should use that value.
- If not provided and `nodeLocalDNS` is enabled, use the NodeLocalDNS IP.
- If not provided and `nodeLocalDNS` is disabled, use the default CoreDNS ClusterIP.

kron4eg:
Support `clusterNetwork.dnsServiceIP` in API versions and conversions, validate the value, and use it for kubelet/OSM when NodeLocalDNS is disabled. Fall back to the default 10th service-subnet IP when unset.

Fixes #4033

**What type of PR is this?**
/kind feature

```release-note
Add dnsServiceIP override for cluster DNS settings
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/addons/operating-system-manager/deployment-controller.yaml
+++ b/addons/operating-system-manager/deployment-controller.yaml
@@ -46,9 +46,7 @@ spec:
             - -log-format=json # json or console
             - -health-probe-address=0.0.0.0:8085
             - -metrics-address=0.0.0.0:8080
-            {{ if .Config.Features.NodeLocalDNS.Deploy -}}
-            - -cluster-dns={{ .Resources.NodeLocalDNSVirtualIP }}
-            {{ end -}}
+            - -cluster-dns={{ .ClusterDNSIP }}
             - -namespace=kube-system
             - -container-runtime={{ .Config.ContainerRuntime }}
             - -pause-image={{ .InternalImages.Get "PauseImage" }}
--- a/docs/api_reference/v1beta2.en.md
+++ b/docs/api_reference/v1beta2.en.md
@@ -1,6 +1,6 @@
 +++
 title = "v1beta2 API Reference"
-date = 2026-04-03T16:57:31+03:00
+date = 2026-04-15T10:59:17+03:00
 weight = 11
 +++
 ## v1beta2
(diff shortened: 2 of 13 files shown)


judge-red:
I'll preface this by saying that I don't understand golang and the code base well enough to fully understand what is really going on - and of course I've seen that this is just a draft at this time.

As said in #4187, which I opened today, a solution for the problem described in #4033 would be most appreciated. However, I think in our case `.Resources.NodeLocalDNSVirtualIP` _is_ the correct IP, since we use the `nodelocaldns-cilium` addon where this is also used. If I understand this PR correctly, there's no path to using `.Resources.NodeLocalDNSVirtualIP` for the cluster-dns IP anymore, so we would have to hardcode that value in the configuration instead.

That's of course still better than the current situation, but maybe you see a solution that could allow for this scenario too.

PR #4042, opened 2026-04-13
Fix Azure CCM and CNM image versions

kron4eg:
Fixes non-existing Azure CCM images.

Fixes 

**What type of PR is this?**
/kind regression

```release-note
Fix Azure CCM and CNM image versions
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/apis/kubeone/config/config.go
+++ b/pkg/apis/kubeone/config/config.go
@@ -192,14 +192,14 @@ func BytesToKubeOneCluster(cluster, tfOutput []byte, credentialsFilePath string,
 		}
 	}
 
-	// Validate the configuration
-	if err := kubeonevalidation.ValidateKubeOneCluster(*internalCluster).ToAggregate(); err != nil {
-		return nil, fail.ConfigValidation(err)
-	}
-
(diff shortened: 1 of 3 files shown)


pkg/apis/kubeone/config/config.go
```diff
@@ -192,16 +192,16 @@ func BytesToKubeOneCluster(cluster, tfOutput []byte, credentialsFilePath string,
...
 	// Apply the dynamic defaults
 	if err := SetKubeOneClusterDynamicDefaults(internalCluster, credentialsFilePath, baseDir); err != nil {
 		return nil, err
 	}
 
+	// Validate the configuration
+	if err := kubeonevalidation.ValidateKubeOneCluster(*internalCluster).ToAggregate(); err != nil {
+		return nil, fail.ConfigValidation(err)
+	}
+
```
kron4eg, line 204: This change is necessary as first we got to read the credentials file and initialize all the credentials from there to the main config and only after that we could validate correctly.

ahmedwaleedmalik (approved):
/approve

PR #4063, opened 2026-04-29
Fix: bastion host can have different ssh key

mohamed-rafraf:

When `bastionPrivateKeyFile` and `sshPrivateKeyFile` were different, KubeOne was using both keys for both the bastion and the target node.

This caused the wrong key (node key) to be tried first on the bastion, which often fails and exhausts authentication attempts before the correct key is used.

On strict servers, this leads to SSH connection failure.

Also, if `sshAgentSocket` was set but the agent had no keys, KubeOne failed instead of falling back to the configured key files.

This PR ensures the following: 

* Use separate SSH keys for bastion and node:

  * bastion uses only `bastionPrivateKeyFile`
  * node uses only `sshPrivateKeyFile`
* Create separate SSH configs for bastion and node connections
* Ensure correct key is used at each step
* If SSH agent is empty, ignore it and continue using key files instead of failing

Fixes #

**What type of PR is this?**

/kind bug

```release-note
The jump host can have a different SSH key.
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/ssh/connection.go
+++ b/pkg/ssh/connection.go
@@ -231,16 +231,17 @@ func NewConnection(connector *Connector, opts Opts) (executor.Interface, error) 
 		agentClient := agent.NewClient(socket)
 
 		signers, signersErr := agentClient.Signers()
-		if signersErr != nil {
+		switch {
+		case signersErr != nil:
 			socket.Close()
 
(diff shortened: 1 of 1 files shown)


kron4eg (approved):
/approve

PR #4088, opened 2026-05-20
Avoid modprobe ip_tables, as it's no more in the newer kernels

### LINKED ISSUE #3887: Cluster with kernel 6.17
As a user, I would like to add a node with kernel 6.17 to my cluster

When I try to add a new node to my cluster which runs mainline kernel 6.17, it fails with an error:

```
sudo modprobe ip_tables
modprobe: FATAL: Module ip_tables not found in directory /lib/modules/6.17.7-1.el8.elrepo.x86_64 
```

kron4eg:
kubernetes has moved to the nftables and latest kernels don't have ip_tables

Fixes #3887

**What type of PR is this?**
/kind regression

```release-note
Avoid modprobe ip_tables
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -21,7 +21,7 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@1e7e51e771db61008b38414a730f564565cf7c20 # v9.2.0
         with:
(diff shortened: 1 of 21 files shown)


adoi (approved):
/approve

PR #4096, opened 2026-05-26
Support for Kubernetes v1.36

### LINKED ISSUE #4071: Support Kubernetes 1.36
This is a collector issue for Kubernetes 1.36 support in KubeOne. The following tasks should be taken care of:

* [x] Check the Kubernetes changelog to ensure there are no breaking changes and removals
* [ ] Update the `build` image to update Sonobuoy (and other dependencies if needed) 
* [x] Update the latest supported Kubernetes version in [the API validation](https://github.com/kubermatic/kubeone/blob/main/pkg/apis/kubeone/validation/validation.go#L40-L45) 
* [x] Update [default admission controllers](https://github.com/kubermatic/kubeone/blob/main/pkg/kubeflags/data.go) if needed 
* [x] Update `pause` image version in `pkg/apis/kubeone/helpers.go` if needed 
* [x] Add the appropriate `cri-tools` version in `pkg/scripts/os_flatcar.go` 
* [x] Update [the stable version marker in Makefile](https://github.com/kubermatic/kubeone/blob/5273f9a372736569c6b09b38f2959019d29e4d6a/Makefile#L24) 
* [x] Add E2E tests inside `tests.yml` 
* [x] Update daily periodics to use the latest Kubernetes release
* [x] Update [the Compatibility Matrix](https://docs.kubermatic.com/kubeone/main/architecture/compatibility/supported-versions/) 
* [x] Create an issue to track updating [images](https://github.com/kubermatic/kubeone/blob/main/pkg/templates/images/images.go)

kron4eg:
This PR introduces initial support for Kubernetes v1.36. Some images will have to be updates closer to the release.

Fixes #4071

**What type of PR is this?**
/kind feature

```release-note
* Add support for Kubernetes v1.36
* Remove everything related to CCM/CSI migration
```

**Documentation**:

```documentation
NONE
```


pkg/apis/kubeone/helpers.go
```diff
@@ -36,7 +36,10 @@ const (
 	credentialSecretName = "kube-system/kubeone-registry-credentials" //nolint:gosec
 )
 
-var postV133Constraint = semverutil.MustParseConstraint(">= 1.33")
+var (
+	postV133Constraint = semverutil.MustParseConstraint(">= 1.33")
```
xmudrii: Nit: I don't think we need this branch given we only support 1.34+

pkg/kubeflags/flags.go
```diff
@@ -24,11 +24,16 @@ import (
 	"k8c.io/kubeone/pkg/semverutil"
 )
 
-var isV135Constraint = semverutil.MustParseConstraint(">= 1.35.0")
+var (
+	isV136Constraint = semverutil.MustParseConstraint(">= 1.36.0")
+	isV135Constraint = semverutil.MustParseConstraint(">= 1.35.0")
```
xmudrii: Nit: isn't this also going to catch v1.36?

pkg/scripts/os_flatcar.go
```diff
@@ -156,33 +156,33 @@ func RemoveBinariesFlatcar() (string, error) {
...
+	// switch kubeSemVer.Minor() {
+	// case 34:
+	// 	return "1.7.1"
+	// case 35:
+	// 	return "1.7.1"
+	// case 36:
+	// 	return "1.7.1"
+	// default:
+	return "1.7.1"
+	// }
```
xmudrii: We should remove this code if it's not used anymore.

xmudrii (changes requested):
Some nits, otherwise LGTM.

addons/csi-vsphere-ks/vsphere-csi-driver.yaml
xmudrii, line 1: Can we remove this addon already? It has been added in https://github.com/kubermatic/kubeone/pull/2327 to facilitate migration from the `kube-system` namespace to the `vsphere-system-csi` namespace, but I think all clusters should have been migrated long time ago.

addons/csi-vsphere/configmap.yaml
```diff
@@ -5,6 +5,6 @@ metadata:
   name: internal-feature-states.csi.vsphere.vmware.com
   namespace: vmware-system-csi
 data:
-  "csi-migration": "{{ .CSIMigration }}"
+  "csi-migration": "false"
```
xmudrii: This flag has always been interesting...

The purpose of this flag is unclear and not well documented. Some understanding of this flag is that it allows the CSI driver to work with volumes that were created with the legacy built-in storage plugin.

That said, if you disable this on clusters that had it enabled, and those clusters still have volumes that are created with the legacy built-in storage plugin, there's a risk that those volumes will stop working.

I'm wondering if we should just enable this by default. CSI migration is already in effect and it's not possible to disable it, this _shouldn't_ affect volumes that are created with the new StorageClass, and old clusters should have this set to `true` already.

addons/csi-vsphere-ks/vsphere-csi-driver.yaml
kron4eg: Good point

Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM docker.io/golang:1.26.1 as builder
+FROM docker.io/golang:1.26 as builder
```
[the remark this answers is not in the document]
kron4eg: This is deliberately, so Go could be updated automatically every time image is built.

xmudrii (approved):
/lgtm
/approve

PR #4098, opened 2026-05-29
Add openstack managed control-plane 

### LINKED ISSUE #4072: Manage Control Plane VMs on OpenStack

As a SysOp / Platform Admin
I would like to create control plane machines with KubeOne
in order to eliminate the need of usage of external tools like Terraform.

kron4eg:

Fixes #4072

**What type of PR is this?**
/kind feature

example config could look like this:

```yaml
kind: KubeOneCluster

name: cluster1

versions:
  kubernetes: 1.35.1

addons:
  enable: true
  addons:
  - name: default-storage-class

controlPlane:
  nodeSets:
    - name: cp
      replicas: 3
      operatingSystem: ubuntu
      ssh:
        publicKeys:
          - ssh-ed25519 xxx
        username: "ubuntu"
        bastion: "xx.xx.xx.xx"
        bastionUser: "ubuntu"
      cloudProviderSpec:
        image: "ubuntu-24.04"
        flavor: "m1.small"
        rootDiskSizeGB: 25
        configDrive: false
        securityGroups: ["cluster1-cluster"]
        network: "cluster1-cluster"
        subnet: "cluster1-cluster"
        tags:
          role: control-plane
          cluster: cluster1

cloudProvider:
  cloudConfig: |
    [Global]
    auth-url={{ .Credentials.OS_AUTH_URL }}
    username={{ .Credentials.OS_USERNAME }}
    password={{ .Credentials.OS_PASSWORD }}
    tenant-id={{ .Credentials.OS_TENANT_ID }}
    domain-name={{ .Credentials.OS_DOMAIN_NAME }}
    region={{ .Credentials.OS_REGION_NAME }}

    [LoadBalancer]
    lb-provider=""
    use-octavia=true

  openstack:
    controlPlane:
      loadBalancer:
        name: "cluster1-kube-apiserver"

dynamicWorkers:
  - name: work
    replicas: 3
    providerSpec:
      sshPublicKeys:
        - ssh-ed25519 xxx
      operatingSystem: ubuntu
      cloudProviderSpec:
        image: "ubuntu-24.04"
        flavor: "m1.small"
        rootDiskSizeGB: 50
        configDrive: false
        securityGroups: ["cluster1-cluster"]
        network: "cluster1-cluster"
        subnet: "cluster1-cluster"
        tags:
          role: worker
          cluster: cluster1
```

```release-note
Add openstack managed control-plane 
```

**Documentation**:

```documentation
TBD
```


pkg/tasks/cloud_resources_provision.go
```diff
@@ -68,80 +59,53 @@ func WithFindControlPlane(t Tasks) Tasks {
...
-		append(
-			Task{
-				Operation: "defaulting cluster hosts",
-				Predicate: func(s *state.State) bool { return len(s.Cluster.ControlPlane.NodeSets) != 0 },
-				Fn:        defaultCluster,
-			},
-		), nil
+			append(generateHetznerControlPlaneTasks(hetznerCAPIMachines)...)
+	case cluster.CloudProvider.Openstack != nil:
+		openstackCAPIMachines, err := generateOpenstackControlPlaneMachines(clusterName, nodeSet, kubeletVersion)
```
buraksekili, line 82: I am a bit confused here and would like to ask for some clarification.

`generateOpenstackControlPlaneMachines` generates machine-controller `Machine` objects by iterating over each `NodeSet` and creating one `Machine` per replica. So if we have 2 nodeSets with 3 replicas each, we get 6 machines.

Then `generateOpenstackControlPlaneTasks` creates one task per machine, each calling `ensureOpenstackControlPlaneVM`. But `hostConfigsFromMachines` gets a single machine and all nodeSets, so every machine gets paired with the first nodeSet's config; however, half of those machines were generated from the second nodeSet though.

Is this a known issue or am I misunderstanding the flow here?

pkg/provisioner/output.go
```diff
@@ -57,7 +57,8 @@ func GetMachineInfo(instance cloud.Instance) Machine {
 					publicAddressIPv6 = address
 				}
 			}
-		case corev1.NodeInternalIP:
+		case corev1.NodeInternalIP, "":
+			// empty addressType will be considered NodeInternalIP
```
buraksekili: hmm, MC returns an emtpy `addressType` for openstack - so, i assume that it includes floating-IPs as well. So, if the floating-ip comes from external ip-pool, again, this means we treat it as NodeInternalIP? what do you think?

https://github.com/kubermatic/machine-controller/blob/090279e10f6557926c29fce95405df4eaea44241/pkg/cloudprovider/provider/openstack/provider.go#L988

buraksekili: also, it looks like that since `instance.Addresses` returns `map`, iterating through this map here again randomizes which IP address is going to be used, as all IP addresses will have `""` and the code picks whichever comes first from the map. so, it is random whether the fixed IP or the floating IP wins

kron4eg: While it's all true, essentially it's the openstack's API at the core of the problem as it doesn't destingush between IPs. It just returns a list of addresses without any kinds.

```json
  "addresses": {
    "artioms-cluster": [
      "192.168.1.14",   <--- this in internal
      "195.1xx.1xx.xxx" <--- this is floating
    ]
  },
```

pkg/provisioner/output.go
```diff
@@ -98,9 +99,10 @@ func GetMachineInfo(instance cloud.Instance) Machine {
 
 func publicAndPrivateIPExist(addresses map[string]corev1.NodeAddressType) bool {
 	var publicIPExists, privateIPExists bool
-	// we only care about ExternalIP and InternalIP specifically, thus nolint
-	//nolint:exhaustive
+
 	for _, addressType := range addresses {
+		// we only care about ExternalIP and InternalIP specifically, thus nolint
+		//nolint:exhaustive
```
buraksekili, line 130: just an observation, `publicAndPrivateIPExist` never returns `true` for openstack due to openstack API, as you already pointed above. The caller of this function runs in retry-loop and burns the full 5x5s budget without ever breaking early.

kron4eg: Yeah, because IPs can be attached asynchronously, we want to wait a bit more.

pkg/provisioner/output.go
kron4eg: I've added parsing of private/public IPs of unknown type

pkg/tasks/cloud_resources_provision.go
kron4eg: no, `hostConfigsFromMachines()` generates the same number of HostConfigs as a product of nodeSets*replicas. For example, once could have 2 nodeSets with 2 replicas each, and 1 nodeSet with 1 replica. Totally 5 nodes, which `hostConfigsFromMachines()` will return.

buraksekili: it looks like, that holds when all machines are passed in a single call, like `lookupOpenstackVMs` does. The case I meant is the per-machine task path, like `generateOpenstackControlPlaneTasks` creates one task per machine, and each task calls `ensureOpenstackControlPlaneVM`, which calls `hostConfigsFromMachines` with a single machine but all nodeSets. as `idx` restarts at 0 on every call, that single machine is always paired with `nodeSets[0]`, even if it was generated from another nodeSet.

for ex, i created https://go.dev/play/p/Fxqr_yJb4XE to demonstrate the failing test case i meant.
```
$ go test ./pkg/tasks/ -run 'TestHostConfigsFromMachines' -v
=== RUN   TestHostConfigsFromMachines
=== RUN   TestHostConfigsFromMachines/all_machines_at_once
=== RUN   TestHostConfigsFromMachines/one_machine_per_call
    cloud_resources_provision_test.go:78: host "cluster1-cp-flatcar-0": expected ssh username "core", got "ubuntu"
    cloud_resources_provision_test.go:81: host "cluster1-cp-flatcar-0": expected OS "flatcar", got "ubuntu"
--- FAIL: TestHostConfigsFromMachines (0.00s)
    --- PASS: TestHostConfigsFromMachines/all_machines_at_once (0.00s)
    --- FAIL: TestHostConfigsFromMachines/one_machine_per_call (0.00s)
FAIL
FAIL    k8c.io/kubeone/pkg/tasks        0.770s
FAIL
```

let me know what do you think

kron4eg: The input for hostConfigsFromMachines() is wrong in this test. In the real code hostConfigsFromMachines() is fed with Machines that correspond to the nodeSets times replicas.

buraksekili (commented):
/lgtm
/approve

PR #4113, opened 2026-06-24
Add successfulJobsHistoryLimit param to backups-restic addon

### LINKED ISSUE #4101: Make successfulJobsHistoryLimit configurable
`successfulJobsHistoryLimit` should be configurable and by default `1` to verify the container log that Backup is actualy working and not generating a false negative.

https://github.com/kubermatic/kubeone/blob/a6b67308d18dd3c2e565d1d545ed150d5ac7fd98/addons/backups-restic/backups-restic.yaml#L29

kron4eg:

The `backups-restic` embedded addon deploys a `CronJob` (`etcd-s3-backup`) that periodically snapshots etcd state and pushes it to an S3-compatible backend using restic. Prior to this change the `successfulJobsHistoryLimit` field on that CronJob was hardcoded to `0`, meaning Kubernetes would immediately garbage-collect every completed (successful) Job pod. This made it impossible to inspect pod logs after a backup run to confirm the backup actually succeeded — operators had to rely on the absence of a failure alert rather than a positive confirmation. The previous default of `0` was also inconsistent with the general Kubernetes ecosystem convention (e.g. KKP's own `etcd-defragger` CronJob uses `1`) and with the already-present `failedJobsHistoryLimit: 1` on the same CronJob.

This PR makes `successfulJobsHistoryLimit` a user-tuneable addon parameter while changing the built-in default from `0` to `1`. With `1` as the default, one successful Job pod is retained after each backup cycle, giving operators the ability to run `kubectl logs -n kube-system <pod>` to verify that the restic upload actually completed without errors. Users who want the previous space-saving behaviour of `0`, or who prefer a larger retention window, can override the value via the addon `Params` map in their `kubeone.yaml`:

```yaml
addons:
  enable: true
  addons:
    - name: backups-restic
      params:
        resticPassword: "<secret>"
        successfulJobsHistoryLimit: 3   # keep last 3 successful job pods
```

The implementation follows the existing pattern already used by other parameters in this addon (e.g. `resticPassword`) and uses the `default` template function so that clusters that do not set the parameter continue to work with the new sensible default of `1`.

Fixes #4101

**What type of PR is this?**
/kind feature

```release-note
Add successfulJobsHistoryLimit param to backups-restic addon
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/addons/backups-restic/backups-restic.yaml
+++ b/addons/backups-restic/backups-restic.yaml
@@ -26,7 +26,7 @@ spec:
   concurrencyPolicy: Forbid
   failedJobsHistoryLimit: 1
   schedule: '@every 30m'
-  successfulJobsHistoryLimit: 0
+  successfulJobsHistoryLimit: {{ default 1 .Params.successfulJobsHistoryLimit }}
   suspend: false
   jobTemplate:
     spec:


buraksekili (approved):
/lgtm
/approve

judge-red:
Awesome, thanks!

PR #4114, opened 2026-06-26
Regenerate apiserver.crt when APIEndpoint.AlternativeNames has something new

### LINKED ISSUE #4060: SAN in kube-apiserver certificates not updated
### What happened?

In some environments, the VIP used for the kube-api after boostrapping the environment needs to be changed. In those cases we updated the "KubeOneCluster" CR, and add a new "apiEndpoint.alternativeName" to the list.
We then run "kubeony apply --force-upgrade", but even though the certificate is recreated, it is still using the original alternativeNames in the SAN value for the cert. It seems like the only way to get the new SAN into the cert is to manually remove the certificate on each control-plane node, and then run Kubeone.
The "kubeone certificates renew" command was also tested, but that did not update the SAN either.

customer ref: 9039

### Expected behavior

Certificates should be updated when running `kubeone apply -f ...` after updating alternativeNames in kubeone spec.

<details>

```console
$ kubeone version
# paste output here
```

</details>

### Provide your KubeOneCluster manifest here (if applicable)

<details>

```yaml
# paste manifest here
```

</details>

kron4eg:

`kubeadm` preserves the existing SANs of a certificate when renewing it. Kubeadm will not inject new Subject Alternative Names into an already-present cert, even if the kubeadm config has changed. This means that adding a new entry to `apiEndpoint.alternativeNames` in a `KubeOneCluster` manifest and running `kubeone apply --force-upgrade` or `kubeone certificates renew` did not actually update the SAN list of the `kube-apiserver` certificate.

This PR fixes the issue by detecting the mismatch before renewal: it reads the current `apiserver.crt` from each control-plane node, compares its `DNSNames` against the `APIEndpoint.AlternativeNames` in the cluster config and if any names are missing — deletes the certificate so that the subsequent `kubeadm init phase certs all` regenerates it with the full, updated SAN list. The standard `kubeadm certs renew all` then follows as usual.

Fixes #4060

**What type of PR is this?**
/kind feature

```release-note
`kubeone certificates renew` will detect and fix the drift between apiserver certificate and APIEndpoint.AlternativeNames
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/cmd/certificates.go
+++ b/pkg/cmd/certificates.go
@@ -65,15 +65,26 @@ func certificatesRenewCmd(rootFlags *pflag.FlagSet) *cobra.Command {
 				return err
 			}
 
-			renew := tasks.Tasks{
-				{Fn: certificate.RenewAll, Operation: "renew all the certificates of the Kubernetes control plane and kubelets"},
-				{Fn: kubeconfig.BuildKubernetesClientset, Operation: "building kubernetes clientset"},
-				{
-					Fn: func(s *state.State) error {
-						return s.RunTaskOnLeader(tasks.ApprovePendingCSR)
-					},
-				},
+			probbing := tasks.WithHostnameOS(nil)
+			probbing = tasks.WithProbesAndSafeguard(probbing)
+
+			if err := probbing.Run(st); err != nil {
+				return err
 			}
+
+			renew := append(
+				tasks.KubernetesConfigFiles(),
+				tasks.Tasks{
+					{Fn: certificate.RenewAll, Description: "Renew all the certificates of the Kubernetes control plane and kubelets"},
+					{Fn: kubeconfig.BuildKubernetesClientset, Operation: "building Kubernetes clientset"},
+					{
+						Fn: func(s *state.State) error {
+							return s.RunTaskOnLeader(tasks.ApprovePendingCSR)
+						},
+						Description: "Approve pending kubelet CSRs",
+					},
+				}...,
+			)
 
 			return renew.Run(st)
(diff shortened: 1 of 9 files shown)


pkg/cmd/certificates.go
```diff
@@ -65,16 +65,27 @@ func certificatesRenewCmd(rootFlags *pflag.FlagSet) *cobra.Command {
...
+			probbing = tasks.WithProbesAndSafeguard(probbing)
+
+			if err := probbing.Run(st); err != nil {
+				return err
 			}
 
+			renew := append(
+				tasks.KubernetesConfigFiles(),
+				tasks.Tasks{
+					{Fn: certificate.RenewAll, Description: "Renew all the certificates of the Kubernetes control plane and kubelets"},
```
buraksekili, line 78: newbie question actually :) `RenewAll` seems only to be reachable from the `kubeone certificates renew` command (`pkg/cmd/certificates.go`), not from the upgrade path that you explained in the PR description - is that expected?

```
kubeone apply --force-upgrade or kubeone certificates renew 
```

I guess, as part of this PR, the latter works fine, but the first one is not addressed if I am not mistaken?

buraksekili (commented):
overall lgtm, i just have a quick newbie question.

pkg/cmd/certificates.go
kron4eg: yes, `certificate.RenewAll()` has been called only from `kubeone certificates renew` and it's continues to be so, to avoid sudden certificate changes.

PR #4156, opened 2026-07-27
Support user-defined custom secrets in credentials.yaml for addon templates

### LINKED ISSUE #3927: Support for user-defined credentials/secrets key-values in KubeOne credentials file

Currently, KubeOne provides support configuration of provider specific credentials in the credentials file which get resolved out of box https://docs.kubermatic.com/kubeone/v1.12/guides/credentials/#credentials-file.
However, as an end user, I am looking for extended usage of credentials.yaml where user-defined key values secrets which can be set and gets resolved during the custom addons  deployment when passed to `kubeone apply .. -c credentials.yaml`.

### Solution details

- Requirement 1
- Requirement 2
- Requirement 3

### Additional information

Internal reference - 8676

steled:

Adds a reserved `customSecrets` top-level key to the KubeOne credentials file (`-c credentials.yaml`), following the same "embedded YAML block string" convention already used by `registriesAuth`. Values placed under `customSecrets` are parsed into a flat `map[string]string` and exposed to addon templates (both built-in and custom addons) as `.CustomCredentials.KEY`, using the exact same `templateData`/Go-template plumbing that already exposes provider credentials as `.Credentials.KEY` (see e.g. `addons/backups-restic/backups-restic.yaml`).

This lets users reference arbitrary secret values from `credentials.yaml` inside their own custom addon manifests, instead of having to pass them via `kubeone.yaml`'s `Addon.Params` (which is typically committed to VCS and not intended for secrets).

Example:
```yaml
# credentials.yaml
AWS_ACCESS_KEY_ID: xxx
AWS_SECRET_ACCESS_KEY: yyy
customSecrets: |
  MY_APP_TOKEN: supersecret
  ANOTHER_KEY: value
```
```yaml
# custom addon manifest
apiVersion: v1
kind: Secret
metadata:
  name: my-app-secret
stringData:
  token: "{{ .CustomCredentials.MY_APP_TOKEN }}"
```

Fixes #3927

**What type of PR is this?**
/kind feature

- Since custom and built-in addons share one `templateData`/template pipeline, there's no way to expose `customSecrets` only to custom addons as the issue title implies — built-in addon templates can reference `.CustomCredentials` too, same as they already can with `.Credentials`. Flagging this as a scope decision rather than resolving it silently.
- Chose a dedicated `customSecrets:` block (Option B) over a flat pass-through merge into the existing `.Credentials` map, to keep user-defined secrets namespaced separately from KubeOne's own provider-credential keys and avoid any future collision risk.
- No changes were needed to the cluster-config parsing in `pkg/apis/kubeone/config/config.go` since this feature is addon-scoped only (per the issue).

```release-note
Added support for a `customSecrets` section in the credentials file, allowing arbitrary user-defined secrets to be referenced from addon templates (built-in and custom) as `.CustomCredentials.KEY`.
```

**Documentation**:
```documentation
TBD (documentation will be added in a follow-up kubermatic/docs PR, alongside the existing `cloudConfig` / `csiConfig` / `registriesAuth` credentials-file documentation)
```

### DIFF
--- a/pkg/addons/applier.go
+++ b/pkg/addons/applier.go
@@ -72,6 +72,7 @@ type templateData struct {
 	Config                                   *kubeoneapi.KubeOneCluster
 	Certificates                             map[string]string
 	Credentials                              map[string]string
+	CustomCredentials                        map[string]string
 	CredentialsCCM                           map[string]string
 	CredentialsCCMHash                       string
 	CCMClusterName                           string
@@ -101,6 +102,11 @@ func newAddonsApplier(s *state.State) (*applier, error) {
 	}
 
 	creds, err := credentials.Any(s.CredentialsFilePath)
+	if err != nil {
+		return nil, err
+	}
+
+	customCreds, err := credentials.Custom(s.CredentialsFilePath)
 	if err != nil {
 		return nil, err
 	}
@@ -176,6 +182,7 @@ func newAddonsApplier(s *state.State) (*applier, error) {
 			"KubernetesCA":                 mcCertsMap[resources.KubernetesCACertName],
 		},
 		Credentials:                         creds,
+		CustomCredentials:                   customCreds,
 		CredentialsCCM:                      credsCCM,
 		CredentialsCCMHash:                  credsCCMHash,
 		CCMClusterName:                      s.LiveCluster.CCMClusterName,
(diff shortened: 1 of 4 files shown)


kron4eg:
@steled thank you for your contribution!

archups:
Seems already included in release v1.14.0

PR #4186, opened 2026-08-14
Move verify prow jobs to GH actions

kron4eg:
Move more prow jobs to GH actions

**What type of PR is this?**
/kind cleanup

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


go.mod
```diff
@@ -1,6 +1,6 @@
 module k8c.io/kubeone
 
-go 1.26.5
+go 1.26.0
```
buraksekili, line 3: Is this intended?

kron4eg: yes, it has previously slipped under the radar but shouldn't be there

buraksekili (approved):
/lgtm
/approve

PR #4192, opened 2026-08-16
Upgrade static workers using individual tasks

### LINKED ISSUE #4059: already handled nodes should be skipped when kubeone retries the upgrade of static workers

When kubeone tries to upgrade static workers and one fails, it always starts from the beginning and is not skipping already upgraded nodes.

As a kubeone user
I would like to skip already upgraded nodes when kubeone retries the upgrade process after a static worker upgrade failed.

### Solution details

- when kubeone retries the upgrade of static workers after it failed on one node it should check if a static worker was already upgraded and skip the upgrade loop for that worker.

taaahaaa:

Static worker upgrades currently run as one task that processes all workers
sequentially. If an upgrade fails on a later worker, the task retry starts from
the first worker and repeats the cordon, drain, and upgrade operations on nodes
that already succeeded.

Generate one task per static worker instead. Each task retains the existing
retry behavior, but a failure now retries only the affected worker. This
matches the existing approach used for control plane followers.

Fixes #4059

**What type of PR is this?**

/kind feature

The tests verify that:

- no tasks are generated when there are no static workers;
- one task is generated per static worker, in configuration order;
- each generated task targets only its corresponding worker.

Verified locally with:

```text
go test ./pkg/... ./test/...
go vet ./pkg/tasks ./pkg/cmd
go test -race ./pkg/tasks -count=1
golangci-lint run --timeout=5m -v ./pkg/... ./test/...
```

```release-note
Static worker upgrade retries now retry only the failed worker instead of repeating upgrades on workers that already succeeded.
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/cmd/apply.go
+++ b/pkg/cmd/apply.go
@@ -388,7 +388,7 @@ func runApplyUpgradeIfNeeded(s *state.State, opts *applyOpts) error {
 			tasksToRun = tasks.WithDisableEncryptionProviders(tasksToRun, s.LiveCluster.EncryptionConfiguration.Custom)
 		}
 
-		tasksToRun = tasks.WithUpgrade(tasksToRun, s.Cluster.Followers()...)
+		tasksToRun = tasks.WithUpgrade(tasksToRun, s.Cluster.Followers(), s.Cluster.StaticWorkers.Hosts)
 
 		if s.ShouldEnableEncryption() {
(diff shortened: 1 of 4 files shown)


kron4eg (approved):
/approve

kron4eg:
@taaahaaa thank you for your contribution!

PR #4206, opened 2026-08-19
Support per-addon secret params in credentials.yaml via addonParams

steled:

Adds a reserved `addonParams` top-level key to the KubeOne credentials file (`-c credentials.yaml`), following the same "embedded YAML block string" convention already used by `customSecrets` (#4156). Its value is a YAML mapping of addon name to a mapping of param name to value:

```yaml
addonParams: |
  backups-restic:
    resticPassword: "..."
```

At render time, these are merged into that specific addon's regular `.Params` map (the same map populated from the KubeOneCluster manifest's `addons[].params` in `pkg/addons/manifest.go`), so existing addon templates - built-in or custom - work completely unmodified; they don't need to know about `.CustomCredentials` at all.

This solves a gap that `customSecrets` (#4156) doesn't cover: built-in addons (e.g. `backups-restic`) already read secret-like values via `.Params.KEY` (e.g. `resticPassword`), which today can only be set via `kubeone.yaml`'s `Addon.Params` - a file that's typically committed to VCS and not intended for secrets. With `addonParams`, operators can move those values into `credentials.yaml` instead, without needing a local addon override just to reference `.CustomCredentials`.

If the same param key is set both in `credentials.yaml`'s `addonParams` and in the KubeOneCluster manifest's `addons[].params` for the same addon, rendering now fails with an explicit error rather than silently picking one.

N/A (no tracking issue)

**What type of PR is this?**
/kind feature

- Chose a dedicated `addonParams:` block, keyed by addon name, over reusing addon names directly as top-level reserved keys in `credentials.yaml` - the credentials file parses into a `map[string]string` (`credentialsFinder.static`), so any nested/structured value has to go through the same "block string, parsed again" trick already used by `customSecrets`/`registriesAuth`/`cloudConfig`/`csiConfig`. A single reserved key avoids turning every current and future addon name into an implicitly reserved word in `credentials.yaml`.
- Erroring on a key defined in both places (rather than silently letting one win) is a deliberate choice to avoid confusing "which source actually applied" situations; happy to change this to "KubeOneCluster manifest wins" if reviewers prefer the more permissive behavior.
- No changes were needed in `pkg/apis/kubeone/config/config.go` - this is addon-scoped only, resolved entirely in `pkg/credentials` and `pkg/addons`.

```release-note
Added support for an `addonParams` section in the credentials file, allowing addon secret parameters (e.g. the built-in `backups-restic` addon's `resticPassword`) to be provided via `credentials.yaml` instead of the KubeOneCluster manifest.
```

**Documentation**:
```documentation
https://github.com/kubermatic/docs/pull/2273
```

### DIFF
--- a/pkg/addons/applier.go
+++ b/pkg/addons/applier.go
@@ -65,6 +65,10 @@ type applier struct {
 	TemplateData templateData
 	LocalFS      fs.FS
 	EmbeddedFS   fs.FS
+	// AddonParams holds per-addon secret parameters from the credentials
+	// file's "addonParams" section, keyed by addon name. See
+	// credentials.AddonParams for details.
+	AddonParams map[string]map[string]string
 }
 
 // TemplateData is data available in the addons render template
@@ -107,6 +111,11 @@ func newAddonsApplier(s *state.State) (*applier, error) {
 	}
 
 	customCreds, err := credentials.Custom(s.CredentialsFilePath)
+	if err != nil {
+		return nil, err
+	}
+
+	addonParams, err := credentials.AddonParams(s.CredentialsFilePath)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


kron4eg (approved):
/approve

kron4eg:
@steled thank you for the contribution!!

PR #4221, opened 2026-09-16
Fix OCI registry auth for helm charts and digest-pinned image retagging

mohamed-rafraf:
**What this PR does / why we need it**
Two independent failures that both surface when installing into a private, air-gapped registry.

`HelmRelease.Auth` is silently ignored for `oci://` chart URLs: it is assigned to `ChartPathOptions.Username/Password`, but Helm only reads those for HTTP chart repos — `LocateChart` branches on `registry.IsOCI()` and hands OCI refs to the registry client, which we build with no credentials, so a private registry's Basic challenge resolves against an empty credential store and the install dies with `basic credential not found` while giving no hint that the configured auth was dropped.

`retagImage` builds every destination as `<registry>/<repo>:<identifier>`, but `Identifier()` returns the digest for a digest-pinned source — producing `<registry>/<repo>:sha256:...`, which is not a parseable reference since a tag cannot contain a colon, so `mirror-images` fails on any digest-pinned image (e.g. `quay.io/operator-framework/olm@sha256:…`) before a single byte is copied.

Fixes #

**What type of PR is this?**

/kind bug

```release-note
- Fixed OCI Helm chart authentication so `HelmRelease.Auth` credentials are correctly passed to private OCI registries.
- Fixed `mirror-images` failing to retag digest-pinned images by preserving the source digest instead of incorrectly using it as a tag.

```

**Documentation**:

```documentation
NONE
```


pkg/localhelm/helm3.go
```diff
@@ -583,19 +591,45 @@ func dependencyUpdate(chartPath string, helmSettings *helmcli.EnvSettings, provi
...
 		registry.ClientOptDebug(debug),
 		registry.ClientOptEnableCache(true),
 		registry.ClientOptWriter(registryWriter),
-	)
+	}
 
-	return &helmaction.Configuration{
-		RegistryClient: registryClient,
-	}, fail.Runtime(err, "initializing new helm registry client")
+	if auth != nil && (auth.Username != "" || auth.Password != "") {
```
rajaSahil, line 626: Shouldn't we use `&&` instead of `||` ?

kron4eg:
@mohamed-rafraf please add a release note

pkg/localhelm/helm3.go
kron4eg: It,s OK. It's for the case when username or password could be omitted (some token for example).

kron4eg:
/lgtm
/approve

/hold

@mohamed-rafraf once you add the release-notes, you may unhold it

